PR #2, opened 2024-10-10
add static analysis gradle plugins (spotless + palantir, errorprone + nullaway)

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-5

The following static analysis gradle plugins are integrated:

1. **code style** checking: **spotless** (including **palantir** component for Java language)
2. **code quality** checking: **errorprone** (including **nullaway** as a component)


README.md
```diff
@@ -1 +1,15 @@
 A MongoDB Dialect for the Hibernate ORM
+
+## Build
+
+[Gradle](https://gradle.org/) is the building tool with [Kotlin](https://kotlinlang.org/) DSL as default.
+
+Starting with Gradle 8.2, creating new builds using gradle init defaults to generating Gradle builds using Kotlin DSL.
```
jyemin, line 7: Seems unnecessary to include this line

README.md
```diff
@@ -1 +1,15 @@
 A MongoDB Dialect for the Hibernate ORM
+
+## Build
+
+[Gradle](https://gradle.org/) is the building tool with [Kotlin](https://kotlinlang.org/) DSL as default.
```
jyemin, line 5: building => build

README.md
```diff
@@ -1 +1,15 @@
...
+
+Starting with Gradle 8.2, creating new builds using gradle init defaults to generating Gradle builds using Kotlin DSL.
+
+See https://blog.gradle.org/kotlin-dsl-is-now-the-default-for-new-gradle-builds for further details.
+
+* Kotlin DSL Primer: https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotlin_dsl
+* Kotlin DSL API: https://docs.gradle.org/current/kotlin-dsl/index.html
+* Migration from Groovy: https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html
```
jyemin, line 13: This line seems unnecessary

build.gradle.kts
```diff
@@ -42,3 +51,35 @@ tasks.named<Test>("test") {
...
     useJUnitPlatform()
 }
+
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// Static Analysis Tasks
+
+spotless {
+    java {
+        // note: you can use an empty string for all the imports you didn't specify explicitly, '|' to join group without blank line, and '\\#` prefix for static imports
+        importOrder("java|javax", "org.hibernate", "com.mongo", "", "\\#")
```
jyemin, line 61: mongo -> mongodb

jyemin, line 61: Just curious why you want a non-default import order. Is the default not good?

copyright.txt
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2008-$YEAR MongoDB, Inc.
```
jyemin, line 2: 2008 -> 2024

build.gradle.kts
```diff
@@ -42,3 +51,35 @@ tasks.named<Test>("test") {
...
+
+        removeUnusedImports()
+
+        // Cleanthat will refactor your code, but it may break your style: apply it before your formatter
+        cleanthat()
```
jyemin, line 66: What will this do by default?  Do we need it?

jyemin (changes requested):

build.gradle.kts
NathanQingyangXu: the main reason is to add group blank line delimiter for the two main importing groups. Not a big deal for sure.

build.gradle.kts
NathanQingyangXu: I didn't try it but it is there in doc. Guess it might change code (reordering imports, removing unused imports, etc.). Without this, not sure whether the previous `removeUnusedImports()` will take effect or not.

NathanQingyangXu: Let me experiment it a little bit and will ping back.

NathanQingyangXu: cleanthat() has nothing to do the import ordering and  removal of unused imports. It has its own complicated mutation rules and seems pretty powerful, e.g. this snippet is from the spotless README.md (https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#cleanthat):

Screenshot 2024-10-11 at 9 49 49 AM

given it is used (with default config) at https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#java, so maybe let us retain it without further customization?

build.gradle.kts
NathanQingyangXu: I reomved both for I think it is over engineering for now

build.gradle.kts
jyemin: I don't know, it seems much less well known than errorprone.  Also, I don't like that the only documentation of SafeAndConsensual is the [source code](https://github.com/solven-eu/cleanthat/blob/master/java/src/main/java/eu/solven/cleanthat/engine/java/refactorer/mutators/composite/SafeAndConsensualMutators.java).

Especially since it's a code mutator and not just a linter, I would feel more comfortable pulling it out.

NathanQingyangXu: I removed it.

jyemin:
> remove usage of jspecify (mongo's nullness annotations suffice)

I don't think we should do this, as the driver's null-related annotations are not intended to be part of the driver API.  They are just an implementation detail.

NathanQingyangXu:
> > remove usage of jspecify (mongo's nullness annotations suffice)
> 
> I don't think we should do this, as the driver's null-related annotations are not intended to be part of the driver API. They are just an implementation detail.

reverted back

jyemin (approved):
LGTM, but added @rozza for a second set of eyes

rozza (approved):
LGTM!

Out of interest was the `.editorconfig` change a manual change or automatic based on the formatting plugins?

NathanQingyangXu:
Palantir Intellij plugin is not well-maintained and latest plugin simply
doesn't work in latest Intellij. Yeah, we can invoke gradle task to
reformat automatically. But an .editorconfig ensures the least surprise so
he/she seldom needs to be bothered with deep diving to figure it out why CI
building failed.

NathanQingyangXu:
Good point on jspecify scope! It would be great to limit its visibility so
our client won't include it.

build.gradle.kts
```diff
@@ -29,6 +33,11 @@ dependencies {
     testImplementation(libs.junit.jupiter)
 
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+
+    errorprone(libs.nullaway)
+    api(libs.jspecify)
```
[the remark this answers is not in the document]
NathanQingyangXu: I found the following statement at https://jspecify.dev/docs/using/

Screenshot 2024-10-21 at 9 56 50 AM

So let us stick to `api` scope for now and always keep it in mind to switch to lower visibility. Currently given almost zero experience on real usage, let us trust the doc. The switch cost should be minor in the future, if needed.

PR #4, opened 2024-10-15
add evergreen integration

NathanQingyangXu:
As discussed, a minimal working evergreen setup suffices for now, as long as it conducts static analysis and unit testing


jyemin (approved):
I assume you weren't able to test this, since there is no Evergreen project.  If so, let's just commit this and then iterate, once we see what the actually build looks like.

.evergreen/.evg.yml
```diff
@@ -0,0 +1,46 @@
+functions:
```
jyemin, line 1: Rename this file to config.yml.  Java driver file name is not what most newer projects use.

jyemin (changes requested):

.evergreen/java-config.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Java configurations for evergreen
+
+export JDK8="/opt/java/jdk8"
```
jyemin, line 5: Let's remove JDK8 and JDK11! :)

jyemin (changes requested):

NathanQingyangXu:
> I assume you weren't able to test this, since there is no Evergreen project. If so, let's just commit this and then iterate, once we see what the actually build looks like.

I could test locally by choosing the branch name (together with evergreen config file path) as below in the project settings page:

Screenshot 2024-10-18 at 10 21 15 AM

That is how I tested it previously. will try to investigate whether we could see the CI effect here by github integration option in the project settings as well.

NathanQingyangXu:
I tested this branch indirectly by creating a separate temporary PR at https://github.com/mongodb/mongo-hibernate/pull/5. Now testing succeeded and I closed that testing PR. Latest commits in this PR reflected the issues flushed out during the testing.

.evergreen/config.yml
```diff
@@ -0,0 +1,47 @@
...
+    - command: git.apply_patch
+
+  exec script:
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src
+        shell: bash
+        script: |
+          ${PREPARE_SHELL}
```
jyemin, line 17: Doesn't this needs to be defined somewhere? I'm surprised this doesn't cause the script to error.

.evergreen/run-test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+############################################
+#            Main Program                  #
+############################################
+source ./.evergreen/java-config.sh
+
+echo "mongo-hibernate: static checking ..."
```
jyemin, line 11:
```suggestion
echo "mongo-hibernate: running tests ..."
```

.evergreen/config.yml
```diff
@@ -0,0 +1,47 @@
...
+    commands:
+      - func: exec script
+        vars:
+          file: .evergreen/run-test.sh
+
+buildvariants:
+
+  - name: simple-variant
+    tags: ["ci"]
+    display_name: Simple Variant
```
jyemin, line 44: Do we want all of these to have the same display name?  In Java driver, for example, the display name for static checks is "Static Checks".

jyemin (changes requested):

.evergreen/config.yml
NathanQingyangXu: This display_name is for variant, not for task, so its counterpart is like: 

**4.2 Sharded Cluster NoAuth NoSSL JDK8 Linux**

.evergreen/config.yml
NathanQingyangXu: I removed it for currently it is not defined.
Guess shell might interpret unknown variable as empty string. For instance, the following command:
```
# echo $UNKNOWN_VARIABLE
```
simply output an empty string without complaining, unless `set -u` is run first.

.evergreen/config.yml
NathanQingyangXu: I changed it to concrete variant name instead. We'd better avoid ambiguous 'simple-variant' expression.

.evergreen/config.yml
```diff
@@ -39,9 +38,9 @@ tasks:
 
 buildvariants:
 
-  - name: simple-variant
+  - name: linux-rhel
     tags: ["ci"]
-    display_name: Simple Variant
+    display_name: Redhat Linux
```
jyemin, line 43: I'd still prefer just Linux until we feel the need to test on multiple linux variants (which we never will !).

jyemin (changes requested):

jyemin (dismissed):

NathanQingyangXu:
I rebased on the main branch and made the following changes:

1. change palantir change to `palantirJavaFormat().formatJavadoc(true)`
2. delete the '.editorconfig' file for it was based on google-java-format. There is no .editorconfig file for palantir, so no bother

jyemin:
FYI, if you force push like this, reviewers are not able to see what has changed since the last review.

gradle/libs.versions.toml
```diff
@@ -4,7 +4,6 @@
 [versions]
 junit-jupiter = "5.10.3"
 spotless = "6.25.0"
-palantir = "2.50.0"
```
jyemin, line 7: Why don't we need this?  Is it because it's picked up from Spotless?

build.gradle.kts
```diff
@@ -62,7 +62,7 @@ spotless {
 
         removeUnusedImports()
 
-        palantirJavaFormat(libs.versions.palantir.get()).style("GOOGLE").formatJavadoc(true)
+        palantirJavaFormat().formatJavadoc(true)
```
jyemin, line 65: As this doesn't have anything to do with Evergreen, can we do a separate PR for it?

jyemin (changes requested):

NathanQingyangXu:
yeah, a separate PR is justified. Will create one.

build.gradle.kts
NathanQingyangXu: yeah, will do.

jyemin (approved):

PR #6, opened 2024-10-21
rectify palantir configure issue

NathanQingyangXu:


build.gradle.kts
```diff
@@ -62,7 +62,7 @@ spotless {
 
         removeUnusedImports()
 
-        palantirJavaFormat(libs.versions.palantir.get()).style("GOOGLE").formatJavadoc(true)
+        palantirJavaFormat(libs.versions.palantir.get()).formatJavadoc(true)
```
jyemin, line 65: Is `libs.versions.palantir.get()` the default?  If so, do we need it?

jyemin (commented):
I'm a bit worried about removing .editorconfig entirely.  There is always going to be a default one, and starting with Google one and modifying it ourselves may be better than none at all. 

What do you think?

NathanQingyangXu:
> I'm a bit worried about removing .editorconfig entirely. There is always going to be a default one, and starting with Google one and modifying it ourselves may be better than none at all.
> 
> What do you think?

I hesitated a lot as well. Yeah, .editorconfig is a programmatic way to ensure developer won't be troubled with common format issue (Palantir IDE plugin seems not even working from my manual testing), so it seems a basic one goes a long way.

It is difficult to get one .editorconfig for palantir, but it is easy to get one from google-java-format (in Intellij IDE, after we install google-java-format, we could export a corresponding .editorconfig file from there). Currently the main difference is we need to change its indent width from 2 to 4.

Without .editorconfig and IDE plugin, for palantir the only option is to run gradle task (e.g. _spotlessJavaApply_). That seems far from being ideal.

could we revert the old .editorconfig (I exported from IDE after google-java-format plugin is installed. But I am not sure whether changing indent width from 2 to 4 is enough.

build.gradle.kts
NathanQingyangXu: I double checked. The default version is based on the following internal "matrix" recommendation:

```
private static final Jvm.Support<String> JVM_SUPPORT = Jvm.<String> support(NAME).add(8, "1.1.0").add(11, "2.28.0").add(21, "2.39.0");
```

So if our Java version is 17, `2.28.0` will be chosen.

I missed such subtlety previously and use the latest version `2.50.0` at `libs.versions.palantir`. Now I am not sure it is a good idea and it seems the default recommended version (i.e. `2.28.0`) is enough?

Do you have more experience on palantir? It is totally new to me. As a starter, I think maybe relying on its internal recommendation based on java version is enough (we could specify version later if we found some defect which was fixed in later version?)?

NathanQingyangXu: I experimented locally and found if we use default version (`2.28.0` corresonding to `java 11`), the `spotlessJavaDiagnose` task will complain as below:

```
Cannot enable formatJavadoc option, make sure you are using Palantir with version 2.36.0 or later
```

So maybe using the latest version is a good choice?

jyemin: sgtm

jyemin (approved):
We can always add .editorconfig later.

LGTM

PR #10, opened 2024-11-07
unblock org.hibernate.Session

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-29

Still a DRAFT PR basing on a feature [branch](https://github.com/mongodb/mongo-hibernate/tree/HIBERNATE-27) on "unblocking org.hibernate.SessionFactory", though it seems to have been stablized.

https://jira.mongodb.org/browse/HIBERNATE-30 is almost done in this PR. Currently only transaction related testing cases are still missing.


.evergreen/static-check.sh
```diff
@@ -11,4 +11,4 @@ source java-config.sh
 echo "mongo-hibernate: static checking ..."
 
 ./gradlew -version
-./gradlew -PxmlReports.enabled=true --info -x test clean check
+./gradlew -PxmlReports.enabled=true --info -x test clean check compileJava
```
NathanQingyangXu, line 14: The above missing gradle task is needed to go about `ErrorProne` and `NullAway` code quality check, both of which happens during `compileJava` phase which has been excluded by `-x test`.

src/main/java/com/mongodb/hibernate/jdbc/AbstractMongoConnection.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public void setNetworkTimeout(Executor executor, int milliseconds) {
+        throw new IllegalStateException("Hibernate won't use this method");
+    }
+
+    @Override
+    public int getNetworkTimeout() {
+        throw new IllegalStateException("Hibernate won't use this method");
+    }
+}
```
NathanQingyangXu, line 167: The above abstract class contains all the `unused` API methods (in non-test code excluding delegator calling) in Hibernate v6.6 (by manual checking in IDE). All the methods really ever used by Hibernate will be left out so `MongoCollection` class needs to implement.

A common pattern for the future JDBC API implementations, as explained in the [package-info.java](https://github.com/mongodb/mongo-hibernate/pull/10/files#diff-5c1f1caf24a8bb65ec214e3bfbc8a6e4a3564946df7cc27668a301d24d2d00e6)

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -0,0 +1,221 @@
...
+    public CallableStatement prepareCall(String mql) {
+        throw new UnsupportedOperationException(
+                "MongoDB's JSON server function has been deprecated in favour of 'aggregate' pipeline");
+    }
+
+    @Override
+    public CallableStatement prepareCall(String mql, int resultSetType, int resultSetConcurrency) {
+        throw new UnsupportedOperationException(
+                "MongoDB's JSON server function has been deprecated in favour of 'aggregate' pipeline");
+    }
```
NathanQingyangXu, line 197: We don't need to support MongoDB's counterpart of SQL's `stored procedure` (as done in Hibernate OGM) for it has been deprecated by aggregate pipeline, as explained at https://www.mongodb.com/resources/products/capabilities/stored-procedures

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -0,0 +1,221 @@
...
+
+    @Override
+    public <T> T unwrap(Class<T> unwrapType) {
+        throw new UnknownUnwrapTypeException(unwrapType);
+    }
+
+    @Override
+    public boolean isWrapperFor(Class<?> unwrapType) {
+        return false;
+    }
```
NathanQingyangXu, line 220: The above two won't be used as `MongoConnectionProvider`; the two `SQLWarning` related methods is used soley in Hibernate's `org.hibernate.engine.jdbc.spi.SqlExceptionHelper` class's `void handleAndClearWarnings()` method as below:
```
public void handleAndClearWarnings(
			Connection connection,
			WarningHandler handler) {
		try {
			if ( logWarnings ) {
				walkWarnings( **connection.getWarnings()**, handler );
			}
		}
		catch (SQLException sqle) {
			// workaround for WebLogic
			LOG.debug( "could not log warnings", sqle );
		}
		try {
			// Sybase fail if we don't do that, sigh...
			**connection.clearWarnings()**;
		}
		catch (SQLException sqle) {
			LOG.debug( "could not clear warnings", sqle );
		}
	}
```
and `org.hibernate.engine.jdbc.connections.internal.DriverManagerConnectionProviderImpl`'s following method:

```
protected Connection releaseConnection(Connection conn) {
			Exception t = null;
			try {
				conn.setAutoCommit( true );
				**conn.clearWarnings()**;
				if ( connectionValidator.isValid( conn ) ) {
					return conn;
				}
```

Seems we don't need to anything.

PR #11, opened 2024-11-07
add static analysis section to README.md

NathanQingyangXu:


README.md
```diff
@@ -18,8 +18,32 @@ Java 17 is the JDK version for development.
...
+$ ./gradlew spotlessApply
+```
+
+#### Code quality check
+
+[Error Prone](https://github.com/tbroyer/gradle-errorprone-plugin) gradle plugin is chosen for Java code qualify analysis during compiling phrase. [NullAway](https://github.com/uber/NullAway) is a Java NPE prevention gradle plugin integrated with Error Prone. [JSpecify](https://jspecify.dev) annotations are used to help NullAway detect potential NPEs.
+
+To check whether either plugin complains, simply run the compiling gradle task:
+```console
+$ ./gradlew compileJava
```
jyemin, line 45: Are you sure you don't have to run `check`?

NathanQingyangXu: I am sure. Maybe `clean` is needed as a precondition but `compileJava` is enough. `check` will include this step plus running test.

But I updated to use `check` instead.

jyemin: No, you're right, we don't want to run the tests here.  If we emulate the driver, we should do something like:

```
-x test -x spotlessApply check testClasses javadoc jar
```

NathanQingyangXu: I just realized `check` won't trigger `ErrorProne` (and `NullAway`) code quality checking for `ErrorProne` is `compileJava` task, which goes beyond the scope of `check`, as shown by the below `dry-run` gradle output:
```
$ ./gradlew -x test clean check --dry-run
:clean SKIPPED
:spotlessInternalRegisterDependencies SKIPPED
:spotlessJava SKIPPED
:spotlessJavaCheck SKIPPED
:spotlessCheck SKIPPED
:check SKIPPED
```
I added a bad case of `NullAway` to test and found the above `check` step won't complain.
Definitively we don't need to specify `-x spotlessApply` for it won't take effect unless it is run manually by end-user. There might be some reason to specify it in java driver project, but I am sure we don't need to exclude it.

I also need to udpate the `static check` evergreen command to include `compileJava` as following:
```
/gradlew -PxmlReports.enabled=true --info -x test clean check compileJava
```

NathanQingyangXu:
> Commented back.

I updated to make the description more accurate and concise. I don't think we need to provide concrete gradle command for end-user should know the basics of gradle usage. As long as we give the relevant gradle tasks, it is good enough. It also makes the section short and to-the-point.

README.md
jyemin: We added `-x spotlessApply` for a good reason in the driver.  We had a release fail due to spotlessApply dirtying the tree, and realized that we hadn't notice it because our static checks were running `spotlessApply` before `spotlessCheck`, fixing the errors before they could be detected.

jyemin: I'm surprised that `check` doesn't depend on `compileJava`.  That seems odd.

jyemin (approved):

README.md
NathanQingyangXu: Traditionally check happens after compilation and usually static analysis is done on bytecode. Only Checker, ErrorProne and Lombok could do magic during compiling phase. They are exceptional, not the norm. Reportedly ErrorProne could lower false positive rate dramatically due to the fact that it has access to source code AST.

PR #12, opened 2024-11-12
add .editorconfig compliant with Palantir

NathanQingyangXu:
I spent quite some time on coming up with a basic .editorconfig compliant with Palantir.

I found Intellij's `Default IDE` Java code style scheme is the closest to Palantir, so I tweak it in IDE to make it closer;

Screenshot 2024-11-12 at 9 37 41 AM

I also tried the .editorconfig file on Chamelion project and found it seems to work fine for majority of the format usages.
However, at least two bugs flushed out for Palantir:

- license content won't be added for package-info.java, as we have found
- importing order can't be configured. Palantir always put static imports to the top, as opposed to putting to the bottom as in most IDEs (including Intellij and this .editorconfig; however, Intellij has another bug that the config in .editorconfig won't take effect as well, :().

This .editorconfig file is the exporting result after tweaking in IDE.

Also fixed some minor verbiage issues in README.md.


jyemin (approved):
I haven't tried it, but if it works well enough, then let's use it.

PR #15, opened 2024-11-14
programmatically customize MongoClient building

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-28

This PR aims to provide the most flexible programmatic MongoClient building, based on Hibernate's Service mechanism (see https://docs.jboss.org/hibernate/orm/6.6/integrationguide/html_single/Hibernate_Integration_Guide.html), particularly the [Custom Service Roles (extending)](https://docs.jboss.org/hibernate/orm/6.6/integrationguide/html_single/Hibernate_Integration_Guide.html#services-extending) section.


src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -73,13 +77,16 @@
  * @see JdbcSettings#JAKARTA_JDBC_PASSWORD
  * @see <a href="https://www.mongodb.com/docs/manual/reference/connection-string/">connection string</a>
  */
-public final class MongoConnectionProvider implements ConnectionProvider, Configurable, Stoppable {
+public final class MongoConnectionProvider
```
jyemin, line 80: Why is this reformatted?  I thought introducing Spotless/Palantir would avoid these sorts of changes.

src/test/java/com/mongodb/hibernate/MongoClientCustomizerTests.java
```diff
@@ -0,0 +1,70 @@
...
+    private void buildSessionFactory(MongoClientCustomizer mongoClientCustomizer) {
+        var settings = Map.<String, Object>of(
+                AvailableSettings.JAKARTA_JDBC_URL, "mongodb://localhost/test",
+                AvailableSettings.DIALECT, MongoDialect.class.getName(),
+                AvailableSettings.CONNECTION_PROVIDER, MongoConnectionProvider.class.getName());
+        var standardServiceRegistry = new StandardServiceRegistryBuilder()
+                .applySettings(settings)
+                .addService(MongoClientCustomizer.class, mongoClientCustomizer)
+                .build();
+        var sessionFactoryBuilder = new MetadataSources(standardServiceRegistry)
```
jyemin, line 62: Any reason why this code looks a bit different than the code in `MongoClientCustomizer` Javadoc?

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -134,8 +141,10 @@ public void configure(Map<String, Object> configValues) {
             throw new NotYetImplementedException("To be implemented after auth could be tested in CI");
         }
 
-        var clientSettings = clientSettingsBuilder.build();
-        this.mongoClient = MongoClients.create(clientSettings);
+        if (mongoClientCustomizer != null) {
+            mongoClientCustomizer.customize(clientSettingsBuilder);
```
jyemin, line 145: Note that the tests all pass even if we change this line to:

```
mongoClientCustomizer.customize(MongoClientSettings.builder()
```

But I don't see a straightforward way yet to test that the customization is actually applied to the `MongoClient`

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: It is because a new interface was added to the `implements` list, ending up with the line exceeding 120 chars real estate, so Spotless/Palantir has to break it down into two lines.

src/test/java/com/mongodb/hibernate/MongoClientCustomizerTests.java
NathanQingyangXu: Good catch! I changed it to be consistent with the usage in Javadoc.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: It is a little bit hard to verify `MongoClient`, but I think it should be done. I could create a pure unit testing case focusing on the `MongoConnectionProvider#configure(Map<String,Object>)` testing, but even in that case we still need to expose the internal `MongoClient` for testing purposes (like creating a package private access method).
Will explore to see whether there is better way. Good point!

NathanQingyangXu: I found Mockito introduced the long-awaited `static mocking` feature since `3.4.0` (see https://www.baeldung.com/mockito-mock-static-methods for a quick tutorial). So I enriched the existing testing case by exploiting this new feature.

src/test/java/com/mongodb/hibernate/MongoClientCustomizerTests.java
NathanQingyangXu: The reason was copying the existing pattern from `SessionFactoryTests` which needs to customize the connection string so using `hibernate.properties` then updating it is really ugly. But `SessionFactoryTests` is special and we should revert back to normal `hibernate.properties` usage.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
jyemin: oh, totally missed that!  Sorry.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
jyemin: I'm not sure if this sort of mocking test is a good idea, as it only finds certain manifestations of the issue we're trying to test.  For example, if you change the provider to:

```
//        var clientSettingsBuilder = MongoClientSettings.builder().applyConnectionString(connectionString);
//
//        if (configValues.get(JAKARTA_JDBC_USER) != null || configValues.get(JAKARTA_JDBC_PASSWORD) != null) {
//            throw new NotYetImplementedException("To be implemented after auth could be tested in CI");
//        }

        if (mongoClientCustomizer != null) {
            MongoClientSettings.Builder builder = MongoClientSettings.builder();
            mongoClientCustomizer.customize(builder);
            builder.build();
        }
        this.mongoClient = MongoClients.create(connectionString);
```

this test will still pass.  And this is why I don't love using mocks!  Let's see if @stIncMale has an opinion.

jyemin (commented):
Added a few comments

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: BTW, I just want to confirm. So if user opts for the programmatic customization, should he/she control everything including connection string and credential? I found the above usage makes some sense as well.

I agree the mock is not the best idea (for one thing, it is convoluted and difficult to understand). I think better alternative is to create a separate pure unit testing class to test `MongoConnectionProvider#configure(Map<String, Object>)`. Then it has still difficulty for it seems the classes involved are pretty "closed" defying testing.

NathanQingyangXu: I think the usage of Mock is too bad. I re-implement the testing case without the usage of Mock and hopefully this time the testing is more robust (covering Jeff's two bad cases).

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+ * serviceRegistryBuilder.addService(MongoClientCustomizer.class, clientCustomizer);
+ *
+ * var sessionFactory = cfg.buildSessionFactory();
+ *
+ * // start using sessionFactory as normal
+ * ...
+ *
+ * </pre>
+ *
+ * <p>The injected {@code clientCustomizer} will be fetched and do its due diligence at later stage of Hibernate
```
jyemin, line 54: Can you reword this to be more precise? The phrase "do its due diligence" is not particularly helpful.  It sounds more like something a lawyer would do ;)

jyemin (changes requested):
One change request on the Javadoc, but also holding off on LGTM because I know @stIncMale has a concern that he will express about how the `MongoClientCustomizer` API interacts with the `MongoClientSettings.Builder` API.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
NathanQingyangXu: updated. Let me know if it needs improvement further. English writing is not my strong point and I always welcome improvement suggestion.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package com.mongodb.hibernate.service;
+
+import com.mongodb.MongoClientSettings;
+import org.hibernate.service.Service;
+
+/**
+ * A Hibernate {@link Service} focusing on customizing {@link com.mongodb.client.MongoClient} creation programmatically
+ * by invoking the various methods of {@link MongoClientSettings.Builder} before its
```
stIncMale, line 24:
```suggestion
 * by invoking various methods of {@link MongoClientSettings.Builder} before its
```

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * var serviceRegistryBuilder = cfg.getStandardServiceRegistryBuilder();
+ * serviceRegistryBuilder.addService(MongoClientCustomizer.class, clientCustomizer);
+ *
+ * var sessionFactory = cfg.buildSessionFactory();
+ *
+ * // start using sessionFactory as normal
+ * ...
+ *
+ * </pre>
```
stIncMale, line 52: Until we have access to [JEP 467: Markdown Documentation Comments](https://openjdk.org/jeps/467), the correct way to mark multi-line code blocks is with

```
<pre>{@code
...
}</pre>
```

```suggestion
 * <pre>{@code
 * var cfg = new Configuration();
 *
 * // configure cfg as you normally do (e.g. add entity classes)
 * ...
 *
 * // connection string, user and password have been applied to the builder already
 * var clientCustomizer = builder -> {
 *     // customize client settings to your heart's content
 *     ...
 * };
 *
 * var serviceRegistryBuilder = cfg.getStandardServiceRegistryBuilder();
 * serviceRegistryBuilder.addService(MongoClientCustomizer.class, clientCustomizer);
 *
 * var sessionFactory = cfg.buildSessionFactory();
 *
 * // start using sessionFactory as normal
 * ...
 * }</pre>
```

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>The injected {@code clientCustomizer} service will be applied at later stage (in
+ * {@link com.mongodb.hibernate.jdbc.MongoConnectionProvider}) of Hibernate bootstrapping.
+ *
+ * @see com.mongodb.hibernate.jdbc.MongoConnectionProvider#configure(java.util.Map)
+ */
+@FunctionalInterface
+public interface MongoClientCustomizer extends Service {
+    void customize(MongoClientSettings.Builder builder);
```
stIncMale, line 61: `ConnectionString` and `MongoClientSettings` conflict with each other, because they are different ways of configuring mostly the same settings. The Java driver avoids coming up with rules on how conflicts are resolved by shifting the task of conflict resolution to an application: an application has to call `MongoClientSettings.Builder.applyConnectionString`, and has a choice to configure `MongoClientSettings.Builder` before this call or after (or even both).

Let's use the same approach here by changing the method to
```java
void customize(MongoClientSettings.Builder mongoClientSettingsBuilder, @Nullable ConnectionString connectionString)
```

This way `MongoClientCustomizer` will be able to decide when `applyConnectionString` should be called, if at all. Also, you may see that the `connectionString` argument is `@Nullable`. I think, we should make `JdbcSettings.JAKARTA_JDBC_USER` optional, because if someone wants to configure `MongoClient` fully programmatically, we must not interfere with that by demanding a connection string to be specified.

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
NathanQingyangXu: Too many options. We'd better finalize the tech design prior to implementation?

For the second option accepting a map, why bother with passing them from `hibernate.property` intermediary (I assume most of the map entries are irrelevant), instead of letting user build a client from scratch in whatever way he/she wants? This is a programattic way user has maximal flexibility, right?

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+import com.mongodb.MongoClientSettings;
+import org.hibernate.service.Service;
+
+/**
+ * A Hibernate {@link Service} focusing on customizing {@link com.mongodb.client.MongoClient} creation programmatically
+ * by invoking the various methods of {@link MongoClientSettings.Builder} before its
+ * {@link MongoClientSettings.Builder#build()} invocation.
+ *
+ * <p>Note that this service is only needed when normal connection string, user and password configurations are not
+ * enough.
```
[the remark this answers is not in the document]
NathanQingyangXu: There are many grey areas if I am the first time user, including:

- whether I need it if I have specified connection string, user and password?
- if I use this service, should I assume I don't need to configure connection string, user and password again?

When I wrote the Javadoc, I felt it challenging to make everything crystal clear.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
stIncMale: > if we do allow user/password (or anything else for that matter) to be specified via configuration, how would those values get to the customizer?

A setting that can be specified via a connection string should not be allowed to be specified via Hibernate ORM configuration properties. That way we won't have the problem you pointed out. I fail to see a single benefit of allowing connection string options to be also specified via Hibernate ORM configuration properties. If someone sees those benefits, could they be stated explicitly?

> We'd better finalize the tech design prior to implementation?

That's why when I created [HIBERNATE-38 Design, implement and document how an application can configure the product](https://jira.mongodb.org/browse/HIBERNATE-38), I specified it as a dependency of https://jira.mongodb.org/browse/HIBERNATE-28 (the task implemented by the current PR).

NathanQingyangXu: Let me provide some historical context. The decision to make use of the following 3 properties as normal way to bootstrap our product was made by @jyemin:

- jakarta.persistence.jdbc.url
- jakarta.persistence.jdbc.user
- jakarta.persistence.jdbc.password

I want to point out that though the above JPA properties are preferred, Hibernate also has its own legacy property counterparts, including:

- hibernate. connection. url
- hibernate. connection. username
- hibernate. connection. password

together with the other multi-db related default catalog and schema.
So a natural question is if our user used the properties from `Environment`, they migtht be surprised that they are not taking effect at all. If we chose `default catalog` and `default schema`, they are even more surprised that they are supported, :).

My personal preference is to avoid them (they are for traditional scenario that a JDBC Driver is provided together with them (another `jakarta.persistence.jdbc.driver` is usually provided together with above three; but our case is definitively special in the sense that we have no JDBC Driver to provide) and choose our own properties (like Spring Data MongoDB's config) so end-user would have crytal clear expectation how to use our product.

From my understanding, if we wanna avoid the multi-db complexity (it is complicated for in native MQL scenario, a default database is still needed, and tbh, I don't think reusing the confusing `catalog` and `schema` is a great product idea in the first place), our product is only expecting two things and nothing more:

- give me a `MongoClient`
- give me a mongo database name

If so, a simplistic tech design is simply:

- introduce a new property (mandatory) like `mongodb.client`, which will accept a working `MongoClient` prepared by end user
- introduce a new property (mandatory) like `mongodb.database`, which denotes the database name

I think even `Service` is not necessary and adding the `mongoClient` into a settings map is perfect fine for now (just the original idea @jyemin previously proposed; in the future when we need to consider using our product in JavaEE environment where it is not easy to insert ad-hoc code to create a `MongoClient`, we could switch to `Service` mechanism).

NathanQingyangXu: > I like this idea, but if we do allow user/password (or anything else for that matter) to be specified via configuration, how would those values get to the customizer?

That is straightforward for we have gotten a mandatory `ConnectionString` already (in `MongoConnectionProvider`). We only need to tweak the logic to make it optional in event of customizer is found.

NathanQingyangXu: > That's why when I created [HIBERNATE-38 Design, implement and document how an application can configure the product](https://jira.mongodb.org/browse/HIBERNATE-38), I specified it as a dependency of https://jira.mongodb.org/browse/HIBERNATE-28 (the task implemented by the current PR).

I am ok with the discussion. My only concern is it might be a lenghy process and might interfere with Milestone 1's progress (if the resulting conclusion is easy to implement, that is also fine). We are still at the very first ticket, but we have found many reasons to digress. Sometimes sticking to a simple solution pays off (no matter whther it is perfect, as long as it could be rectified later), and that is why we spent so much time on tech doc writing. 

Another option is to move the discuss to Milestone 2 and stick to the current approach based on the three JPA Hibernate properties? I am even more versed in the details and complexity, but as I explained, we'd better focus.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -0,0 +1,65 @@
...
+@FunctionalInterface
+public interface MongoClientCustomizer extends Service {
+
+    /**
+     * Customize {@link MongoClientSettings} building.
+     *
+     * @param builder a {@link MongoClientSettings.Builder} instance which has been created with blank state (without
+     *     applying the connectionString parameter if provided)
+     * @param connectionString the {@link ConnectionString} created from {@value JdbcSettings#JAKARTA_JDBC_URL}
+     *     configuration if provided; provided for reference alone; could be null
```
stIncMale, line 61: The documentation of Hibernate ORM uses multiple terms to refer to "configuration property" (the thing that has name, e.g., `jakarta.persistence.jdbc.url`, and value):

- "configuration setting", see https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#settings
- just "setting", see _"The settings ..."_ in https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/cfg/AvailableSettings.html.
- "configuration property", https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#envers-configuration
- just "property", see the `property` XML element in `hibernate.cfg.xml`.

Such pointless term overloading is not good, but then on top of that, our Hibernate dialect refers to those things as just "configuration"<sup>1</sup>. I propose us to pick a single term of the aforementioned (preferably, "configuration property", as it does not conflict with `MongoClientSettings`), and stick to it, until we have too large of a codebase to fix.

<sup>1</sup> Using the term "configuration" to refer to a single configuration property is especially bad because it conflicts with `org.hibernate.cfg.Configuration`, and because "configuration" is a collective term, not usually used to refer to a single configuration property.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -110,38 +100,53 @@ public <T> T unwrap(Class<T> unwrapType) {
     @Override
     public void configure(Map<String, Object> configValues) {
         var jdbcUrl = configValues.get(JAKARTA_JDBC_URL);
-        if (jdbcUrl == null) {
-            throw new HibernateException("Configuration is required: " + JAKARTA_JDBC_URL);
+
+        if (mongoClientCustomizer == null && jdbcUrl == null) {
+            throw new HibernateException(format(
+                    "Configuration [%s] is required unless MongoClientCustomizer is provided", JAKARTA_JDBC_URL));
```
stIncMale, line 106: Instead of hardcoding the interface name in the message, let's use the `.class` reference. This way this usage of the interface is explicit and visible to the compiler.

```suggestion
                    "Configuration [%s] is required unless %s is provided",
                    JAKARTA_JDBC_URL, MongoClientCustomizer.class.getName()));
```

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -110,38 +100,53 @@ public <T> T unwrap(Class<T> unwrapType) {
...
+        if (mongoClientCustomizer != null) {
+            clientSettingsBuilder = MongoClientSettings.builder();
+            mongoClientCustomizer.customize(clientSettingsBuilder, connectionString);
+        } else {
+            clientSettingsBuilder = MongoClientSettings.builder().applyConnectionString(connectionString);
+        }
+        mongoClient = MongoClients.create(clientSettingsBuilder.build());
+    }
+
+    private ConnectionString getConnectionString(Object jdbcUrl) {
```
stIncMale, line 121:
```suggestion
    private static ConnectionString getConnectionString(Object jdbcUrl) {
```

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -34,52 +36,40 @@
...
- *         <td>{@code userName} for {@link com.mongodb.MongoCredential#createCredential(String, String, char[])}</td>
- *         <td></td>
- *     </tr>
- *     <tr>
- *         <td>{@value JdbcSettings#JAKARTA_JDBC_PASSWORD}</td>
- *         <td>{@code password} for {@link com.mongodb.MongoCredential#createCredential(String, String, char[])}</td>
- *         <td></td>
- *     </tr>
- * </table>
+ * <p>{@link MongoConnectionProvider} uses the {@value JdbcSettings#JAKARTA_JDBC_URL} to create a MongoDB <a
```
stIncMale, line 51: [optional]

I believe, it should be either

```java
uses the {@value JdbcSettings#JAKARTA_JDBC_URL} configuration property to create
```
or
```java
uses {@value JdbcSettings#JAKARTA_JDBC_URL} to create
```
but not `the {@value JdbcSettings#JAKARTA_JDBC_URL}`.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -34,52 +36,40 @@
...
+ * href="https://www.mongodb.com/docs/manual/reference/connection-string/">connection string</a>. The configuration is
+ * mandatory if {@link MongoClientCustomizer} service is not used. Otherwise, a {@link ConnectionString} instance will
+ * be provided to {@link MongoClientCustomizer#customize(MongoClientSettings.Builder, ConnectionString)} method as a
```
stIncMale, line 54: [optional]

I believe, it should be either

```
provided to the {@link MongoClientCustomizer#customize(MongoClientSettings.Builder, ConnectionString)} method as a
```
or
```
provided to {@link MongoClientCustomizer#customize(MongoClientSettings.Builder, ConnectionString)} as a
```

stIncMale (changes requested):

stIncMale:
The last reviewed comment: 6fe6fb13921c51b2393f40267408dc178c55d6c1.

src/test/java/com/mongodb/hibernate/jdbc/MongoClientCustomizerTests.java
```diff
@@ -69,13 +68,7 @@ void testMongoClientCustomizerThrowException() {
...
         var cfg = new Configuration();
-        var standardServiceRegistry = cfg.getStandardServiceRegistryBuilder()
-                .addService(MongoClientCustomizer.class, mongoClientCustomizer)
-                .build();
-        var sessionFactoryBuilder = new MetadataSources(standardServiceRegistry)
-                .getMetadataBuilder()
-                .build()
-                .getSessionFactoryBuilder();
-        return sessionFactoryBuilder.build();
+        cfg.getStandardServiceRegistryBuilder().addService(MongoClientCustomizer.class, mongoClientCustomizer);
```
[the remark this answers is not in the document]
NathanQingyangXu: Good catch!! I didn't notice that. Seems we have to abandon the `Configuration` then. You are totally correct. We should use the way in `SessionFactoryTests.buildSessionFactory()`. It is also the recommended way in Hibernate User Guide at https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#bootstrap-native

Updated. Tons of thanks to you.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
NathanQingyangXu: Good point! Let us use `configuration property` then. Updated. I also struggled whether to use `configValues` or `configurationValues` for the `Map<String, Object>` parameter in `MongoConnectionProvider#configure()`. Seems `configurationValues` are more common, but `configValues` shows up as well.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: of course you are right. I chose the first option.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: yeah, I chose the first option above

src/test/java/com/mongodb/hibernate/jdbc/MongoClientCustomizerTests.java
```diff
@@ -0,0 +1,75 @@
...
+import com.mongodb.hibernate.service.MongoClientCustomizer;
+import java.util.UUID;
+import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
+import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.service.spi.ServiceException;
+import org.junit.jupiter.api.Test;
+
+class MongoClientCustomizerTests {
```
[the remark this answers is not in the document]
NathanQingyangXu: added by using `@ParameterizedTest`. Let me know if you have better idea.

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
stIncMale: There are multiple other places where this change applies to, but they were left unchanged.

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -30,20 +30,20 @@
...
- *
  * var clientCustomizer = (clientSettingsBuilder, connectionString) -> {
  *     // customize client settings to your heart's content
  *     ...
  * };
  *
- * var serviceRegistryBuilder = cfg.getStandardServiceRegistryBuilder();
- * serviceRegistryBuilder.addService(MongoClientCustomizer.class, clientCustomizer);
+ * var standardServiceRegistryBuilder = new StandardServiceRegistryBuilder();
+ * standardServiceRegistryBuilder.addService(MongoClientCustomizer.class, clientCustomizer);
```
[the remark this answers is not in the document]
NathanQingyangXu: good catch. The only issue is the line would be very long. I broke the long line into two as per the convention of Spotless/Palantir.

stIncMale (dismissed):

NathanQingyangXu:
@stIncMale , I need to solve conflict with main (another PR got merged previously) and then after committing the merging, I found stack-check is broken, so I committed again to remove unused import. Now this PR requires your approval again. thx

PR #16, opened 2024-11-15
Bump Hibernate ORM dependency version

NathanQingyangXu:
from v6.6.1.Final to v6.6.2.Final with more than 50 bugfix related commits (https://hibernate.atlassian.net/issues/?jql=project%20%3D%20HHH%20AND%20fixVersion%20in%20%286.6.2%29%20ORDER%20BY%20updated)

also renamed the spotless license file to avoid confusion (original `copyright.txt` would make people assume it is the generic stuff, but it is spotless-specific; the naming is copied from Hibernate ORM (https://github.com/hibernate/hibernate-orm/blob/main/spotless.license.java). 

Better naming suggesiton is welcome!


README.md
```diff
@@ -19,6 +19,8 @@ Initially Hibernate ORM v6.6 is the dependency version.
 
 MongoDB v6 is the minimal version this dialect supports.
 
+> [Standalone instance](https://www.mongodb.com/docs/v4.4/reference/glossary/#std-term-standalone) will not be supported. It is recommended to [convert it to a replica set](https://www.mongodb.com/docs/manual/tutorial/convert-standalone-to-replica-set/).
```
stIncMale, line 22: We should link to the current version of the docs:

```suggestion
> [Standalone instance](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-standalone) will not be supported. It is recommended to [convert it to a replica set](https://www.mongodb.com/docs/manual/tutorial/convert-standalone-to-replica-set/).
```

stIncMale (changes requested):

.editorconfig
```diff
@@ -149,7 +149,7 @@ ij_java_method_parameters_right_paren_on_new_line = false
 ij_java_method_parameters_wrap = off
 ij_java_modifier_list_wrap = false
 ij_java_multi_catch_types_wrap = normal
-ij_java_names_count_to_use_import_on_demand = 3
+ij_java_names_count_to_use_import_on_demand = 30
```
NathanQingyangXu, line 152: the above changes will end up with solving the annoying `import static wildcard` issue which troubled me all the time.
Spotless decides to ignore such common issue due to backward compatibility: https://github.com/diffplug/spotless/issues/649, unfortunately

stIncMale (approved):

PR #17, opened 2024-11-25
transaction demarcation

NathanQingyangXu:
After finishing implementing `java.sql.Connection`, the natural followup is to implement `java.sql.Transaction`.

Note that Hibernate supports both JTA managed transaction and "resource local" pure JDBC transaction, and only the latter applies to our case.

so whenever a Hibernate conceptual transaction is needed to start (not the same as physical transaction, which is usually started lazily when real query is issued), the following JDBC statement should be invoked:

```
var conn = // some jdbc Connection
conn.setAutoCommit(false); // previous `autoCommit` state should be true
```

Also note that we had decided not to support JDBC transaction isolation level.

[HIBERNATE-30](https://jira.mongodb.org/browse/HIBERNATE-30)


src/test/java/com/mongodb/hibernate/SessionFactoryTests.java
```diff
@@ -50,22 +52,33 @@ void testInvalidConnectionString() {
...
+            assertDoesNotThrow(() -> sessionFactory.openSession().close());
         }
     }
 
+    @Test
+    void testInTransaction() {
+        try (var sessionFactory = buildSessionFactory()) {
+            sessionFactory.inTransaction(session -> {});
+        }
+    }
```
NathanQingyangXu, line 65: some code refactoring was done in existing `SessionFactoryTests` so no need of empty map to be passed in default scenario.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -16,56 +16,297 @@
...
+            @ValueSource(booleans = {true, false})
+            void testSQLExceptionThrowWhenCalledOnClosedConnection(boolean autoCommit) throws SQLException {
+                // given
+                mongoConnection.close();
+                verify(clientSession).close();
+
+                // when && then
+                assertThrows(SQLException.class, () -> mongoConnection.setAutoCommit(autoCommit));
+                verifyNoMoreInteractions(clientSession);
+            }
```
stIncMale, line 187: Let's use the same approach when we call `verify(clientSession).close()` and `verifyNoMoreInteractions(clientSession)` in other similar tests:

- `TransactionIsolationLevelTests.testSQLExceptionThrownWhenConnectionClosed`
- `RollbackTests.testSQLExceptionThrownWhenConnectionClosedAlready`
- `CommitTests.testSQLExceptionThrownWhenConnectionClosedAlready`

Also, it seems that the method names of these tests should be the same, which is not the case. Descriptions should be similar, but currently, some of these tests have them, while others don't have them at all.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -16,56 +16,297 @@
...
+                // then
+                verifyNoMoreInteractions(clientSession);
+            }
+
+            @Nested
+            class AutoCommitStateChangedTests {
+
+                @ParameterizedTest(name = "AutoCommit (true -> false): start transaction (successful: {0})")
+                @ValueSource(booleans = {true, false})
+                void testTringToStartTransactionWhenAutoCommitChangedToFalse(boolean successful) throws SQLException {
```
stIncMale, line 116: Tring -> Trying

```suggestion
                void testTryingToStartTransactionWhenAutoCommitChangedToFalse(boolean successful) throws SQLException {
```

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -16,56 +16,297 @@
...
+                    verify(clientSession).commitTransaction();
+
+                    assertFalse(mongoConnection.getAutoCommit());
+
+                    // when
+                    mongoConnection.setAutoCommit(true);
+
+                    // then
+                    assertTrue(mongoConnection.getAutoCommit());
+                    verify(clientSession, times(2)).hasActiveTransaction();
```
stIncMale, line 171: Given that we don't actually care about `hasActiveTransaction` being called twice, let's make the assertion less strict and, therefore, less prone to rising false alarms:

```suggestion
                    verify(clientSession, atLeast(1)).hasActiveTransaction();
```

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -56,43 +59,67 @@ final class MongoConnection extends ConnectionAdapter {
     @Override
     public void setAutoCommit(boolean autoCommit) throws SQLException {
         checkClosed();
-        throw new NotYetImplementedException(
-                "To be implemented in scope of https://jira.mongodb.org/browse/HIBERNATE-30");
+        if (autoCommit != this.autoCommit) {
```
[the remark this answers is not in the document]
NathanQingyangXu: yeah, this is a clean approach advocated in `Clean Code` book as well.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -56,43 +59,67 @@ final class MongoConnection extends ConnectionAdapter {
...
+                throw new SQLException("Failed to commit transaction", e);
+            }
+            try {
+                if (!autoCommit) {
+                    clientSession.startTransaction();
+                }
+                this.autoCommit = autoCommit;
+            } catch (RuntimeException e) {
+                throw new SQLException("Failed to start transaction", e);
+            }
```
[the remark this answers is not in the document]
NathanQingyangXu: that is a good point. The concern here is Hibernte ORM has dictated the transaction pattern in a framework way so we don't have luxurary to go about customization as PostgreSQL. Let me explore its code logic to see whether there is some way out.

From my current understanding, Hibernate ORM (it has a `transaction` root package in its codebase), it supports two ways: 1) JTA; 2) resource local or JDBC driver way (as per point No.2 above). It doesn't support deep customization for JTA mode, but that is out of question for our JDBC emulation approach.

We can discuss later after I read through the relevant Hibernate code logic.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -56,43 +59,67 @@ final class MongoConnection extends ConnectionAdapter {
...
-        throw new NotYetImplementedException(
-                "To be implemented in scope of https://jira.mongodb.org/browse/HIBERNATE-30");
+        if (autoCommit != this.autoCommit) {
+            try {
+                if (clientSession.hasActiveTransaction()) {
+                    clientSession.commitTransaction();
+                }
+            } catch (RuntimeException e) {
+                throw new SQLException("Failed to commit transaction", e);
+            }
```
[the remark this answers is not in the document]
NathanQingyangXu: One issue with `assertFalse()` is it will throw `Error`, not `Exception`, so it should only used in fatal error case wherein Hibernate will give up totally; but if it is not fatal, Hibernate might handle elegantly if needed for `SQLException` is the contract.

For our current design (no connection pool and no JTA or any transaction customization usage), `assertFalse()` might be fine, but I want to be cautious here.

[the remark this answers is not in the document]
NathanQingyangXu: oh, saw we have the precondition of `doCommit()` that `autoCommit` is false, so I think the `assertFalse()` makes sense.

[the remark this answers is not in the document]
NathanQingyangXu: well, there could be still edge case that `clientSession.hasActiveTransaction()` is true but `autoCommit` is true. but let us use `assertFalse()` for now.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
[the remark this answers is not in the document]
NathanQingyangXu: JDBC has no explicit transaction beginning statement, and yeah, `setAutoCommit(false)` is the precondition to start a transaction implicitly (it is up to JDBC Driver to do some optimization as PostgreSQL). We are emulating JDBC flow using Mongo Java driver and live in the tight jacket of Hibernate ORM framework. Maybe it is a good perf opt ticket in the future (if it is really worthwhile), but for now I guess we have limited choice here.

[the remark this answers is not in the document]
NathanQingyangXu: So here is the relevant Hibernate User Guide: https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#transactions-api
Screenshot 2024-12-05 at 2 23 55 PM

The relevant code logic is at https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/AbstractLogicalConnectionImplementor.java#L69

So it seems Hibernate transaction demarcation is purely based on `setAutoCommit(false)`. Particularly, the transaction API will set its state based on the assumption that once `setAutoCommit(false)` is called, the transaction state will be `ACTIVE`. For smooth Hibernate integration, we'd better align with Hibernate's expectation.

So the only question outstanding is:

- could we delay `clientSession.startTransaction()` lazily when the first `java.sql.Statement` invocation is called?

Well, it seems doable (if so, we could hold off the review of this PR until we finished the whole JDBC flow, that is fine), but it might bring about lots of complexity, including:

- Hibernate integration might be broken and we need to fix all kinds of edge case (like some exception is thrown from current session prior to first statement issuing or delayed transaction starting, then Hibernate will try to rollback its transaction and hopefully it won't trigger some of our internal SQLException or even worse, some AssertError)
- unit testing and integration testing will be challenging for we need to take into consideration all the lazy initialization corner cases

Another option is to take a simple approach without emulating PostgreSQL and conduct perf optimization after milestone No. 1. I am afraid if we implement the lazy demarcation idea, milestone will be postponed again and again.

How do you think?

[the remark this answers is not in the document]
NathanQingyangXu: I think you are right. `setAutoCommit(false)` should not trigger transaction starting. It should be done lazily during `Statement` or `PreparedStatement` executing.
Updated the PR. Tons of thanks to your good catch!

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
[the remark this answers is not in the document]
stIncMale: > there could be still edge case that clientSession.hasActiveTransaction() is true but autoCommit is true

You are right, a situation when an auto-committing connection has an active transaction, and `setAutoCommit(false)` is called, seems fine, yet the `assertFalse(autoCommit)` check in `doCommit` fails in such a situation. Let's remove the check.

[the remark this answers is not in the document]
NathanQingyangXu: thx. removed.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -94,6 +87,9 @@ public void commit() throws SQLException {
 
     private void doCommit() throws SQLException {
         assertFalse(autoCommit);
+        if (!clientSession.hasActiveTransaction()) {
+            return;
```
stIncMale, line 88: If we do this, then we should remove the `clientSession.hasActiveTransaction()` check from `setAutoCommit`. However, I am not sure we should do such a change (the same goes about the similar change in `rollback`): `com.mongodb.client.ClientSession.commitTransaction`/`abortTransaction` throw `IllegalStateException` if the `TransactionState` is anything but `IN`. I think, we should preserve that behavior (as it was done before this change), instead of pretending that we have successfully committed / rolled back in situations when we did not commit / roll back anything.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -111,24 +107,6 @@ void testNoOpWhenAutoCommitNotChanged(boolean autoCommit) throws Exception {
             @Nested
             class AutoCommitStateChangedTests {
 
-                @ParameterizedTest(name = "AutoCommit (true -> false): start transaction (successful: {0})")
-                @ValueSource(booleans = {true, false})
-                void testTryingToStartTransactionWhenAutoCommitChangedToFalse(boolean successful) throws SQLException {
```
stIncMale: Let's add a test instead of `testTryingToStartTransactionWhenAutoCommitChangedToFalse`, which verifies that `setAutoCommit` does not start a transaction when the auto-commit state changes?

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
NathanQingyangXu: This is necessary to work around Hibernate's conceptual transaction mechanism. If we don't do this, the following testing case in `com.mongodb.hibernate.SessionTests` will fail:

```
    @Test
    void testBeginTransaction() {
        assertDoesNotThrow(() -> session.beginTransaction().commit());
    }
```
for the simple reason that physical transaction has not been started. This is a dilemma due to discrepancy between Hibernate conceptual transaction (started by calling `setAutoCommit(false)`) and physical transaction (which is started when first statement is issued).

My biggest concern is if we didn't do this, we might face more trouble during Hibernate integration. Also, I checked existing h2, postgreSQL JDBC driver implementation, they all allow for such behaviour; JDBC spec doesn't disallow it as well.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
NathanQingyangXu: Good point! Added. Thanks.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
NathanQingyangXu: I checked a couple of existing JDBC implementations (including PostgreSQL, H2, SQL Server and MyriaDB). All of them are based on the logic that no-op if no transaction exists.

stIncMale: Thank you, I agree.

Let's then

1. either remove the `clientSession.hasActiveTransaction()` check from `setAutoCommit`, and rename the `doCommit` to something like `doCommitIfNeeded`,
2. or add a similar check to `commit`, and replace the `clientSession.hasActiveTransaction()` check in `doCommit` with `assertTrue(clientSession.hasActiveTransaction())`.

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
NathanQingyangXu: Good catch again! I adopted your first suggestion.

PR #20, opened 2024-11-29
Minimal Mongo AST

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-31


NathanQingyangXu:
I didn't include `AstNodeType` for it is not used even in POC project. Will add them back if needed

jyemin:
> I didn't include AstNodeType for it is not used even in POC project. Will add them back if needed

I included it in POC because it was used in LINQ, though I don't recall off hand what it was used for.

jyemin:
@NathanQingyangXu should we hold off on this and instead add AST nodes as needed to fully implement features?

NathanQingyangXu:
> @NathanQingyangXu should we hold off on this and instead add AST nodes as needed to fully implement features?

Yeah, I think that is more manageable and verifiable. I think we could close this ticket and plan to add AST node rendering incrementally in each of CRUD tickets.

PR #26, opened 2025-01-27
basic insertion

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-14

The first CRUD component implementation PR (choosing insertion as the first step due to the data dependency for the other components requires insertion as pre-condition; though Java driver-based native insertion is also fine). After this PR is merged, the other `update`, `delete` and `load` PR could be implemented in parallel. This PR will lay down important infrastructure and would streamline the other three.

As in POC project, mongoast stuff (inspired by C# mongo driver's Linq: https://github.com/mongodb/mongo-csharp-driver/tree/main/src/MongoDB.Driver/Linq) was used as elegant translation result of SQL AST scanning. Only insertion related mongoast stuff was included and incrementally introducing more mongoast stuff will be an important part of the later CRUD PRs.

Hibernate relies on `SqlAstWalker` as the SQL AST scanning abstraction based on well-known visitor pattern. In origiinal SQL translation scenario, string concatenation seems to suffice so the visitor walker interface's all visitor methods don't return any value (using `void` as the return type); but given our internal mongoast generation implmenetation plan, it is natural to return value from the visitor methods. So a tricky data exchange mechanism was introduced (previously we called it `Attachment` though it is misnomer). Type safety is a core concern but I think what really matters is the type, including generics info (which would be erased by Java at runtime; but there is well-known solution to retain generics info like Guava's TypeToken: https://www.baeldung.com/guava-reflection). Another tricky part is the data exchange mechanism shares something with Stack in the sense that once its value is fetched, its previous state should be restored. The necessity could only be shown by code debugging when visiting SQL AST tree.

### DIFF
--- a/src/main/java/com/mongodb/hibernate/translate/AstVisitorValueHolder.java
+++ b/src/main/java/com/mongodb/hibernate/translate/AstVisitorValueHolder.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2024-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.translate;
+
+import static com.mongodb.hibernate.internal.MongoAssertions.assertNotNull;
+import static com.mongodb.hibernate.internal.MongoAssertions.assertNull;
+import static com.mongodb.hibernate.internal.MongoAssertions.assertTrue;
+import static java.lang.String.format;
+
+import com.google.common.reflect.TypeToken;
+import com.mongodb.hibernate.translate.mongoast.AstNode;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * A data exchange mechanism to overcome the limitation of various visitor methods in
+ * {@link org.hibernate.sql.ast.SqlAstWalker} don't return value; Returning values is required during MQL translation
+ * (e.g. returning intermediate MQL {@link AstNode}).
+ *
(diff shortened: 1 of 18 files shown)


src/main/java/com/mongodb/hibernate/translate/MqlTranslator.java
```diff
@@ -0,0 +1,595 @@
...
+
+    @Override
+    @SuppressWarnings({"unchecked"})
+    public T translate(JdbcParameterBindings jdbcParameterBindings, QueryOptions queryOptions) {
+        if (statement instanceof TableMutation) {
+            TableMutation<T> tableMutation = (TableMutation<T>) statement;
+            if (tableMutation instanceof TableInsert) {
+                return translateTableMutation(tableMutation);
+            } else {
+                return (T) Mockito.mock(JdbcMutationOperation.class);
```
jyemin, line 176: This is strange.  I've never seen Mockito used outside of test code.  Can you explain why this is here, and what the alternative is if we don't want this dependency?

src/main/java/com/mongodb/hibernate/translate/AstVisitorValueHolder.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+
+package com.mongodb.hibernate.translate;
+
+import static com.mongodb.hibernate.internal.MongoAssertions.assertNotNull;
+import static com.mongodb.hibernate.internal.MongoAssertions.assertNull;
+import static com.mongodb.hibernate.internal.MongoAssertions.assertTrue;
+import static java.lang.String.format;
+
+import com.google.common.reflect.TypeToken;
```
jyemin, line 24: Does Hibernate already have a Guava dependency or would this be an additional dependency?  If the latter, I'm not sure it's wise to take it on.  What would be the alternative if we decide against it (I don't recall that the POC used it in the Attachment class)?

src/main/java/com/mongodb/hibernate/translate/mongoast/AstElement.java
```diff
@@ -0,0 +1,34 @@
+/*
```
jyemin, line 1: This whole package should be under `internal` I think.

jyemin (commented):
Looking good!

A few initial comments.

jyemin:
Why did you close this (after I had already commented :) )?

jyemin:
I copied the comments over to the next PR.

PR #28, opened 2025-01-28
minor improvements

NathanQingyangXu:
another incremental tech improvement PR:

1. make gradle compileJava warning treated as error
2. fix some typo
3. bump dependency versions
4. remove unnecessary `@Nullable` usage on `MongoPreparedStatement`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -103,6 +103,7 @@ tasks.withType<JavaCompile>().configureEach {
 tasks.compileJava {
     // The check defaults to a warning, bump it up to an error for the main sources
     options.errorprone.error("NullAway")
+    options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))
 }
 
 tasks.compileTestJava {
--- a/src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
+++ b/src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
@@ -50,7 +50,6 @@ import org.bson.BsonString;
 import org.bson.BsonType;
 import org.bson.BsonValue;
 import org.bson.types.Decimal128;
-import org.jspecify.annotations.Nullable;
 
 /**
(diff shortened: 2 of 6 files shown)


build.gradle.kts
```diff
@@ -103,6 +103,7 @@ tasks.withType<JavaCompile>().configureEach {
 tasks.compileJava {
     // The check defaults to a warning, bump it up to an error for the main sources
     options.errorprone.error("NullAway")
+    options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))
```
NathanQingyangXu, line 106: The former makes it easy to surpress deprecation warning for false positive; true positive warning will lead to evergreen failure.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -245,35 +227,35 @@ public void addBatch() throws SQLException {
...
         checkParameterIndex(parameterIndex);
         throw new NotYetImplementedException("To implement in scope of https://jira.mongodb.org/browse/HIBERNATE-42");
     }
 
     @Override
-    public void setNull(int parameterIndex, int sqlType, @Nullable String typeName) throws SQLException {
+    public void setNull(int parameterIndex, int sqlType, String typeName) throws SQLException {
         checkClosed();
         checkParameterIndex(parameterIndex);
         throw new NotYetImplementedException("To be implemented during Array / Struct tickets");
```
NathanQingyangXu, line 261: I checked the Javadoc to ensure we should not use `@Nullable`; also, Hibernate would have invoke the `setNull()` method for null value. See https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/type/descriptor/jdbc/BasicBinder.java for further details.

stIncMale (approved):
The last reviewed commit is b0f59920327286faabb401e7d8a76f6f4491b0b4.

PR #31, opened 2025-01-29
Design, implement and document how an application can configure the product

stIncMale:
The corresponding design section is [Configuring](https://docs.google.com/document/d/1yNM-DC7zF70wyCUQxViVLNzkGFGOfYH4jJZru2QbacU/edit?tab=t.0#heading=h.vt9n8w7yf1fd).

We do not want to a programmatic configurator to be specified via [`StandardServiceRegistryBuilder.applySetting`](https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/boot/registry/StandardServiceRegistryBuilder.html#applySetting(java.lang.String,java.lang.Object)), because this approach may work only with [native bootstrapping](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#bootstrap-native), and is not available when [Jakarta bootstrapping](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#bootstrap-jpa) is used. Providing a programmatic configurator as a service works in both cases.

[HIBERNATE-38](https://jira.mongodb.org/browse/HIBERNATE-38)


src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -106,49 +102,32 @@ public <T> T unwrap(Class<T> unwrapType) {
...
-        } else {
-            clientSettingsBuilder = MongoClientSettings.builder().applyConnectionString(connectionString);
+        try {
+            MongoDialectSettings.Builder mongoDialectSettingsBuilder = MongoDialectSettings.builder(configProperties);
+            if (mongoDialectConfigurator != null) {
+                mongoDialectConfigurator.configure(mongoDialectSettingsBuilder);
+            }
+            config = mongoDialectSettingsBuilder.build();
+        } catch (RuntimeException e) {
+            throw new HibernateException(
```
stIncMale, line 121: Do we have to throw `HibernateException` here? If yes, why?

stIncMale: @NathanQingyangXu, sorry, forgot to ping you right away for ☝️

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+ *         </tr>
+ *     </tbody>
+ * </table>
+ *
+ * @see MongoDialectConfigurator
+ */
+public final class MongoDialectSettings {
+    private final MongoClientSettings mongoClientSettings;
+    private final String databaseName;
+    private final boolean autoCommit;
```
stIncMale, line 104: We may trivially support configuring an external `MongoClient` via `MongoDialectSettings`, as opposed to creating an internal one, enalbing applications to reuse clients.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+ *                 </ul>
+ *             </td>
+ *             <td>{@value AvailableSettings#AUTOCOMMIT}, if configured; otherwise {@code false}.</td>
+ *         </tr>
+ *     </tbody>
+ * </table>
+ *
+ * @see MongoDialectConfigurator
+ */
+public final class MongoDialectSettings {
```
stIncMale, line 101: This API is named and designed after `MongoClientSettings`. Given that it allows an application to access `MongoClientSettings.Builder`, this approach seems reasonable to me.

src/integrationTest/java/com/mongodb/hibernate/SessionFactoryTests.java
```diff
@@ -16,7 +16,7 @@
 
 package com.mongodb.hibernate;
 
-import static org.hibernate.cfg.JdbcSettings.JAKARTA_JDBC_URL;
+import static org.hibernate.cfg.AvailableSettings.JAKARTA_JDBC_URL;
```
stIncMale, line 19: Let's always refer to the out-of-the-box configuration properties via `AvailableSettings`, as all of them are available through this interface.

src/integrationTest/resources/hibernate.properties
```diff
@@ -1,3 +1,3 @@
-jakarta.persistence.jdbc.url=mongodb://localhost/mongo-hibernate-test?directConnection=false
 hibernate.dialect=com.mongodb.hibernate.dialect.MongoDialect
 hibernate.connection.provider_class=com.mongodb.hibernate.jdbc.MongoConnectionProvider
+jakarta.persistence.jdbc.url=mongodb://localhost/mongo-hibernate-test?directConnection=false
```
stIncMale, line 3: I rearranged the lines in the order of their importance: the dialect is and the provider are the most important ones, as they plug our product into the Hibernate ORM framework. `jakarta.persistence.jdbc.url` is the next most important configuration property.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -106,49 +102,32 @@ public <T> T unwrap(Class<T> unwrapType) {
...
-        final MongoClientSettings.Builder clientSettingsBuilder;
-        if (mongoClientCustomizer != null) {
-            clientSettingsBuilder = MongoClientSettings.builder();
-            mongoClientCustomizer.customize(clientSettingsBuilder, connectionString);
-        } else {
-            clientSettingsBuilder = MongoClientSettings.builder().applyConnectionString(connectionString);
+        try {
+            MongoDialectSettings.Builder mongoDialectSettingsBuilder = MongoDialectSettings.builder(configProperties);
+            if (mongoDialectConfigurator != null) {
+                mongoDialectConfigurator.configure(mongoDialectSettingsBuilder);
```
stIncMale, line 117: Note that there is no longer a way for an application to decide whether to apply the connection string (if any) before or after doing other configuration changes, because we apply it. If an application absolutely has to apply its connection string after doing some other configuration changes, then it can omit configuring `JAKARTA_JDBC_URL`, and apply whatever connections string it needs to how it wants that via `MongoDialectConfigurator`.

src/test/resources/hibernate.properties
```diff
@@ -1,4 +1,4 @@
-jakarta.persistence.jdbc.url=mongodb://localhost/mongo-hibernate-test?directConnection=false
 hibernate.dialect=com.mongodb.hibernate.dialect.MongoDialect
 hibernate.connection.provider_class=com.mongodb.hibernate.jdbc.MongoConnectionProvider
 hibernate.boot.allow_jdbc_metadata_access=false
+jakarta.persistence.jdbc.url=mongodb://localhost/mongo-hibernate-test?directConnection=false
```
stIncMale, line 4: I rearranged the lines in the order of their importance: the dialect and the provider are the most important ones, as they plug our product into the Hibernate ORM framework. 

`jakarta.persistence.jdbc.url` is almost the next most important configuration property, however, `hibernate.boot.allow_jdbc_metadata_access` is still more important as it is conceptually used by Hibernate ORM before `jakarta.persistence.jdbc.url`, to decide whether the latter is to be used for gathering metadata needed to instantiate the dialect specified in `hibernate.dialect`.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -63,6 +68,10 @@ class MongoConnectionTests {
     @Mock
     private MongoClient mongoClient;
 
+    @Spy
+    private MongoDialectSettings config =
+            MongoDialectSettings.builder(emptyMap()).databaseName("testDbName").build();
```
stIncMale, line 73: I had to use `@Spy` instead of `@Mock` here so that Mockito injected the value of the `config` field specified in the code into `mongoConnection`, as opposed to generating the `config` value on its own.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
```diff
@@ -370,4 +380,37 @@ void testSQLExceptionThrownWhenMetaDataFetchingFailed() {
...
+        MongoDatabase mongoDatabase;
+
+        MongoConnection mongoConnection;
+
+        @BeforeEach
+        void setUp() {
+            String mongoDatabaseName = config.getDatabaseName();
+            doReturn(mongoDatabaseName).when(mongoDatabase).getName();
+            when(mongoClient.getDatabase(eq(mongoDatabaseName))).thenReturn(mongoDatabase);
+            mongoConnection = new MongoConnection(config, mongoClient, clientSession);
```
stIncMale, line 396: I had to construct `mongoConnection` on my own, without Mockito injection via `@InjectMocks MongoConnection mongoConnection`, because mocking `mongoDatabase` in `setUp` happens after that injection. I need mocking to be done before the injection, so that a properly mocked `mongoConnection` is passed to the constructor of `MongoConnection`.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
```diff
@@ -35,72 +35,64 @@
...
 
 class MongoConnectionProviderTests {
 
     @Nested
-    class MongoClientCustomizerTests {
+    class MongoDialectConfiguratorTests {
 
-        @ParameterizedTest
-        @ValueSource(booleans = {true, false})
-        void testMongoClientCustomizerTakeEffect(boolean customizerAppliesConnectionString) {
```
stIncMale, line 60: `customizerAppliesConnectionString` is not longer relevant, as we apply the connection string before using the configurator. See https://github.com/mongodb/mongo-hibernate/pull/31#discussion_r1939903441.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
stIncMale: With the design as is, we don't need to expose neither `MongoDialectSettings` nor `MongoDialectSettings.builder` as part of our API. However, hiding them requires taking an approach different from that of the `MongoClientSettings`. I think, I am fine with the approach I propose in this PR (because I am trying to make `MongoDialectSettings` similar to `MongoClientSettings` here). But if there are concerns about exposing `MongoDialectSettings`, please bring them up.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -41,14 +39,14 @@
...
+    private final MongoDatabase mongoDatabase;
     private final MongoConnection mongoConnection;
     private final ClientSession clientSession;
 
     private boolean closed;
 
-    public MongoStatement(MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection) {
-        this.mongoClient = mongoClient;
+    public MongoStatement(MongoDatabase mongoDatabase, ClientSession clientSession, MongoConnection mongoConnection) {
+        this.mongoDatabase = mongoDatabase;
```
stIncMale, line 49: We could instead keep requiring `mongoClient`, then get `mongoDatabase` as `mongoClient.getDatabase(mongoConnection.getSchema())`, but requiring `mongoDatabase` directly seems better to me.

src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
```diff
@@ -90,6 +91,12 @@ public String getSQLKeywords() {
         return "";
     }
 
+    /** @return {@value #SCHEMA_TERM}. */
+    @Override
+    public String getSchemaTerm() {
+        return SCHEMA_TERM;
+    }
+
```
stIncMale, line 99: I chose to say that a MongoDB database is a JDBC schema (as opposed to saying that it's a JDBC catalog), because semantically [MongoDB database](https://www.mongodb.com/docs/manual/reference/glossary/#:~:text=Center%20Awareness.-,database,in%20the%20file%20system.%20One%20MongoDB%20server%20typically%20has%20multiple%20databases.,-database%20command) is most similar to [PostgreSQL schema](https://www.postgresql.org/docs/current/ddl-schemas.html), which [their JDBC driver uses as the JDBC schema](https://github.com/pgjdbc/pgjdbc/blob/15596302e6ab9ccdefe25c9bd91284d5f232ee27/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L713).

src/main/java/com/mongodb/hibernate/service/MongoClientCustomizer.java
```diff
@@ -1,65 +0,0 @@
...
- * ...
- *
- * var sessionFactory = metadataBuilder().build().getSessionFactoryBuilder().build();
- *
- * // start using sessionFactory as normal
- * ...
- * }</pre>
- */
-@FunctionalInterface
-public interface MongoClientCustomizer extends Service {
```
stIncMale, line 52: I renamed `MongoDialectConfigurator` to `MongoDialectConfigurator` for the following reasons:

- It's not just `MongoClient` that is it capable of configuring, or that we are intending to configure, but rather the whole product ("MongoDB Dialect for the Hibernate ORM").
- Our dialect is a customization of the Hibernate ORM, while this class is merely a configurator of that customization. I think, the work "configure" more accurately describes what this service does than "customize".

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
NathanQingyangXu: That is Hibernate's internal exception base class (other exception like `ServiceException` extends it) so Hibernate has elegant exception handling. Usually `RuntimeException` should be wrapped up somewhere. You can see its Javadoc at https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/HibernateException.java

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+                }
+
+                static RuntimeException failedToParse(String propertyName, String propertyValue, Type type) {
+                    return new RuntimeException(format(
+                            "Failed to get %s from configuration property [%s] with value [%s]",
+                            type.getTypeName(), propertyName, propertyValue));
+                }
+            }
+        }
+    }
```
NathanQingyangXu, line 292: Hibernate has internal configuration parsing helper class: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/jpa/internal/util/ConfigurationHelper.java.

Seems throwing `IllegalArgumentException` is ok. Throwing `HibernateException` is not a hard rule and I think its wrapping will be done by Hibernate internally.

gradle/libs.versions.toml
```diff
@@ -19,6 +20,7 @@ buildconfig = "5.5.1"
 [libraries]
 junit-jupiter = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junit-jupiter" }
 junit-platform-launcher = { module = "org.junit.platform:junit-platform-launcher" }
+assertj = { module = "org.assertj:assertj-core", version.ref = "assertj" }
```
NathanQingyangXu, line 23: thanks for introducing assertJ dependency. Sensible usage of it will enhance testing code readability dramatically!

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
NathanQingyangXu: I am not sure `autoCommit` is a good candidate for it seems an open question now. In JDBC, regardless of whether commit is auto, commit happens all the time. Currently we only commit in face of `autoCommit` value is false.

src/integrationTest/java/com/mongodb/hibernate/SessionFactoryTests.java
NathanQingyangXu: agree. The original reason to use `JdbcSettings` is to emphasize the relevant sub scope.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -208,15 +210,30 @@ public DatabaseMetaData getMetaData() throws SQLException {
...
+     * @return The current schema name.
+     *     <p>Default value: {@link MongoDialectSettings#getDatabaseName()}.
      * @see org.hibernate.tool.schema.extract.internal.AbstractInformationExtractorImpl
      */
     @Override
-    public @Nullable String getSchema() throws SQLException {
+    public String getSchema() throws SQLException {
         checkClosed();
-        return null;
+        return mongoDatabase.getName();
```
NathanQingyangXu, line 236: So the above two methods are only used during Hibernate's DDL step (per our latest design change, we might well delete these two methods for we won't support them at all; DDL is irrelevant to NoSQL; indexing and unique constraint feature later on won't rely on the above methods as well).

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+                this.autoCommit = autoCommit;
+            }
+        }
+
+        /**
+         * Configures {@link MongoDialectSettings}.
+         *
+         * <p>Note that if you {@link MongoClientSettings.Builder#applyConnectionString(ConnectionString)} with
+         * {@linkplain ConnectionString#getDatabase() database name} configured, you still must configure the database
+         * name via {@link Builder#databaseName(String)}, as there is no way for that to happen automatically.
```
NathanQingyangXu, line 180: I like your design and I think it is better than previous implementation.
If I were the common user I would expect the database name in connection string should be the default one if no further db name is configured via current design to override it. but I agree it is not easy to achieve that without keeping the connection string as a field.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+        return new MongoDialectSettings.Builder(notNull("configProperties", configProperties));
+    }
+
+    /** A builder for {@code MongoDialectSettings}. */
+    public static final class Builder {
+        private final MongoClientSettings.Builder mongoClientSettingsBuilder;
+        private @Nullable String databaseName;
+        private boolean autoCommit;
+
+        private Builder(Map<String, Object> configProperties) {
```
NathanQingyangXu, line 162: In Hibernate codebase, usually the naming of the map is `configurationValues` or `configValues` for Hibernate does allow for ad-hoc value to be used as value, not restrictive to java `Properties`.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
NathanQingyangXu: but if we made the following code changes, it seems we can achieve the goal to avoid double db name specification:
```
private String defaultDatabaseName;

public MongoDialectSettings.Builder applyConnectionString(String connectionString) {
      mongoClientSettingsBuilder.applyConnectionString(connectionString);
      defaultDatabaseName = new ConnectionString(connectionString).getDatabase();
}

// ... use defaultDatabaseName if needed

```

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    /**
+     * Creates a new {@link MongoDialectSettings.Builder} based on {@code configProperties}.
+     *
+     * @param configProperties The {@linkplain Configurable#configure(Map) configuration properties}.
+     * @return A new {@link MongoDialectSettings.Builder}.
+     */
+    public static MongoDialectSettings.Builder builder(Map<String, Object> configProperties) {
+        return new MongoDialectSettings.Builder(notNull("configProperties", configProperties));
```
NathanQingyangXu, line 153: I think given `configProperties` is not marked as `@Nullable`, we could save the usage of `notNull()` and trust `NullAway` to cover our asses (in theory at least any usage inconsistent with the `jspecify` annotation would be treated as bug during compiling phase). This is one of improvements in this project compared to Java driver project.

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class MongoDialectSettingsTests {
+    @Test
+    void requiresPropertiesThatHaveNoDefaults() {
+        NullPointerException e = assertThrows(NullPointerException.class, () -> MongoDialectSettings.builder(emptyMap())
+                .build());
+        assertThat(e.getMessage()).matches("databaseName must not be null");
```
NathanQingyangXu, line 40: `matches` means RegExp matching. `equals` might be better though for now there is no difference between them.

but imagine if in the future some innocent guy inserted `.` to the end of the message than there might be some subtleties unexpected.

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Test
+    void defaults() {
+        MongoDialectSettings config = MongoDialectSettings.builder(emptyMap())
+                .databaseName("testDbName")
+                .build();
+        assertEquals(MongoClientSettings.builder().build(), config.getMongoClientSettings());
+        assertFalse(config.isAutoCommit());
```
NathanQingyangXu, line 49: we might well use `assertThat()` consistently for all new testing assertion scenarios. It is a change for good if you get used to it; also it has potential of fluent API usage.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
NathanQingyangXu: Here is an ad-hoc config usage in Hibernate ORM codebase: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/boot/cfgxml/internal/CfgXmlAccessServiceImpl.java

```
public class CfgXmlAccessServiceImpl implements CfgXmlAccessService {
	private final LoadedConfig aggregatedCfgXml;

	public CfgXmlAccessServiceImpl(Map<?,?> configurationValues) {
		aggregatedCfgXml = (LoadedConfig) configurationValues.get( LOADED_CONFIG_KEY );
	}

	@Override
	public LoadedConfig getAggregatedConfig() {
		return aggregatedCfgXml;
	}
}
```

`configProperties` is fine but it might mislead reader into thinking that it comes from `hibernate.properties` file exclusively. Actually it could be populated in ad-hoc way during Hibernate bootstrapping.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -16,17 +16,21 @@
...
 
+import com.mongodb.hibernate.jdbc.MongoConnectionProvider;
 import org.hibernate.dialect.DatabaseVersion;
 import org.hibernate.dialect.Dialect;
 import org.hibernate.engine.jdbc.dialect.spi.DialectResolutionInfo;
 
 /**
- * A MongoDB {@link Dialect} for {@linkplain #getMinimumSupportedVersion() version 6.0 and above}.
+ * A MongoDB {@link Dialect} for {@linkplain #getMinimumSupportedVersion() version 6.0 and above}. Must be used together
+ * with {@link MongoConnectionProvider}.
```
NathanQingyangXu, line 26: not sure why we need to emphasize the combination. To me the two classes are not tightly coupled. Traditionally `Dialect` is used to customize programmatically (as opposed to `hibernate.properties` config file). `ConnectionProvider` customization is in the scope of properties file without corresponding Dialect method to be overridden.

But it is ok to retain it. Now I think your point is the two mandatory entries in hibernate.properties:
```
hibernate.dialect=com.mongodb.hibernate.dialect.MongoDialect
hibernate.connection.provider_class=com.mongodb.hibernate.jdbc.MongoConnectionProvider
```

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+        public MongoDialectSettings.Builder applyToMongoClientSettings(
+                final Consumer<MongoClientSettings.Builder> configurator) {
+            notNull("configurator", configurator).accept(mongoClientSettingsBuilder);
+            return this;
+        }
+
+        /**
+         * Sets the name of a MongoDB database used as the {@linkplain Connection#getSchema() JDBC schema} of a
+         * {@linkplain Connection connection} {@linkplain MongoConnectionProvider#getConnection() obtained} from
+         * {@link MongoConnectionProvider}.
```
NathanQingyangXu, line 195: I understand the introduction of JDBC schema is to pave the way for future ticket, but this linkage is confusing for me and seems not compelling. Our end user might simply wants a database name without understanding why JDBC schema is involved.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+         *
+         * @param databaseName The name of the default MongoDB database.
+         * @return {@code this}.
+         * @see MongoDialectSettings#getDatabaseName()
+         * @see DatabaseMetaData#getSchemaTerm()
+         */
+        public MongoDialectSettings.Builder databaseName(String databaseName) {
+            this.databaseName = notNull("databaseName", databaseName);
```
NathanQingyangXu, line 203: seems unnecessary if we trust `NullAway`'s protection.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
```diff
@@ -0,0 +1,293 @@
...
+
+        /**
+         * Sets the name of a MongoDB database used as the {@linkplain Connection#getSchema() JDBC schema} of a
+         * {@linkplain Connection connection} {@linkplain MongoConnectionProvider#getConnection() obtained} from
+         * {@link MongoConnectionProvider}.
+         *
+         * @param databaseName The name of the default MongoDB database.
+         * @return {@code this}.
+         * @see MongoDialectSettings#getDatabaseName()
+         * @see DatabaseMetaData#getSchemaTerm()
```
NathanQingyangXu, line 200: Do we need to emphasize the confusing `DatabaseMetadata#getSchemaTerm()`? Even in Hibernate ORM codebase this method is never used at all.

src/main/java/com/mongodb/hibernate/internal/MongoChecks.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return {@code value}.
+     * @throws NullPointerException If {@code value} is {@code null}.
+     */
+    public static <T> T notNull(final String name, @Nullable T value) {
+        assertNotNull(name);
+        if (value == null) {
+            throw new NullPointerException(name + " must not be null");
+        }
+        return value;
+    }
```
NathanQingyangXu, line 46: So we have already a `MongoAssertion` internal class to assert at runtime whether some value is null; I think given the usage of `NullAway`, this new class is not needed for method argument nullness checking, so it seems to me its function is duplicated with `MongoAssertion` class? The exception types are different, so that is the reason we introduce this class?

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
NathanQingyangXu: This approach works. More intuitive and idematic way might be:

```
    @Mock(lenient = true)
    private MongoDialectSettings config;

    @InjectMocks
    private MongoConnection mongoConnection;

    @BeforeEach
    void setUp() {
        doReturn("testDbName").when(config).getDatabaseName();
    }
```
the benefit is subjective. It ends up with avoiding an unnecessary hack so it is easy to understand and change in the long future.

src/main/java/com/mongodb/hibernate/service/MongoDialectConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hibernate.service.Service;
+
+/**
+ * A Hibernate {@link Service} an application may use for programmatically configuring {@link MongoDialectSettings}.
+ *
+ * <p>An example usage is as follows:
+ *
+ * <pre>{@code
+ * var dialectConfigurator = (dialectSettingsBuilder) -> {
+ *     // configure client settings to your heart's content
```
NathanQingyangXu, line 30: the above comment might be outdated for `databaseName` and `autoCommit` is also in the scope

src/main/java/com/mongodb/hibernate/service/MongoDialectConfigurator.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * // start using sessionFactory as normal
+ * ...
+ * }</pre>
+ */
+@FunctionalInterface
+public interface MongoDialectConfigurator extends Service {
+
+    /**
+     * Configures {@code dialectSettingsBuilder}.
```
NathanQingyangXu, line 51: seems `Configures {@link DialectSettings} through its builder` might be accurate.

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
```diff
@@ -0,0 +1,138 @@
...
+import com.mongodb.MongoClientSettings;
+import java.util.Map;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class MongoDialectSettingsTests {
+    @Test
+    void requiresPropertiesThatHaveNoDefaults() {
+        NullPointerException e = assertThrows(NullPointerException.class, () -> MongoDialectSettings.builder(emptyMap())
+                .build());
```
NathanQingyangXu, line 39: seems it is counter-intuitive not to support
```
MongoDialectSettings.builder()
```
in lieu of
```
MongoDialectSettings.builder(emptyMap())
```
an overloaded method seems good to have

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
NathanQingyangXu: nvm. saw the `.matches()` usage later in the same class. It makes sense to use it as it is then.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
NathanQingyangXu: This works but seems not the most elegant solution. Given the logic change (we build `MongoDatabase` within constructor as a by-product), `@InjectMocks` seems inappropriate. 
The following code will work and make it easier to maintain in the future:
```
class MongoConnectionTests {

    @Mock
    private ClientSession clientSession;

    @Mock
    private MongoClient mongoClient;

    @Mock
    private MongoDialectSettings config;

    @Mock
    private MongoDatabase mongoDatabase;

    private MongoConnection mongoConnection;

    @BeforeEach
    void setUp() {
        doReturn("testDbName").when(config).getDatabaseName();
        doReturn(mongoDatabase).when(mongoClient).getDatabase("testDbName");
        mongoConnection = new MongoConnection(config, mongoClient, clientSession);
    }
    
    ... ...
    @Nested
    class SetGetSchemaTests {

        @BeforeEach
        void setUp() {
            doReturn("testDbName").when(mongoDatabase).getName();
        }

        @Test
        void getSchemaDefault() throws SQLException {
            assertEquals(config.getDatabaseName(), mongoConnection.getSchema());
        }
        ... ...
   }
```

NathanQingyangXu:
overall it looks great and the `MongoDialectSettings` related stuff is a big improvement from my perspective. Left some comments. Recommended removing the JDBC schema related stuff for they are confusing and not strictly necessary for now. Will approve after that. Great work!

NathanQingyangXu (changes requested):
will approve once JDBC schema stuff is removed

src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
NathanQingyangXu: it might be right. But as a general rule, we only implemented JDBC methods Hibernate really called and we decided to support, right? It seems confusing why we add this method only we could call it. The discussion about the mapping seems irrelevant in this PR. We could provide a database without going through JDBC schema, right?

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
NathanQingyangXu: I doubt whether we need to include autoCommit() for it is internal Hibernate implememntatoin detail and its business logic might be complex that "we set a Connection's autoCommit status after it was created". That part might interfere with Hibernate's internal working. For instance, without our doing anything, the Connection would be set autoCommit as expected.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -48,38 +48,34 @@
...
 
     @Override
     public Connection getConnection() throws SQLException {
         try {
             var client = assertNotNull(mongoClient);
             var clientSession = client.startSession();
-            return new MongoConnection(client, clientSession);
+            var initializedConfig = assertNotNull(config);
+            var connection = new MongoConnection(initializedConfig, client, clientSession);
+            connection.setAutoCommit(initializedConfig.isAutoCommit());
```
NathanQingyangXu, line 77: the above step is confusing for Hibernate internally will set autoCommit based on the common configuration already. Should we duplicate what Hibernate has done (which might be more complicated than the above)?

You can test by removing the above statement and you can see Hibernate is working just as expected.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
stIncMale: Agreed to create `MongoConnection` manually and get rid of `@InjectMocks`. That way we don't need to `@Mock`/`@Spy` on `MongoDialectSettings`.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
stIncMale: See https://github.com/mongodb/mongo-hibernate/pull/31#discussion_r1949677069.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
stIncMale: Discussed. This is fine.

src/main/java/com/mongodb/hibernate/service/MongoDialectConfigurator.java
stIncMale: Discussed.

```java
// configure the dialect using `dialectSettingsBuilder`
```java

File: src/integrationTest/java/com/mongodb/hibernate/BasicInsertionTests.java [hunk 31]
```diff
@@ -132,7 +133,7 @@ private List<BsonDocument> getCollectionDocuments() {
     }
 
     private void assertCollectionContainsOnly(BsonDocument expectedDoc) {
-        assertThat(getCollectionDocuments()).asList().singleElement().isEqualTo(expectedDoc);
+        assertThat(getCollectionDocuments()).asInstanceOf(LIST).singleElement().isEqualTo(expectedDoc);
```
stIncMale, line 136: `asList` is deprecated.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
```diff
@@ -177,18 +187,54 @@ private void verifyMongoClient(
...
+            standardServiceRegistryBuilder.applySetting(JAKARTA_JDBC_URL, connectionString);
         }
         return new MetadataSources(standardServiceRegistryBuilder.build())
                 .getMetadataBuilder()
                 .build()
                 .getSessionFactoryBuilder()
                 .build();
     }
+
+    private static final class NoopServiceRegistryImplementor implements ServiceRegistryImplementor {
```
stIncMale, line 206: Mockito fails to mock `ServiceRegistryImplementor` despite the latter being an interface (it mocks other interfaces just fine). I failed to quickly figure out what to do with that, and decided that implementing a noop implementation manually is an OK compromiss.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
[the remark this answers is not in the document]
stIncMale: Done in fdc4c5feb1dff31b5751105ae5c9be4e83fc9ab0.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
stIncMale: It is up to a connection provider to respect `AvailableSettings.AUTOCOMMIT` or not. We discussed and agreed not to respect it.

Done in 0155ba09610e16853c3e6c9fbf41302186c39b8d.

src/main/java/com/mongodb/hibernate/service/MongoDialectConfigurator.java
stIncMale: Done in fdc4c5feb1dff31b5751105ae5c9be4e83fc9ab0.

src/main/java/com/mongodb/hibernate/service/MongoDialectConfigurator.java
stIncMale: Done in fdc4c5feb1dff31b5751105ae5c9be4e83fc9ab0.

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
stIncMale: Discussed, agreed to leave this up to the code writer. I am leaving `assertFalse` here.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
stIncMale: Discussed. I think that `configurationValues` is wrong - that map contains configuration property names as keys, and configuration property values as values. Hibernate ORM API docs describe this map as [_"The configuration properties"_](https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/service/spi/Configurable.html#configure(java.util.Map)), which makes much more sense than `configurationValues`, which Hibernate ORM for some reason uses as the parameter name.

I renamed it nonetheless in fdc4c5feb1dff31b5751105ae5c9be4e83fc9ab0, because you want this.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
stIncMale: We agreed not to interpret the MongoDB database as JDBC schema, at least for now.

Done in 3bb8ee965f06d51bdc5b8fd333fa346363017c4b.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
stIncMale: I don't think we should copy that class from Hibernate ORM to our code.

src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
stIncMale: We agreed not to interpret the MongoDB database as JDBC schema, at least for now.

Done in 3bb8ee965f06d51bdc5b8fd333fa346363017c4b.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
stIncMale: I removed `SetGetSchemaTests` in 3bb8ee965f06d51bdc5b8fd333fa346363017c4b because we agreed not to interpret the MongoDB database as JDBC schema, at least for now, and as a result, `MongoConnection` no longer overrides `setSchema`.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionTests.java
stIncMale: Done in fdc4c5feb1dff31b5751105ae5c9be4e83fc9ab0.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
stIncMale: I agree that auto-commit shouldn't be configurable programmatically via our API.

Done in 0155ba09610e16853c3e6c9fbf41302186c39b8d.

src/main/java/com/mongodb/hibernate/dialect/MongoDialectSettings.java
NathanQingyangXu: I think it is good to follow some convention; sometimes it is not technically accurate for historical reason but we have multiple perspectives to consider.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
NathanQingyangXu: This is interesting. The first time I ever encountered such Mockito issue.

A better alternative to assert that `autoCommit` is true by default is to verify in integration testing, which is more compelling and intuitive. E.g. we could add the following method to `SessionTests` in integrationTest folder:

```
    @Test
    void defaultAutoCommitTrue() {
        assertDoesNotThrow(() -> session.doWork(connection -> {
            assertTrue(connection.getAutoCommit());
        }));
    }
```

stIncMale: Thank you for helping me with a solution.

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
stIncMale, line 1: I wonder if we should move this interface to `com.mongodb.hibernate.cfg`.

README.md
```diff
@@ -1,23 +1,26 @@
-# A MongoDB Dialect for the Hibernate ORM
+# MongoDB extension of Hibernate ORM
```
stIncMale, line 1: Renamed the product in as we discussed.

README.md
```diff
@@ -1,23 +1,26 @@
...
 
-This project aims to provide a library to seamlessly integrate MongoDB with Hibernate ORM. Hibernate _ORM_ is a powerful **O**bject-**r**elational **m**apping tool. Due to the SQL and JDBC standards, Hibernate ORM could centralize each SQL vendor's idiosyncrasies in the so-called _Hibernate Dialect_. This project will include a document database member in the Hibernate's Dialect family.
+This product enables applications to use databases managed by the [MongoDB](https://www.mongodb.com/) DBMS
+via [Hibernate ORM](https://hibernate.org/orm/).
 
 ## Overview
 
-MongoDB speaks _MQL_ (**M**ongoDB **Q**uery **L**anguage in JSON format) instead of SQL. This project creates a MongoDB Hibernate Dialect by:
+MongoDB speaks MQL (**M**ongoDB **Q**uery **L**anguage) instead of SQL. This product works by:
```
stIncMale, line 8: The MongoDB DBMS does not understand JSON, and does not accept MQL encoded as JSON.

src/integrationTest/java/com/mongodb/hibernate/BasicInsertionTests.java
```diff
@@ -115,12 +116,8 @@ void testEntityWithEmbeddedFieldInsertion(SessionFactoryScope scope) {
     }
 
     private void onMongoCollection(Consumer<MongoCollection<BsonDocument>> collectionConsumer) {
-        var connectionString = new ConnectionString(new Configuration().getProperty(JAKARTA_JDBC_URL));
-        try (var mongoClient = MongoClients.create(MongoClientSettings.builder()
-                .applyConnectionString(connectionString)
-                .build())) {
-            var collection =
-                    mongoClient.getDatabase(connectionString.getDatabase()).getCollection("books", BsonDocument.class);
+        try (var mongoClient = MongoClients.create(config.mongoClientSettings())) {
```
stIncMale, line 119: We now use `MongoConfiguration` instead of reading and parsing `JAKARTA_JDBC_URL` directly.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -130,6 +132,9 @@ abstract class AbstractMqlTranslator<T extends JdbcOperation> implements SqlAstT
 
     AbstractMqlTranslator(SessionFactoryImplementor sessionFactory) {
         this.sessionFactory = sessionFactory;
+        assertNotNull(
+                assertNotNull(sessionFactory.getServiceRegistry().getService(StandardServiceRegistryScopedState.class))
+                        .getConfiguration());
```
stIncMale, line 137: This way we can be sure that we will be able to access `MongoConfiguration` from our translators if we need that.

src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Map;
+import org.hibernate.HibernateException;
+import org.hibernate.boot.registry.StandardServiceInitiator;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.service.Service;
+import org.hibernate.service.UnknownServiceException;
+import org.hibernate.service.spi.ServiceRegistryImplementor;
+import org.jspecify.annotations.Nullable;
+
+public final class StandardServiceRegistryScopedState implements Service {
```
stIncMale, line 37: We will be able to use the single instance of this service as shared state between the MQL translation code (plugged via `MongoDialect`) the JDBC adapter code (plugged via `MongoConnectionProvider`). This is the solution to the "we can't use static state, if we want to support multiple session factories per application" issue.

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
```diff
@@ -0,0 +1,58 @@
...
+ * ...
+ *
+ * var sessionFactory = metadataBuilder().build().getSessionFactoryBuilder().build();
+ *
+ * // start using sessionFactory as normal
+ * ...
+ * }</pre>
+ */
+@FunctionalInterface
+public interface MongoConfigurationContributor extends Service {
```
stIncMale, line 49: Hibernate ORM seem to name `*Contributor` things that add plugins or configuration: `MetadataContributor`, `ServiceContributor`, I followed that naming.

src/test/java/com/mongodb/hibernate/dialect/MongoDialectSettingsTests.java
stIncMale: This method became the constructor of `MongoConfigurationBuilder`, and I added a constructor with no parameters for testing purposes.

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
```diff
@@ -16,177 +16,74 @@
...
+import org.mockito.MockMakers;
+import org.mockito.junit.jupiter.MockitoExtension;
 
+@ExtendWith(MockitoExtension.class)
 class MongoConnectionProviderTests {
+    @Mock(mockMaker = MockMakers.PROXY)
+    private ServiceRegistryImplementor serviceRegistry;
 
-    @Nested
-    class MongoClientCustomizerTests {
```
stIncMale, line 58: This was refactored into a top-level `MongoConfigurationContributorTests`.

stIncMale:
@NathanQingyangXu I made so significant changes to this PR, that it does not make sense to attempt and review only the new commits. I recommend re-reviewing the whole thing. Also, maybe a code walkthrough will be useful.

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
NathanQingyangXu: if it is a service, moving to `service`package makes more sense to me

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: yeah, this is a placeholder. When we do need it, we could update it accordingly. Currently it is fine to be as it is now.
`requireService()` usage might be a little better, but given this is a placeholder, it doesn't matter.

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
NathanQingyangXu: good point!!

NathanQingyangXu (approved):
Great job!! I could not find any defect at all, :).

src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
stIncMale: > A better alternative to assert that autoCommit is true by default is to verify in integration testing

A test via `Session` does not test what I am testing here. Only be testing the provider directly can we verify that the connections it produces have auto-commit enabled.

NathanQingyangXu:
> @NathanQingyangXu I made so significant changes to this PR, that it does not make sense to attempt and review only the new commits. I recommend re-reviewing the whole thing. Also, maybe a code walkthrough will be useful.

Definitively I will review the whole thing with fresh eyes. It goes beyond my wildest expectation, :).

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
stIncMale: The approach to internal packages is irrelevant here, because the question is about `com.mongodb.hibernate.service.MongoConfigurationContributor`, which is part of our API and is not internal.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
stIncMale: Done in a8198a18c9479e86bc17825c9e1621daf6459f97.

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
stIncMale: @NathanQingyangXu, you raised a concern that all `*Contributor`s in Hibernate ORM API are annotated with `@JavaServiceLoadable`, and suggested that because of that, we should name our interface `MongoConfigurationContributor`, as users will expect it to be loadable via `ServiceLoader`.

However, a quick search found the following counterexamples:
- https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/boot/model/source/spi/EntityNamingSourceContributor.html
- https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/sql/results/graph/DatabaseSnapshotContributor.html

They tell us that Hibernate ORM does not imply loadability via `ServiceLoader` by the `*Contributor` suffix.

stIncMale: Should we rename `MongoConfigurationContributor.contribute` back to `configure`, or leave it as is?

stIncMale: Discussed, the method will be renamed back.

NathanQingyangXu (approved):

NathanQingyangXu (approved):

src/main/java/com/mongodb/hibernate/service/MongoConfigurationContributor.java
stIncMale: Discussed, moved to a new `spi` subpackage.

PR #35, opened 2025-02-04
Hibernate batch processing

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-35

Hibernate uses JDBC batch feature only for `PreparedStatement`, so the code logic changes are centralized in `MongoPreparedStatement`, not in `MongoStatement`. On top of JDBC's batch feature, Hibernate builds some higher abstraction (see `BatchImpl` class: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/engine/jdbc/batch/internal/BatchImpl.java), so we need to code our logic in the context of Hibernate's usage.

Internally bulk writing API methods are used as implementation. `ReplaceOne` write model is not added for it is out of scope of milestone1.

It is a pity there is no possibility to showcase the batch processing using high-level Hibernate's CRUD processing context.
I tested it on top of our first CRUD PR and confirmed that it works as expected (by checking logs and explore in debugging mode).

The pattern of batch processing in `PreparedStatement` is as follows:

```
for (var i = 0;  i < batch_size; i++) {
    pstmt.setXXX(1, ...);
    pstmt.setXXX(2, ...);
    ... ...
    pstmt.addBatch();
    pstmt.executeBatch();
}

```

Only when `PreparedStatement` is to be released (but not closed), should the `clearBatch()` method be invoked. As the pattern above shows, the internal statement batch should have been emptied after `executeBatch()` invocation (Hibernate would guarantee there is no dangling last batch without executing), so in theory there is not much to be done in the `clearBatch()` method (its name might be misleading into thinking that it is in charge of emptying the batch storage internally).


src/main/java/com/mongodb/hibernate/internal/ConfigurationHelper.java
```diff
@@ -0,0 +1,47 @@
...
+        if (value instanceof Integer) {
+            return (Integer) value;
+        }
+        if (value instanceof String) {
+            return Integer.parseInt((String) value);
+        }
+        throw new ConfigurationException("Could not determine how to handle configuration value [name=" + name
+                + ", value=" + value + "(" + value.getClass().getName() + ")] as int");
+    }
+}
```
NathanQingyangXu, line 47: the above util class was copied from Hibernate's internal class.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -312,4 +371,71 @@ private void checkParameterIndex(int parameterIndex) throws SQLException {
...
         }
     }
+
+    private int[] doExecuteBatch() throws SQLException {
+        startTransactionIfNeeded();
+        try {
+            var writeModels = new ArrayList<WriteModel<BsonDocument>>(commandBatch.size());
+            String collectionName = null;
+            for (BsonDocument command : commandBatch) {
+                var commandName = command.getFirstKey();
```
jyemin, line 381: Just so I understand correctly.  In Hibernate usage scenarios, is it the case that the command name will be the same for all commands in the batch?

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -312,4 +371,71 @@ private void checkParameterIndex(int parameterIndex) throws SQLException {
...
+                    assertTrue(collectionName.equals(curCollectionName));
+                }
+
+                List<WriteModel<BsonDocument>> subWriteModels;
+
+                switch (assertNotNull(commandName)) {
+                    case "insert":
+                        var documents = command.getArray("documents");
+                        subWriteModels = new ArrayList<>(documents.size());
+                        for (var document : documents) {
```
jyemin, line 395: Just so I understand correctly.  In Hibernate usage scenarios, is it the case that there will only ever be one document in this array?  And same question for `updates` and `deletes` arrays below.

jyemin (commented):
Initial comments

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -50,23 +61,46 @@
...
+import org.jspecify.annotations.Nullable;
 
 /**
  * MongoDB Dialect's JDBC {@link java.sql.PreparedStatement} implementation class.
  *
  * <p>It only focuses on API methods MongoDB Dialect will support. All the other methods are implemented by throwing
  * exceptions in its parent {@link PreparedStatementAdapter adapter interface}.
  */
-final class MongoPreparedStatement extends MongoStatement implements PreparedStatementAdapter {
+class MongoPreparedStatement extends MongoStatement implements PreparedStatementAdapter {
```
stIncMale, line 72: What is the reason behind making this class extensible in this PR?

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -260,6 +288,37 @@ public void setNull(int parameterIndex, int sqlType, String typeName) throws SQL
         throw new NotYetImplementedException("To be implemented during Array / Struct tickets");
     }
 
+    @Override
+    public void addBatch() throws SQLException {
+        checkClosed();
+        if (!batchable) {
```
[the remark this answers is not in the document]
NathanQingyangXu: Yeah, we can trust Hibernate only calls this method when batch processing is enabled. I added this for robustness. I removed this redundant checking.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: From current understanding, we can only guarantee that the collection or table name is the same (Hibernate groups statements internally), but it seems we can't assume the command names are the same.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: It is perfectly possible for HQL's insertion to have multiple documents (similar to SQL). Below is the relevant ANTLR BNF rules for HQL's insert statement:

```
/**
 * An 'insert' statement
 */
insertStatement
	: INSERT INTO? targetEntity targetFields (queryExpression | valuesList) conflictClause?
	;
... ...
/**
 * A 'values' clause in an 'insert' statement, with one of more tuples of values to insert
 */
valuesList
	: VALUES values (COMMA values)*
	;
/**
 * A tuple of values to insert in an 'insert' statement
 */
values
	: LEFT_PAREN expressionOrPredicate (COMMA expressionOrPredicate)* RIGHT_PAREN
	;
```

so some HQL insert statements like `insert Book (title, year) values values ("War and Peace", 1869), ("Crime and Punishment", 1866) is perfectly fine, just as the SQL counterpart.

but only insert HQL statement has such possibility.

We also need to take into consideration of native query (which will go through the flow as well; only difference is the mql is not generated by Hibernate). If we support native query case, I think it is perfectly possible to have multiple documents.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -50,23 +61,46 @@
...
+    private BsonDocument command;
 
     private final List<Consumer<BsonValue>> parameterValueSetters;
 
     public MongoPreparedStatement(
             MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection, String mql) {
+        this(mongoClient, clientSession, mongoConnection, mql, false);
+    }
+
+    public MongoPreparedStatement(
```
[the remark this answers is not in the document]
NathanQingyangXu: good catch! changed.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
jyemin: I don't see how that grammar would result in multiple inserts (except in batch processing scenario, where there is a single insert but multiple sets of parameters).

NathanQingyangXu: You were asking about the above `documents`, right? Defintively one inserts, but our insertion MQL translation might need to end up with multiple `documents` in the `documents` field. Currently our mongoast model dicates single array element.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: Yeah, by default Hibernate will group by the combination of the following two entries:

- entity name (without hierachy it maps to table or collection name)
- mutation type

So I could add another validation rule here as I've done for collection name; though seems not mandatory as the collection uniqueness validation.

jyemin: No need.  Just checking my mental model.

NathanQingyangXu: simplified

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -59,15 +71,35 @@
...
+    MongoPreparedStatement(
+            MongoClient mongoClient,
+            ClientSession clientSession,
+            MongoConnection mongoConnection,
+            String mql,
+            boolean batchable) {
         super(mongoClient, clientSession, mongoConnection);
-        this.command = BsonDocument.parse(mql);
-        this.parameterValueSetters = new ArrayList<>();
+        if (batchable) {
```
jyemin, line 93: The `batchable` parameter and this conditional logic should be removed.  Just because batching is global enabled doesn't mean it's actually going to be used in this particular statement, which I think you can only determine once `#addBatch` is called.  So whatever initialization is done in the constructor has to work for both cases.

If this is being done as a performance optimization, it is a premature one.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: This might be one of the best code review comments I ever received. I removed its usage and found this PR was simplified so much!
Yeah, it is premature perf opt (to save some command prototype creation, etc.).

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -106,8 +110,11 @@ public <T> T unwrap(Class<T> unwrapType) {
     }
 
     @Override
-    public void configure(Map<String, Object> configValues) {
-        var jdbcUrl = configValues.get(JAKARTA_JDBC_URL);
+    public void configure(Map<String, Object> configurationValues) {
+
+        batchSize = ConfigurationHelper.getInt(AvailableSettings.STATEMENT_BATCH_SIZE, configurationValues, 0);
```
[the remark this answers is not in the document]
NathanQingyangXu: the latest commit has removed the usage of `batchable` and `addBatch()` method alone will decide. For that reason, no configuration is needed at all in this PR. I removed the copied `ConfigurationHelper` as well.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -59,15 +69,20 @@
  */
 final class MongoPreparedStatement extends MongoStatement implements PreparedStatementAdapter {
 
-    private final BsonDocument command;
+    private final List<BsonDocument> commandBatch;
+    private final BsonDocument commandPrototype;
```
jyemin, line 73: I suspect

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -260,6 +269,27 @@ public void setNull(int parameterIndex, int sqlType, String typeName) throws SQL
         throw new NotYetImplementedException("To be implemented during Array / Struct tickets");
     }
 
+    @Override
+    public void addBatch() throws SQLException {
+        checkClosed();
+        commandBatch.add(command);
+        command = commandPrototype.clone();
+        parameterValueSetters.clear();
+        parseParameters(command, parameterValueSetters);
```
jyemin, line 278: It's surprising that this parameter parsing work has to be repeated for each batch.  Can you think of a way to avoid that?

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -260,6 +269,27 @@ public void setNull(int parameterIndex, int sqlType, String typeName) throws SQL
...
+    }
+
+    @Override
+    public void clearBatch() throws SQLException {
```
jyemin, line 282: The Javadoc for this method says:

> Empties this Statement object's current list of SQL commands.

Any reason not to do that here (and assert it in tests)?

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -312,4 +342,80 @@ private void checkParameterIndex(int parameterIndex) throws SQLException {
                     parameterIndex, parameterValueSetters.size()));
         }
     }
+
+    private int[] doExecuteBatch() throws SQLException {
```
jyemin, line 346: While I appreciate small methods, this one isn't pulling enough weight and should be inlined into its caller.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: Firstly the Javadoc is for `Statement`, `PreparedStatement`'s parent class. But given only `PreparedStatement` is implementing it, the normal procedure is to empty the internal parameters list (remember `PreparedStatement` will reuse the same parameterized SQL, so only parameters list or matrix matters). But another important tech detail is once `executeBatch()` is done, the parameter list will be emptied (`Statement` will invoke `executeBatch(String)` so the Javadoc makes sense).

So technically if we trust Hibernate is working as expected, we could only expect that when `clearBatch()` is invoked, there is nothing to be done (maybe setting the internal list to null but it must have been empty).

The subtle hidden doc is discussed at https://stackoverflow.com/questions/9192541/are-the-parameter-values-automatically-cleared-after-executing-a-batch-of-sql-co as well.

I also tried to explain the subtlety at the beginning of the PR description recently. The Javadoc is only meant for `Statement` when different SQL statements need to be stored and after all the invocation of `executeBatch(sql)` is done could all of them are ready to be deleted.

However, I changed the logic to empty the `commandBatch` list and invoke it in the final clause of `executeBatch()`, just as some guy in the above discussion suggests.

I also added a testing case to avoid the confusion.

NathanQingyangXu: I also checked some JDBC driver's implementation logic. For H2, `clearBatch()` simply set the internal list to null without emptying it; in PG, it does empty it maybe due to robustness reason. Regardless, each driver will empty the parameters list (not SQL statements list) at the end of `executeBatch()` method internally.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: It might be counter-intuitive but to me it seems inevitable.

So compared to POC project in which the parameter parsing and populating happens at the last instance when execution is required. This is one implementation plan but we changed it to preemptive alternative to parse beforehand and populating parameters incremetally during the `setXXX()` method invocation. What are the benefits compared to the approach in POC?

The benefits are two fold: 1) timely validation, so if a parameter index is out of range, setXXX() method would throw exception without giving us a late surprise; 2) timely value setting validation for the parameter populating is done preemptively.

But given MongoDB has no parameterizing mechanism, parsing seems inevitable. In normal JDBC driver case, the bulk of work happens on the DB server side so driver simply passes parameter values to the server without much ado; but in our case, we have to combine both client side and server side job into one.

If non-batch scenario requires parsing (regardless of whether it happens preemptively or lazily), I don't see there is easy way to avoid repeating it one by one. Maybe there is some magic like creating a prototype parsing once and apply it to any command cloned from the prototype. 

I will keep thinking about it and it seems a big challenge.

NathanQingyangXu: It turns out not to be difficult. I changed the implementation by reusing the same setter to reset parameters; whenever a command needs to be added to the batch, we first create its clone.

NathanQingyangXu: I studied the code of `BsonDocument#clone()` as below:
```
public BsonDocument clone() {
        BsonDocument to = new BsonDocument(this.size());

        for(Map.Entry<String, BsonValue> cur : this.entrySet()) {
            switch (((BsonValue)cur.getValue()).getBsonType()) {
                case DOCUMENT:
                    to.put((String)((String)cur.getKey()), (BsonValue)((BsonValue)cur.getValue()).asDocument().clone());
                    break;
                case ARRAY:
                    to.put((String)((String)cur.getKey()), (BsonValue)((BsonValue)cur.getValue()).asArray().clone());
                    break;
                case BINARY:
                    to.put((String)((String)cur.getKey()), (BsonValue)BsonBinary.clone(((BsonValue)cur.getValue()).asBinary()));
                    break;
                case JAVASCRIPT_WITH_SCOPE:
                    to.put((String)((String)cur.getKey()), (BsonValue)BsonJavaScriptWithScope.clone(((BsonValue)cur.getValue()).asJavaScriptWithScope()));
                    break;
                default:
                    to.put((String)cur.getKey(), (BsonValue)cur.getValue());
            }
        }

        return to;
    }
```

It seems BsonString's object reference will be copied, right? Only BINARY type seems to be based on content?

jyemin: Yeah, I was confusing it in my head with serialization.  Since binary is going to be rare, this is probably ok then.

NathanQingyangXu: In traditional Hibernate, we use `Blob`, `Clob`, `NBlob`, `NClob` to represent binary JDBC types. I remembered in our scope doc we excluded them (OGM mapped them to `gridfs`), so it seems perfectly ok for now.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -47,12 +47,16 @@ class MongoStatement implements StatementAdapter {
 
     private boolean closed;
 
-    public MongoStatement(MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection) {
+    MongoStatement(MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection) {
```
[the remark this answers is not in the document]
jyemin: Sorry, I see we have HIBERNATE-40 to address this, so resolving.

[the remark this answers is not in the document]
NathanQingyangXu: sorry, I saw your comment too late. I've implemented your request in a new commit. Let me know whether it is fine. it is not hard to refactor.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -260,6 +269,99 @@ public void setNull(int parameterIndex, int sqlType, String typeName) throws SQL
...
+            var rowCounts = new int[commandBatch.size()];
+
+            // MongoDB bulk write API returns row counts grouped by mutation types, not by each command in the batch,
+            // so returns 'SUCCESS_NO_INFO' to work around
+            Arrays.fill(rowCounts, Statement.SUCCESS_NO_INFO);
+
+            return rowCounts;
+
+        } catch (RuntimeException e) {
+            throw new SQLException("Failed to run bulk operation: " + e.getMessage(), e);
```
jyemin, line 359: According to Javadoc, this should throw a `BatchUpdateException`.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -260,6 +269,99 @@ public void setNull(int parameterIndex, int sqlType, String typeName) throws SQL
...
+            }
+            getMongoDatabase()
+                    .getCollection(collectionName, BsonDocument.class)
+                    .bulkWrite(getClientSession(), writeModels);
+
+            var rowCounts = new int[commandBatch.size()];
+
+            // MongoDB bulk write API returns row counts grouped by mutation types, not by each command in the batch,
+            // so returns 'SUCCESS_NO_INFO' to work around
+            Arrays.fill(rowCounts, Statement.SUCCESS_NO_INFO);
```
jyemin, line 354: Not sure if it's worth trying to do better here when possible, but we could fill this in `1` for every insert. Also, I believe that the new client bulkWrite will allow us to return correct results here if we ask for verbose results.  But we would have to be able to detect somehow that we are running on MongoDB 8.0+.

@NathanQingyangXu do you know what Hibernate does with the results of this method?  Does it change anything for the application if we just return `SUCCESS_NO_INFO` here?

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: Hibernate has complicated validation logic for batch updating. But `Statement.SUCCESS_NO_INFO` is a backdoor Hibernate will accept. Here is the relevant code logic in its Expectation class (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/jdbc/Expectations.java#L71):

```
static void checkBatched(int expectedRowCount, int rowCount, int batchPosition, String sql) {
		switch (rowCount) {
			case EXECUTE_FAILED:
				throw new BatchFailedException( "Batch update failed: " + batchPosition );
			case SUCCESS_NO_INFO:
				LOG.debugf( "Success of batch update unknown: %s", batchPosition );
				break;
			default:
				if ( expectedRowCount > rowCount ) {
					throw new StaleStateException(
							"Batch update returned unexpected row count from update " + batchPosition
									+ actualVsExpected( expectedRowCount, rowCount )
									+ " [" + sql + "]"
					);
				}
				else if ( expectedRowCount < rowCount ) {
					throw new BatchedTooManyRowsAffectedException(
							"Batch update returned unexpected row count from update " + batchPosition
									+ actualVsExpected( expectedRowCount, rowCount ),
							expectedRowCount, rowCount, batchPosition );
				}
		}
	}
```
Yeah, for insertion, we could provide more accuate row count, but for updating (esp updateMany model), it might lead to challenge. I am not sure whether we could do better or whether it is worth while.

NathanQingyangXu: @jyemin , btw, I fixed the 'deleted' error by adding complete batch updating integration testing case in the latest commit.

NathanQingyangXu: We could definitively do better. Our metadata implementation opens the way to do mongodb version-specific optimization, so it seems it is perfectly doable, though it seems the current approach might suffice.

There are other improvement possibilities. Like syncing up bulk write option with Hibernate's following two properties (https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#batch-jdbcbatch):

- hibernate.order_updates
- hibernate.order_inserts

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: from the Javadoc, both `MongoBulkWriteException` and `MongoException` could be thrown, right? However, they both are `RuntimeException`, so if some exception is thrown other than the bulk writing API call, we can still ensure the JDBC contract is observed (`SQLException` is thrown). Currently it seems there is no other possible way to trigger exception, so I could change the exception to more generic `MongoException`, but `RuntimeException` seems safer to satisfy JDBC API contract (`SQLException` is thrown, not `RuntimeException`).

jyemin: I'm talking about the Javadoc for `java.sql.Statement#executeBatch`, and actually, I realize it's more complicated than I realized

```
SQLException – if a database access error occurs, this method is called on a closed Statement or the driver does not support batch statements. Throws BatchUpdateException (a subclass of SQLException) if one of the commands sent to the database fails to execute properly or attempts to return a result set.
SQLTimeoutException – when the driver has determined that the timeout value that was specified by the setQueryTimeout method has been exceeded and has at least attempted to cancel the currently running Statement
```

So there's a lot of exception translation that seems to be required.  We can defer to another ticket though, if you'd prefer.

NathanQingyangXu: as per the Javadoc for `executeBatch()`, only one specific `SQLTimeoutException` is mentioned with complicated preconditions as below:
```
     * @throws SQLTimeoutException when the driver has determined that the
     * timeout value that was specified by the {@code setQueryTimeout}
     * method has been exceeded and has at least attempted to cancel
     * the currently running {@code Statement}
```
Both `#setQueryTimeout()` and `#cancel()` should have been tried, if strict JDBC-specific convention is followed.
I checked Hibernate's usage of the `SQLException`. It has an elegant `SQLException`->`JDBCException` mechanism as belows:
```
/**
 * An object that interprets JDBC {@link SQLException}s and converts
 * them to subtypes of Hibernate {@link JDBCException}s.
 *
 * @author Steve Ebersole
 */
public interface SQLExceptionConverter extends Serializable {
	/**
	 * Convert the given {@link SQLException} to a subtype of
	 * {@link JDBCException}.
	 *
	 * @param sqlException The {@code SQLException} to be converted
	 * @param message An optional error message
	 * @param sql The SQL statement that resulted in the exception
	 *
	 * @return The resulting {@code JDBCException}.
	 *
	 * @see org.hibernate.exception.ConstraintViolationException
	 * @see org.hibernate.exception.JDBCConnectionException
	 * @see org.hibernate.exception.SQLGrammarException
	 * @see org.hibernate.exception.LockAcquisitionException
	 */
	JDBCException convert(SQLException sqlException, String message, String sql);
}
```
but it seems unless SQLEXception is related to the 4 listed kinds of JDBCException source, Hibernate's usually end up with a logging only.

NathanQingyangXu: I added the translation logic to map `MongoOperationTimeoutException` to `SQLTimeoutException`, if only for better logging purpose. Seems it is the only special `SQLException` type per the Javadoc.
A unit testing case is added as well.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -354,9 +356,10 @@ public int[] executeBatch() throws SQLException {
             Arrays.fill(rowCounts, Statement.SUCCESS_NO_INFO);
 
             return rowCounts;
-
+        } catch (MongoOperationTimeoutException e) {
```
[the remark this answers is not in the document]
jyemin: Yes, setQueryTimeout seems like a good analog for maxTime and eventually CSOT.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: I saw one of the two goals of https://jira.mongodb.org/browse/HIBERNATE-40 has been exception translation and `SQLTimeoutException` was particularly mentioned in a comment. Seems we can reuse it for the same purpose?

NathanQingyangXu: I've created a DRAFT PR for https://jira.mongodb.org/browse/HIBERNATE-40, including the timeout processing logic: https://github.com/mongodb/mongo-hibernate/pull/40/files#diff-26873092cefd2c1dcee321b6878df0f194d4b160c1eb623a33656e0f0299603d

jyemin:
Yes, we can continue.  Did you mean to re-request reviews?

NathanQingyangXu:
> Yes, we can continue. Did you mean to re-request reviews?

Yeah, I clicked the re-request button then.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
jyemin: Does this need a `TODO HIBERNATE-40` as per our conventions?

jyemin (approved):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
stIncMale: I think, we should not add `TODO` comments, unless they are really needed (I am not sure we follow this principle).

But if we find ourselves in a situation when we do have to leave such a comment, then it has to be in the form of `TODO-<Jira ID>`, and the description of the corresponding Jira ticket should mention that the ticket has these comments, which need to be addressed.

stIncMale (commented):
[I have not reviewed this PR]

We still have `TODO-HIBERNATE-35` comments in the code, but all of them are supposed to be addressed as part of https://jira.mongodb.org/browse/HIBERNATE-35, which this PR implements. What is the plan for those?

I overlooked. Those references should have been deleted for Hibernate only supports batch processing in `PreparedStatement`, not `Statement`.

src/integrationTest/java/com/mongodb/hibernate/jdbc/BatchUpdateIntegrationTests.java
```diff
@@ -0,0 +1,278 @@
...
+                                                    """),
+                                            BsonDocument.parse(
+                                                    """
+                                                    {q: {_id: {$eq: 8}}, limit: 0}
+                                                    """));
+                        },
+                        command4 -> assertThat(command4.getFirstKey()).isEqualTo("commitTransaction"));
+    }
+
+    // TODO-HIBERNATE-19 https://jira.mongodb.org/browse/HIBERNATE-19
```
stIncMale, line 266: This PR adds a `TODO-HIBERNATE-19` not, but https://github.com/mongodb/mongo-hibernate/pull/59 for https://jira.mongodb.org/browse/HIBERNATE-19 has already been created. Now https://github.com/mongodb/mongo-hibernate/pull/59 (HIBERNATE-19) depends on the current PR (https://github.com/mongodb/mongo-hibernate/pull/35 (HIBERNATE-35)), yet they are being worked on concurrently.

If we proceed as is, it is not impossible that we'll merge https://github.com/mongodb/mongo-hibernate/pull/59 (HIBERNATE-19) first, then the currernt PR (https://github.com/mongodb/mongo-hibernate/pull/35 (HIBERNATE-35)), and the `TODO-HIBERNATE-19` added by the current PR will stay in `main` despite HIBERNATE-19 having been resolved.

What is our plan on making sure the above is not going to happen?

NathanQingyangXu: seems we can avoid such deadlock by disabling one PR for the time being until the other one got merged?

NathanQingyangXu: I disabled this PR for the time being so we could focus on #59.

NathanQingyangXu:
closed to avoid conflict with https://github.com/mongodb/mongo-hibernate/pull/59. Will open it when the dependency PR has been merged.

src/integrationTest/java/com/mongodb/hibernate/jdbc/BatchUpdateIntegrationTests.java
stIncMale: I worry that when we create another PR for HIBERNATE-35, it will be based on this one, and the `TODO-HIBERNATE-19` will slip there, despite HIBERNATE-19 having been closed. Can we replace this `TODO-HIBERNATE-19` comment with a `TODO-HIBERNATE-35` comment explaining what needs to be done, and that it should be done after implementing HIBERNATE-19.

NathanQingyangXu: did it. good point.

NathanQingyangXu:
This PR is re-opened for the "basic update translation" is merged so a complete integration testing suite is possible.
@jyemin , the only new thing is the new integration testing case (including 'insertion', 'deletion' and 'update') at https://github.com/mongodb/mongo-hibernate/pull/35/files#diff-208727884b6fd50b4fce3475acee3115f5a5d880432232cf412e3a37ee48a235

PR #37, opened 2025-02-07
Do miscellaneous improvements

stIncMale:


src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -63,10 +64,11 @@ final class MongoPreparedStatement extends MongoStatement implements PreparedSta
...
     private final List<Consumer<BsonValue>> parameterValueSetters;
 
-    public MongoPreparedStatement(
-            MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection, String mql) {
+    MongoPreparedStatement(
+            MongoClient mongoClient, ClientSession clientSession, MongoConnection mongoConnection, String mql)
+            throws SQLSyntaxErrorException {
         super(mongoClient, clientSession, mongoConnection);
-        this.command = BsonDocument.parse(mql);
+        this.command = parse(mql);
```
stIncMale, line 71: Here we parse `BsonDocument` the same exact way `MongoStatement` does that, which involves throwing `SQLSyntaxErrorException`.

src/main/java/com/mongodb/hibernate/internal/NotYetImplementedException.java
```diff
@@ -22,8 +22,8 @@
  * A temporary marker exception to denote that the feature in question is in the scope of MongoDB dialect but has not
  * been implemented yet.
  *
- * <p>Ultimately all of its references should be eliminated sooner or later, and then this class is supposed to be
- * deleted prior to product release.
+ * <p>TODO-HIBERNATE-45 Ultimately all of its references should be eliminated sooner or later, and then this class is
```
stIncMale, line 25: The ticket: https://jira.mongodb.org/browse/HIBERNATE-45

NathanQingyangXu: tbh, the ticket is not quite necessary for whenever we finished an undone feature, defintively we will remove the corresponding exception. If any feature is not finished, I am afraid we still need the exception linger there.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: if we don't have a hard rule, we hope we have some flexibiliity. `parse()` is a common method name and I think giving it a context improves code reading and maintainance.

NathanQingyangXu (dismissed):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
stIncMale: I qualified the call in 9dbc8028f8a7b43853054e7b227494a4681e73a9.

src/main/java/com/mongodb/hibernate/internal/NotYetImplementedException.java
stIncMale: This task will server as a reminder that we have unfinished business, and may eventually push us either towards implementing everything where this exception is thrown, or deciding that some of those places are not to be implemented at all, or coming up with a different more permanent solution.

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
NathanQingyangXu: thanks. Another compelling example is the following common usage:
```
var map = Maps.of("key1", "value1", "key2", "value2");
```

NathanQingyangXu (approved):

PR #39, opened 2025-02-12
configure errorprone scope to `compileJava` alone

NathanQingyangXu:
minor gradle improvement. Previously we exclude errorprone usage from unit testing, but it still applies to integration testing.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -95,13 +95,13 @@ spotless {
 }
 
 tasks.withType<JavaCompile>().configureEach {
+    options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))
     if (name == "compileJava") {
         options.errorprone {
             disableWarningsInGeneratedCode.set(true)
             option("NullAway:AnnotatedPackages", "com.mongodb.hibernate")
             error("NullAway")
         }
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -95,20 +95,17 @@ spotless {
...
+    if (name == "compileJava") {
+        options.errorprone {
+            disableWarningsInGeneratedCode.set(true)
+            option("NullAway:AnnotatedPackages", "com.mongodb.hibernate")
+            error("NullAway")
+        }
+        options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))
+    } else {
+        options.errorprone.isEnabled.set(false)
     }
```
stIncMale, line 107: Let's do `options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))` for all compile tasks, not only for production code

```suggestion
    options.compilerArgs.addAll(listOf("-Xlint:deprecation", "-Werror"))
    if (name == "compileJava") {
        options.errorprone {
            disableWarningsInGeneratedCode.set(true)
            option("NullAway:AnnotatedPackages", "com.mongodb.hibernate")
            error("NullAway")
        }
    } else {
        options.errorprone.isEnabled.set(false)
    }
```

stIncMale (changes requested):

stIncMale (approved):

PR #41, opened 2025-02-12
Add multiline string formatter

katcharov:
The existing formatter does not catch multiline string indentation inconsistencies ([example](https://github.com/mongodb/mongo-hibernate/blob/cd5bdba521ac688af8a6a757f59d3cfd74cb07f8/src/integrationTest/java/com/mongodb/hibernate/jdbc/MongoStatementIntegrationTests.java#L79-L92)). These show up a fair bit in tests.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
+import com.diffplug.spotless.FormatterFunc
+import com.diffplug.spotless.FormatterStep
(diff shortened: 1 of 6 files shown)


NathanQingyangXu:
thanks. Reality is the `Palantir` did a sloppy job on auto-formatting multi-line text block, which seems so important to your testing case. A timely PR.

NathanQingyangXu (approved):

PR #43, opened 2025-02-15
fix IDEA bug labelling integration testing folder not as test source sets

NathanQingyangXu:
It is a bug not fixed in IDEA: https://youtrack.jetbrains.com/issue/IDEA-234382/Gradle-integration-tests-are-not-marked-as-test-sources-resources

Currently fix the issue by idea plugin


NathanQingyangXu:
@rozza , this might be inspirting to your work as well; feel free to review this PR.

build.gradle.kts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("UnstableApiUsage")
+
```
NathanQingyangXu, line 18: the above is to suppress the yellow warning in the below snippet:
```
integrationTestImplementation(libs.hibernate.testing) {
    }
```

build.gradle.kts
```diff
@@ -44,29 +47,36 @@ sourceSets {
...
 
-        testClassesDirs = sourceSets["integrationTest"].output.classesDirs
-        classpath = sourceSets["integrationTest"].runtimeClasspath
-        shouldRunAfter("test")
+        testClassesDirs = integrationTestSourceSet.output.classesDirs
+        classpath = integrationTestSourceSet.runtimeClasspath
 
         useJUnitPlatform()
-
-        testLogging { events("passed") }
```
NathanQingyangXu, line 63: the above comes from gradle doc. I doubt we need it for one thing it lacks "skipped" and "failed"; for another thing, if the `test` task doens't specify it, why not do the same thing for `integration test`?

build.gradle.kts
```diff
@@ -44,29 +46,37 @@ sourceSets {
...
+    useJUnitPlatform()
+    testLogging { events(TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.FAILED) }
+}
 
-        testLogging { events("passed") }
+// https://youtrack.jetbrains.com/issue/IDEA-234382/Gradle-integration-tests-are-not-marked-as-test-sources-resources
+idea {
+    module {
+        testSources.from(integrationTestSourceSet.allSource.srcDirs)
+        testResources.from(integrationTestSourceSet.resources.srcDirs)
```
stIncMale, line 74: This is supposed to be needed only for Gradle < 7.3, as the bug was fixed in 7.3. But that, apparently, is not the case.

NathanQingyangXu: that might be because of some misleading comments in the bug ticket, but the issue is still there for sure.

NathanQingyangXu:
> Is it possible to suppress that waring only or the line/block specified, and not the whole file?

Yeah, it could. Just like in Java. Updated.

stIncMale (approved):

PR #44, opened 2025-02-18
basic deletion translation

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-17

The overlooked `mongoast` unit testing cases are added. Straightforward implementation touching one or two visiter methods; as discussed, `optimistic locking` is excluded from the scope for lack of verification (error case) possibility for now.


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -203,6 +214,39 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
 
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Table Mutation: deletion
+
+    @Override
+    public void visitStandardTableDelete(TableDeleteStandard tableDeleteStandard) {
+        if (tableDeleteStandard.getWhereFragment() != null) {
+            // only applies when the table is a collection table with where restriction
+            throw new FeatureNotSupportedException();
+        }
```
NathanQingyangXu, line 224: The above feature only applies for collection table (e.g. `one-to-many` or `many-to-many` association table), so it is far ahead of what we can verify (we need to finish table joining first, which might be the last task in the last milestone). Without verificaation, it is easy to implement it for the where fragment here is some native MQL filter.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -203,6 +214,39 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+
+        if (tableDeleteStandard.getNumberOfKeyBindings() > 1) {
+            throw new FeatureNotSupportedException("MongoDB doesn't support id spanning across multiple columns");
+        }
```
NathanQingyangXu, line 228: The only possible scenario that id spanning across multiple columns is `composite id` based on `@Embeddable` implemented by flattenning the composite id to multiple columns. This seems to conflict with MongoDB's feature (`_id` column has only one and can't be shared), so we can safely throw exception here.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -203,6 +214,39 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+
+        astVisitorValueHolder.yield(COLLECTION_MUTATION, new AstDeleteCommand(tableName, new AstAndFilter(filters)));
+    }
+
+    private AstFilter getAstFieldFilter(ColumnValueBinding columnValueBinding) {
+        var astFilterField =
+                new AstFilterField(columnValueBinding.getColumnReference().getColumnExpression());
+        if (columnValueBinding.getValueExpression() == null) {
+            return new AstFieldOperationFilter(astFilterField, new AstComparisonFilterOperation(EQ, NULL));
+        }
```
NathanQingyangXu, line 246: This is confusing. How come the static column value expression is null? Well, it is possible, to accommodate the edge case that in versionless optimistic locking scenario (see https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#locking-optimistic-versionless), some field value is null during deletion. It will trigger dynamic deletion translation as opposed to static deletion MQL.

I provided a testing case to showcase its necessity (https://github.com/mongodb/mongo-hibernate/pull/44/files#diff-9d97b68ca93ea3fee5c3f3aa6dc5838ffc65903c76f343672f22dfce806b47f6R189)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -203,6 +214,39 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+        assertTrue(tableDeleteStandard.getNumberOfKeyBindings() == 1);
+
+        var tableName = tableDeleteStandard.getTableName();
+
+        var filters = new ArrayList<AstFilter>(1 + tableDeleteStandard.getNumberOfOptimisticLockBindings());
+        filters.add(getAstFieldFilter(tableDeleteStandard.getKeyBindings().get(0)));
+        tableDeleteStandard.forEachOptimisticLockBinding(
+                (idx, optimisticLockBinding) -> filters.add(getAstFieldFilter(optimisticLockBinding)));
+
+        astVisitorValueHolder.yield(COLLECTION_MUTATION, new AstDeleteCommand(tableName, new AstAndFilter(filters)));
```
NathanQingyangXu, line 238: We need to accommodate optimistic locking fields; otherwise the important optimisitc locking mechanism (https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#locking-optimistic) won't work.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudTests.java
```diff
@@ -0,0 +1,292 @@
...
+
+@SessionFactory(exportSchema = false)
+@DomainModel(
+        annotatedClasses = {
+            BasicCrudTests.Book.class,
+            BasicCrudTests.BookWithEmbeddedField.class,
+            BasicCrudTests.BookOptimisticLockWithVersion.class,
+            BasicCrudTests.BookOptimisticLockWithoutVersion.class
+        })
+class BasicCrudTests implements SessionFactoryScopeAware {
```
NathanQingyangXu, line 57: I refactored the previous `BasicInsertionTests` to accommodate all the `basic CRUD testing` cases. `hibernate-testing` library doesn't integrate with JUnit 5 features elegantly (e.g. canonical usage requires passing `SessionFactoryScope` as the only testing method parameter, which conflicts with `@ParameterizedTests` in the first place), but it provides `SessionFactoryScopeAware` to work around.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: A good resource showcasing how multiple `keyBindings` is used is at https://github.com/hibernate/hibernate-orm/tree/main/hibernate-core/src/test/java/org/hibernate/orm/test/onetoone/embeddedid

Below is a typical `EmbeddedId` usage (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/test/java/org/hibernate/orm/test/onetoone/embeddedid/OneToOneJoinColumnsEmbeddedIdTest.java#L96):
```
        @Entity(name = "EntityA")
	public static class EntityA {

		@EmbeddedId
		private EntityAKey entityAKey;

		private String name;

		@OneToOne(mappedBy = "entityA", fetch = FetchType.LAZY)
		private EntityB entityB;

		public EntityA() {
		}

		public EntityA(EntityAKey key, String name) {
			this.entityAKey = key;
			this.name = name;
		}

		public EntityAKey getEntityAKey() {
			return entityAKey;
		}

		public String getName() {
			return name;
		}

		public EntityB getEntityB() {
			return entityB;
		}
	}

	@Embeddable
	public static class EntityAKey {

		@Column(name = "id1")
		private Integer id1;

		@Column(name = "id2")
		private String id2;

		public EntityAKey() {
		}

		public EntityAKey(Integer id1, String id2) {
			this.id1 = id1;
			this.id2 = id2;
		}
	}
```
so two columns will be combined to act as primary key (traditional `@Embeddable` usage), which has no counterpart in MongoDB.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstComparisonFilterOperation.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast.filter;
+
+import com.mongodb.hibernate.internal.translate.mongoast.AstValue;
+import org.bson.BsonWriter;
+
+public record AstComparisonFilterOperation(AstComparisonFilterOperator operator, AstValue value)
```
jyemin, line 22: In match syntax, the value of a comparison filter has to be a constant, e.g. `1` or `"1"`.  In the C# implementation this is enforced by requiring `value` to be of type `BsonValue`.  We can't quite do the same here because we also have to support `AstPlaceholder`, but perhaps we can introduce a super-interface that both `AstPlaceholder` and `AstLiteralValue` extend, and use that as the generic type for `AstVisitorValueDescriptor#FIELD_VALUE`.

jyemin (commented):
Looking good.  Just noticed one thing while comparing with the C# Linq AST.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstComparisonFilterOperation.java
NathanQingyangXu: Just a naive questoin. Why did we introduce `AstValue` in the first place? If we used `BsonValue` directly as in C#, it seems to simplify a lot. Maybe `AstPlaceHolder` (now renamed to `AstParameterMarker`) is the reason?

NathanQingyangXu: but if we treat `AstParameterMarker` as a synonym to `new BsonUndefined()`, seems it works. From my perspective, there is no need to make it inherit `AstNode` (or any other `AstValuel`), for the criterion is whether we want to render it specially. `BsonValue` rendering seems straightforward without need of special handling?

NathanQingyangXu: > In match syntax, the value of a comparison filter has to be a constant, e.g. `1` or `"1"`. In the C# implementation this is enforced by requiring `value` to be of type `BsonValue`. We can't quite do the same here because we also have to support `AstPlaceholder`, but perhaps we can introduce a super-interface that both `AstPlaceholder` and `AstLiteralValue` extend, and use that as the generic type for `AstVisitorValueDescriptor#FIELD_VALUE`.

Technically speaking, I think `AstPlaceHolder` is kind of `AstLiteralValue` from translating perspective (its special meaning only manifests itself in JDBC `PreparedStatement` layer). If we do this, we could simply by using the generic `AstLiteralValue` for `AstVisitorValueDescriptor#FIELD_VALUE` elegantly.

jyemin: Oh, I'm so sorry.  This already works exactly as I wanted.  I must have mis-read `AstValue` as `AstNode`.  `AstValue` is exactly the supertype I was asking for.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -217,11 +223,35 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+
+        if (tableDeleteStandard.getNumberOfKeyBindings() > 1) {
+            throw new FeatureNotSupportedException("MongoDB doesn't support id spanning multiple columns");
+        }
+        assertTrue(tableDeleteStandard.getNumberOfKeyBindings() == 1);
+        var keyBinding = tableDeleteStandard.getKeyBindings().get(0);
+
+        var tableName = tableDeleteStandard.getMutatingTable().getTableName();
+        var astFilterField = new AstFilterField(keyBinding.getColumnReference().getColumnExpression());
+        var astValue = acceptAndYield(keyBinding.getValueExpression(), FIELD_VALUE);
```
stIncMale, line 246: Here we read `getColumnExpression` at first, then `getValueExpression`. Let's have he same order in the `visitStandardTableInsert` method, to simplify comparing the logic. Currently, that method reads `getValueExpression` at first, then `getColumnExpression`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -217,11 +223,35 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
     }
 
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Table Mutation: deletion
+
+    @Override
+    public void visitStandardTableDelete(TableDeleteStandard tableDeleteStandard) {
```
stIncMale, line 229: We talked about `TableDeleteStandard.getMutationComment` during the code walkthrough, but I don't remember what we decided about it.

If we silently ignore it now here, we may never remember that we are not handling it properly. We also can't throw `FeatureNotSupportedException` if this comment is present, because it is always present, as it is generated by Hibernate ORM.

The best way forward appears to be creating a ticket for supporting comments, which we know may be given to us either via `TableDeleteStandard.getMutationComment`, or via `QueryOptions.getComment`.

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/jdbc/adapter/package-info.java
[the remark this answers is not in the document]
NathanQingyangXu: my bad. It is leftover from other PRs and I included it by accident.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstFilterField.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast.filter;
+
+import com.mongodb.hibernate.internal.translate.mongoast.AstNode;
+import org.bson.BsonWriter;
+
+public record AstFilterField(String path) implements AstNode {
```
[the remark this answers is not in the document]
NathanQingyangXu: I agree `AstFilterFieldPath` seems more accurate

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstComparisonFilterOperator.java
```diff
@@ -0,0 +1,31 @@
...
+package com.mongodb.hibernate.internal.translate.mongoast.filter;
+
+public enum AstComparisonFilterOperator {
+    EQ("$eq");
+
+    AstComparisonFilterOperator(String operatorName) {
+        this.operatorName = operatorName;
+    }
+
+    public String operatorName() {
```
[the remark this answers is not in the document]
NathanQingyangXu: of course you are right. I ovrerlooked for I copied it somewhere without much inspection.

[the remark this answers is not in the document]
NathanQingyangXu: I changed it in the same commit solving the issue before this (renaming method name)

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstFilter.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast.filter;
+
+import com.mongodb.hibernate.internal.translate.mongoast.AstNode;
+
+public interface AstFilter extends AstNode {}
```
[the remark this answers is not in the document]
NathanQingyangXu: we need it for sure. Very soon we need another `AstAndFilter` (see https://github.com/mongodb/mongo-csharp-driver/blob/main/src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstAndFilter.cs). There are other possibilities and you can check it out from C# codebase at https://github.com/mongodb/mongo-csharp-driver/tree/main/src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters

src/integrationTest/java/com/mongodb/hibernate/BasicCrudTests.java
```diff
@@ -0,0 +1,219 @@
...
+                session.remove(book);
+            });
+
+            // then
+            assertThat(getCollectionDocuments()).isEmpty();
+        }
+    }
+
+    private void onMongoCollection(Consumer<MongoCollection<BsonDocument>> collectionConsumer) {
+        var connectionString = new ConnectionString(new Configuration().getProperty(JAKARTA_JDBC_URL));
```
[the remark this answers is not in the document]
NathanQingyangXu: adopted

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -201,7 +207,7 @@ public void visitStandardTableInsert(TableInsertStandard tableInsertStandard) {
         var tableName = tableInsertStandard.getTableName();
         var astElements = new ArrayList<AstElement>(tableInsertStandard.getNumberOfValueBindings());
         for (var columnValueBinding : tableInsertStandard.getValueBindings()) {
-            var astValue = acceptAndYield(columnValueBinding.getValueExpression(), FIELD_VALUE);
+            var astValue = acceptAndYield(assertNotNull(columnValueBinding.getValueExpression()), FIELD_VALUE);
```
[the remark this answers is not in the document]
NathanQingyangXu: I added because `ColumnValueBinding#getValueExpression()` could return `null`. An example is in optimistic locking scenario (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java#L9021): 

Screenshot 2025-02-26 at 10 29 04 AM

so at least in theory, we are not sure whether we will end up NPE. Hibernate doesn't use nullness annotations extensively (they had started using checker framework but gradually and SQL Ast model is not covered).

I will remove this dubious usage (seems the above expression won't return null value for `insertion` case but I am not 100% sure); but I wanna raise the awareness that our `NullAway` could only protect us in our code and potential NPE is still possible.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: I created a ticket at https://jira.mongodb.org/browse/HIBERNATE-53

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstNodeRenderTests.java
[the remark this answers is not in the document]
NathanQingyangXu: done. Let me know whether we could improve further. I covered two commands and some filters classes. Guess that should be enough for other lower-level renderings have been covered by them indirectly.

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AbstractAstNodeTests.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import org.bson.json.JsonWriter;
+
+public abstract class AbstractAstNodeTests {
+
+    protected void assertJsonRendered(AstNode node, String expectedJson) {
```
stIncMale, line 27: Let's reorder the parameters to be consistent with the parameter order in JUnit5: expected (`expectedJson`) comes first, and the actual (`node`) comes second.

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstFieldOperationFilterTests.java
```diff
@@ -0,0 +1,41 @@
...
+import com.mongodb.hibernate.internal.translate.mongoast.AstLiteralValue;
+import org.bson.BsonInt32;
+import org.junit.jupiter.api.Test;
+
+class AstFieldOperationFilterTests extends AbstractAstNodeTests {
+
+    @Test
+    void testRendering() {
+        // given
+        var astFilterFieldPath = new AstFilterFieldPath("$fieldPath");
```
stIncMale, line 31: `AstFilterFieldPath` does not check or expect anything from the `path`, but at least for now, the way `path` is specified is by specifying just the field name. So maybe let's use just the "fieldName" string literal instead of "$fieldPath" in this test?

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/AstDeleteCommandTests.java
```diff
@@ -0,0 +1,48 @@
...
+        var filter = new AstFieldOperationFilter(
+                new AstFilterFieldPath("$isbn"),
+                new AstComparisonFilterOperation(EQ, new AstLiteralValue(new BsonString("978-3-16-148410-0"))));
+
+        var deleteCommand = new AstDeleteCommand(collection, filter);
+
+        // when && then
+        var expectedJson =
+                """
+                {"delete": "books", "deletes": [{"q": {"$isbn": {"$eq": "978-3-16-148410-0"}}, "limit": 0}]}\
```
stIncMale, line 44: There seem to be no reason for using text blocks in `AstDeleteCommandTests`, `AstInsertCommandTests`, `AstComparisonFilterOperationTests`, `AstFieldOperationFilterTests`, as all of them contain only one. Let's just use single-line string literals?

stIncMale (changes requested):
The last reviewed commit is 32f4fa2b349c464f1536e793b05334da82bbb72b.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
NathanQingyangXu: ok. Luckily we are sure that `keyBinding.keyBinding.getColumnReference().getColumnExpression()` in deletion case is never null!

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstFieldOperationFilterTests.java
NathanQingyangXu: definitively

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/AstDeleteCommandTests.java
NathanQingyangXu: The reason is to avoid the cluttering escaping of double quotes. So below is the new one-liner:
```
"{\"delete\": \"books\", \"deletes\": [{\"q\": {\"isbn\": {\"$eq\": \"978-3-16-148410-0\"}}, \"limit\": 0}]}"
```
for simple case, it is fine.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -184,7 +184,7 @@ public DatabaseMetaData getMetaData() throws SQLException {
             var versionArray = commandResult.getList("versionArray", Integer.class);
             if (versionArray.size() < 2) {
                 throw new SQLException(
-                        format("Unexpected versionArray [%s] field length (should be 2 or more)", versionArray));
+                        format("Unexpected versionArray [%s] fieldPath length (should be 2 or more)", versionArray));
```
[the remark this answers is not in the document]
jyemin: Use Intellij's Rename refactoring instead of string-based search and replace and you'll avoid inadvertent changes like this.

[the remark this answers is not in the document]
NathanQingyangXu: Yeah, initially I did use refactoring, but then it is also desirable to change variable names which only shows up in code text, and that is when I had to use search and replace. Just didn't pay due attention.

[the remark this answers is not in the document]
stIncMale: > it is also desirable to change variable names which only shows up in code text

Idea's "rename" refactoring addresses those. And when there are text occurrences it suspect need to be changed, it shows the the preview of the refactoring, allowing you to choose what changes to exclude from refactoring.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -197,12 +203,17 @@ <R extends AstNode> R acceptAndYield(SqlAstNode node, AstVisitorValueDescriptor<
...
+    public void visitStandardTableInsert(TableInsertStandard tableInsert) {
+        var tableName = tableInsert.getTableName();
+        var astElements = new ArrayList<AstElement>(tableInsert.getNumberOfValueBindings());
+        for (var columnValueBinding : tableInsert.getValueBindings()) {
             var columnExpression = columnValueBinding.getColumnReference().getColumnExpression();
+
+            if (columnValueBinding.getValueExpression() == null) {
+                throw new FeatureNotSupportedException();
+            }
+            var astValue = acceptAndYield(columnValueBinding.getValueExpression(), FIELD_VALUE);
```
stIncMale, line 215: Let's rewrite the same as proposed below. This way it is obvious that `valueExpression` can't be `null` past the `if` block, and also that we pass the same value we have just checked for `null` to `acceptAndYield`.

```suggestion
            var valueExpression = columnValueBinding.getValueExpression();
            if (valueExpression == null) {
                throw new FeatureNotSupportedException();
            }
            var astValue = acceptAndYield(valueExpression, FIELD_VALUE);
```

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/AstDeleteCommandTests.java
stIncMale: I did not think about this. But now, when you presented this reason, I agree that we should use multiline blocks to avoid having to escape `"`.

Suggestion for more efficient review process: if you have an objection to a proposal, or suspect that a reviewer may have failed to consider something important when making a proposal, it is better to bring that up (for example, a as a reply in the PR conversation), and see what comes from it, rather than making a change right away, and then undoing it.

stIncMale (changes requested):
The last reviewed commit is 395d00ecc293d3b5a5be22f7392b3dbd8444f4f9.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudTests.java
```diff
@@ -0,0 +1,216 @@
...
+    void beforeEach() {
+        var config = new MongoConfigurationBuilder(
+                        sessionFactoryScope.getSessionFactory().getProperties())
+                .build();
+        if (mongoClient == null) {
+            mongoClient = MongoClients.create(config.mongoClientSettings());
+        }
+        collection = mongoClient.getDatabase(config.databaseName()).getCollection("books", BsonDocument.class);
+        collection.drop();
+    }
```
stIncMale, line 69: [I am leaving this comment just to make sure we don't forget to make the change]

@NathanQingyangXu, as you proposed, we should make `mongoClient` an instance field, and initialize it in an instance `@BeforeAll` method. The latter also applies to the `collection` field. The only thing we need to do in the `@BeforeEach` method is to call `collection.drop();`

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/AstDeleteCommandTests.java
NathanQingyangXu: well, tbh, for one-liner I didn't think it makes much difference, as I mentioned above. but thanks for your suggestion. will keep that in mind in future.

stIncMale (approved):

stIncMale (commented):
The last reviewed commit is 4585e82fe6b4c62e8e8824048eba71261e62b661.

PR #46, opened 2025-02-19
Use non-existent host name in unit tests

stIncMale:
This way we cannot accidentally write a unit test that needs a successful connection to a MongoDB deployment.

### DIFF
--- a/src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
+++ b/src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
@@ -63,7 +63,7 @@ class MongoConnectionProviderTests {
 
             // given
 
-            var hostInConnectionString = customizerAppliesConnectionString ? "my-host" : "localhost";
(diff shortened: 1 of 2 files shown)


src/test/java/com/mongodb/hibernate/jdbc/MongoConnectionProviderTests.java
```diff
@@ -63,7 +63,7 @@ void testMongoClientCustomizerTakeEffect(boolean customizerAppliesConnectionStri
 
             // given
 
-            var hostInConnectionString = customizerAppliesConnectionString ? "my-host" : "localhost";
+            var hostInConnectionString = "host";
```
stIncMale, line 66: As far as I can tell, there is no need to use two different values here.

NathanQingyangXu: That is a leftover I forgot to remove when I refactored the testing logic to avoid internal class reliance.

NathanQingyangXu (approved):

PR #49, opened 2025-02-24
set id field name to "_id"

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-25

A self-contained PR focusing on making id field name hard-coded as `_id`.

During implementation, it is found hard to satisfy the following AC:

```
But if primary key column name has been specified as below example:

@Entity
class Book {
    @Id
    @Column(name = "id")
    int id;
} 
We have to violate user's intention by throwing exception to explain that MongoDB has its hardcoded field name.
```

The reason is Hibernate doesn't expose such metadata publicly in the extension point currently used. Seems another extension point or contributor needs to be exploited. Also, the column spec could come from multiple sources, including:

1. JPA entity field annotation
2. hbm traditional config
3. Jarkarta XML config

So it seems complicated than originally expected. For that reason, a separate ticket would be created and this PR only focuses on one thing. Set id column name to `_id` regardless.


NathanQingyangXu:
Found a better implementation plan, so close this PR for the time being.

PR #50, opened 2025-02-25
update diagram in README.md and other minor improvements

NathanQingyangXu:
Update the diagram (changing `MongoDB Dialect` to `MongoDB Extension` and some other minor improvements


build.gradle.kts
```diff
@@ -148,14 +147,14 @@ class MultilineFormatter : Serializable {
...
-    } else {
-        options.errorprone.isEnabled.set(false)
+    when (this) {
+        tasks.compileJava.get() ->
+            options.errorprone {
+                disableWarningsInGeneratedCode = true
+                option("NullAway:AnnotatedPackages", "com.mongodb.hibernate")
+                error("NullAway")
+            }
+        else -> options.errorprone.isEnabled = false
```
NathanQingyangXu, line 157: recently I am learning Kotlin language. The above changes are to make our Kotlin usage more idiomatic.

build.gradle.kts
```diff
@@ -187,8 +185,6 @@ dependencies {
         exclude(group = "org.apache.logging.log4j", module = "log4j-core")
     }
 
-    integrationTestRuntimeOnly(libs.junit.platform.launcher)
-
```
NathanQingyangXu, line 191: the above JUnit dependencies might conflict with the ones brought in by `hibernate-testing`. For that reason, we can simplify by keeping from specifying them in the first place.

stIncMale (changes requested):

build.gradle.kts
stIncMale: If we have a dependency, and Hibernate ORM happens to also have it, we should not avoid specifying that dependency in our `build.gradle.kts` because of a possible version conflict. If this conflict happens, we should resolve it in an appropriate way.

It is even more surprising that we specify the version of the `org.junit.platform:junit-platform-launcher` dependency, but do not specify this dependency.

NathanQingyangXu: I reverted back the changes related to JUnit. I found bumping JUnit dependencies to latest version would break our integration testing due to version conflicts. It takes time to fix it but I don't think it is worth while currently.

NathanQingyangXu:
I also added a `gradle.properties` to override the default JVM options in evergreen. So the following warnings won't show up in evergreen logs:
Screenshot 2025-02-25 at 10 14 02 AM

stIncMale (approved):

PR #51, opened 2025-02-26
minor improvements to gradle usage

NathanQingyangXu:
- add `eclipse` plugin usage (if `idea` has been used)
- remove unnecessary junit launcher dependency (`junit-jupiter` is a bom dependency which will include it transitively)
- eliminate dependencies duplication across `test` and `integrationTest` tasks

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,6 +23,7 @@ import org.gradle.api.tasks.testing.logging.TestLogEvent
 version = "1.0.0-SNAPSHOT"
 
 plugins {
+    eclipse
     idea
     `java-library`
     alias(libs.plugins.spotless)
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -20,7 +20,6 @@ buildconfig = "5.5.1"
 
 [libraries]
 junit-jupiter = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junit-jupiter" }
-junit-platform-launcher = { module = "org.junit.platform:junit-platform-launcher" }
```
NathanQingyangXu, line 23: we don't need the above for `junit-jupiter` is enough. See example usage of JUnit dependency in gradle at https://github.com/junit-team/junit5-samples/blob/r5.4.0/junit5-jupiter-starter-gradle-kotlin/build.gradle.kts

stIncMale (changes requested):
The changes in the PR cause build warnings. To see them, we can use `--warning-mode all`. So when I run `./gradlew clean spotlessApply check compileJava --warning-mode all`, I see

> Task :test
The automatic loading of test framework implementation dependencies has been deprecated. This is scheduled to be removed in Gradle 9.0. Declare the desired test framework directly on the test suite or explicitly declare the test framework implementation dependencies on the test's runtime classpath. Consult the upgrading guide for further information: https://docs.gradle.org/8.10.2/userguide/upgrading_version_8.html#test_framework_implementation_dependencies

I am not sure which of the changes causes it, but we don't have it in `main`.

build.gradle.kts
```diff
@@ -170,24 +171,21 @@ buildConfig {
...
 // Dependencies
 
+listOf(libs.junit.jupiter, libs.assertj, libs.logback.classic).forEach {
+    dependencies {
+        testImplementation(it)
+        integrationTestImplementation(it)
+    }
+}
+
 dependencies {
```
stIncMale, line 181: Can we move this inside the `dependencies` block as follows?

```suggestion
dependencies {
    listOf(libs.junit.jupiter, libs.assertj, libs.logback.classic).forEach {
        testImplementation(it)
        integrationTestImplementation(it)
    }
```

NathanQingyangXu: updated with inclusion of the missing junit-platform-launcher

NathanQingyangXu:
> The changes in the PR cause build warnings. To see them, we can use `--warning-mode all`. So when I run `./gradlew clean spotlessApply check compileJava --warning-mode all`, I see
> 
> > Task :test
> > The automatic loading of test framework implementation dependencies has been deprecated. This is scheduled to be removed in Gradle 9.0. Declare the desired test framework directly on the test suite or explicitly declare the test framework implementation dependencies on the test's runtime classpath. Consult the upgrading guide for further information: https://docs.gradle.org/8.10.2/userguide/upgrading_version_8.html#test_framework_implementation_dependencies
> 
> I am not sure which of the changes causes it, but we don't have it in `main`.

Good catch! The reason is Gradle requires either explicit junit platform launcher inclusion or test suite plugin (which is in incubating state). I reverted it back.

I also figured out the `hibernate-testing` being broken issue. Internally it has `api` dependency on Junit API v5.11.4, and new version's API changes ending up with runtime error. Just sticking to its JUnit API version works.

gradle/libs.versions.toml
NathanQingyangXu: It is not 100% correct for Gradle v9 will require explicit testing platform spec

stIncMale (approved):

PR #53, opened 2025-02-26
simplify testing code by use of @AutoClose annotation from JUnit 5

NathanQingyangXu:
seems an improvement to avoid tedious and error-prone closing code logic. Applies to both static and non-static fields. I tested locally by setting breakpoint in debug mode and found it is working as expected.


stIncMale (approved):
This is so cool!

PR #59, opened 2025-03-02
basic update translation

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-19

a straightforward PR


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -229,30 +233,54 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+        assertTrue(tableMutation.getNumberOfKeyBindings() == 1);
+        var keyBinding = tableMutation.getKeyBindings().get(0);
 
-        var tableName = tableDelete.getMutatingTable().getTableName();
         var astFilterFieldPath =
                 new AstFilterFieldPath(keyBinding.getColumnReference().getColumnExpression());
         var astValue = acceptAndYield(keyBinding.getValueExpression(), FIELD_VALUE);
-        var keyFilter = new AstFieldOperationFilter(astFilterFieldPath, new AstComparisonFilterOperation(EQ, astValue));
-        astVisitorValueHolder.yield(COLLECTION_MUTATION, new AstDeleteCommand(tableName, keyFilter));
+        return new AstFieldOperationFilter(astFilterFieldPath, new AstComparisonFilterOperation(EQ, astValue));
```
NathanQingyangXu, line 283: the common id filter is refactored into its own method for it applies to both deletion and updating with no difference (both mutation classes inherits from the common `RestrictedTableMutation` class)

src/main/java/com/mongodb/hibernate/internal/translate/TableMutationMqlTranslator.java
```diff
@@ -38,20 +36,10 @@ final class TableMutationMqlTranslator<O extends JdbcMutationOperation> extends
...
 
-        if (tableMutation instanceof TableInsert || tableMutation instanceof TableDelete) {
-            return translateTableMutation();
-        } else {
-            // TODO-HIBERNATE-19 https://jira.mongodb.org/browse/HIBERNATE-19
-            return (O) new NoopJdbcMutationOperation();
-        }
-    }
-
-    private O translateTableMutation() {
```
NathanQingyangXu, line 54: finally we finished TableMutation so the temporary adapter is not needed

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstDocument.java
```diff
@@ -19,7 +19,7 @@
 import java.util.List;
 import org.bson.BsonWriter;
 
-public record AstDocument(List<? extends AstElement> elements) implements AstValue {
+public record AstDocument(List<AstElement> elements) implements AstValue {
```
NathanQingyangXu, line 22: we don't need such lowerbound generics feature for `Record` is final

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
```diff
@@ -141,33 +142,61 @@ void testEntityWithEmbeddedFieldInsertion() {
...
+            book.id = 1;
+            book.title = "War and Peace";
+            book.author = "Leo Tolstoy";
+            book.publishYear = 1867;
+            session.persist(book);
+            session.flush();
+
+            book.title = "Insurrection";
+            book.publishYear = 1899;
+        });
```
NathanQingyangXu, line 185: from Hibernate javadoc, you might see 'merge' method is recommended in preference to `update`; but merge usage will require other translation logic. Not sure whether it should be in the scope of this ticket. If the update is the last action within a session, nothing needs to be done further.

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
```diff
@@ -170,6 +172,46 @@ void testSimpleDeletion() {
...
+            assertThat(statementInspector.getSqlQueries())
+                    .containsExactly(
+                            """
+                            {"insert": "books", "documents": [{"author": {"$undefined": true}, "publishYear": {"$undefined": true}, "title": {"$undefined": true}, "_id": {"$undefined": true}}]}\
+                            """,
+                            """
+                            {"update": "books", "updates": [{"q": {"_id": {"$eq": {"$undefined": true}}}, "u": {"$set": {"author": {"$undefined": true}, "publishYear": {"$undefined": true}, "title": {"$undefined": true}}}, "multi": true}]}\
+                            """);
+        }
+    }
```
NathanQingyangXu, line 213: to prove that the udpating did happen (not one sinlge insertion with latest state), a good feature of hibernate-testing was used here to verify. Note that by default table mutation static update SQL will include all fields to for perf reason

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
```diff
@@ -170,6 +172,46 @@ void testSimpleDeletion() {
         }
     }
 
+    @Nested
+    class UpdateTests {
```
NathanQingyangXu, line 176: `@Nested` is used to accommodate the future new testing case after `optimistic locking` feature is taken into consideration, even though currently there is only one testing case in this PR.

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
```diff
@@ -41,10 +42,12 @@
...
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
-@SessionFactory(exportSchema = false)
-@DomainModel(annotatedClasses = {BasicCrudTests.Book.class, BasicCrudTests.BookWithEmbeddedField.class})
-class BasicCrudTests implements SessionFactoryScopeAware {
+@SessionFactory(exportSchema = false, useCollectingStatementInspector = true)
+@DomainModel(annotatedClasses = {BasicCRUDTests.Book.class, BasicCRUDTests.BookWithEmbeddedField.class})
+class BasicCRUDTests implements SessionFactoryScopeAware {
```
stIncMale, line 50: 1. You renamed this class to `BasicCrudIntegrationTests` in https://github.com/mongodb/mongo-hibernate/pull/57. What's the point of renaming it to `BasicCRUDTests` here?
2. We capitalize only the first letter, if needed, in an acronym in names of program elements: `UuidCodec` (the acronym is "UUID"), `AutomaticPojoCodec` (the acronym is "POJO"), JsonMode` (the acronym is "JSON").

So let's undo the change, and merge `main` into this branch once https://github.com/mongodb/mongo-hibernate/pull/57 is merged into `main`.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
```diff
@@ -169,6 +170,43 @@ void testSimpleDeletion() {
...
+
+                book.title = "Insurrection";
+                book.publishYear = 1899;
+                session.merge(book);
+            });
+
+            assertCollectionContainsOnly(
+                    BsonDocument.parse(
+                            """
+                            {"_id": 1, "author": "Leo Tolstoy", "publishYear": 1899, "title": "Insurrection"}\
```
katcharov, line 197:
```suggestion
                book.title = "Resurrection";
                book.publishYear = 1899;
                session.merge(book);
            });

            assertCollectionContainsOnly(
                    BsonDocument.parse(
                            """
                            {"_id": 1, "author": "Leo Tolstoy", "publishYear": 1899, "title": "Resurrection"}\
```

minor nit, just keeping our Tolstoy records accurate 🙂

NathanQingyangXu: Good catch!!!

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
stIncMale: I tried annotating `Book` with [`@DynamicUpdate`](https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/annotations/DynamicUpdate.html), and it just worked, barring the `statementInspector.getSqlQueries` assertion, of course. Should we add a test for that?

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
```diff
@@ -169,6 +170,43 @@ void testSimpleDeletion() {
...
+                session.merge(book);
+            });
+
+            assertCollectionContainsOnly(
+                    BsonDocument.parse(
+                            """
+                            {"_id": 1, "author": "Leo Tolstoy", "publishYear": 1899, "title": "Resurrection"}\
+                            """));
+            assertThat(statementInspector.getSqlQueries())
+                    .containsExactly(
```
stIncMale, line 200: Why do we use `SQLStatementInspector` here instead of checking the data in the DB? We don't use this approach for other tests in `BasicCrudIntegrationTests`, because we don't really care about the MQL generated, we only care about the data in the DB.

stIncMale (changes requested):

stIncMale:
The last reviewed commit is e988b33351b82cccddaa3a87d6e71c3eb58a2eb1.

katcharov:
I will take myself off as a formal reviewer for now - if a secondary is needed, please do add me once the primary is complete.

src/integrationTest/java/com/mongodb/hibernate/BasicCRUDTests.java
NathanQingyangXu: good point. I think we should cover this. 

There is also a `@DynamicInsert`, which won't include fields with null values, so it will align with the null-free insertion case we discussed. I might add it later but not in this PR.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
NathanQingyangXu: The reason I used it is to verify that the end result is based on `insertion->update` steps, not one single insertion combining both. If this is a invalid concern, I could remove its usage.

NathanQingyangXu: I think you are right. No need to test the implementation steps. Only the end result matters. The usage of `SQLStatementInspector` is deleted.

NathanQingyangXu:
> The last reviewed commit is [e988b33](https://github.com/mongodb/mongo-hibernate/commit/e988b33351b82cccddaa3a87d6e71c3eb58a2eb1).

I've finished resolving the comments, so the PR is ready for a new book.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -229,30 +233,54 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+    }
+
+    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Table Mutation: update
+
+    @Override
+    public void visitStandardTableUpdate(TableUpdateStandard tableUpdate) {
+        if (tableUpdate.getWhereFragment() != null) {
+            throw new FeatureNotSupportedException();
+        }
```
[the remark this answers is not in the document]
stIncMale: Why was the condition `tableUpdate.getReturningColumns() != null && !tableUpdate.getReturningColumns().isEmpty()` chosen over the proposed `tableUpdate.getNumberOfReturningColumns() > 0`? I am asking because the latter seem to be much more readable.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
```diff
@@ -193,7 +244,22 @@ static class Book {
         int publishYear;
     }
 
-    @Entity(name = "BookWithEmbeddedField")
+    @Entity
+    @Table(name = "books")
+    @DynamicUpdate
+    static class BookDynamicUpdated {
```
stIncMale, line 250: `BookDynamicUpdated` -> `BookDynamicallyUpdated`

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -229,30 +233,54 @@ public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
...
+            var astValue = acceptAndYield(valueBinding.getValueExpression(), FIELD_VALUE);
+            updates.add(new AstFieldUpdate(columnExpression, astValue));
+        }
+        astVisitorValueHolder.yield(
+                COLLECTION_MUTATION,
+                new AstUpdateCommand(tableUpdate.getMutatingTable().getTableName(), keyFilter, updates));
+    }
 
-        if (tableDelete.getNumberOfOptimisticLockBindings() > 0) {
+    private AstFilter getKeyFilter(RestrictedTableMutation<?> tableMutation) {
```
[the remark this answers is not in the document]
stIncMale: Let's then use `AbstractRestrictedTableMutation<? extends MutationOperation>` as the type of the `tableMutation` parameter of the `getKeyFilter` method? It is the most specific type suitable for all the current and foreseen usages.

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
stIncMale: > I used it is to verify that the end result is based on insertion->update steps, not one single insertion combining both

We know that it can't be just a single insertion combining both changes because we explicitly call `flush` before doing more changes.

stIncMale (changes requested):
The last reviewed commit is 90d78760c805c79a372b8f3e334cfdef9d3592ac.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
NathanQingyangXu: ok

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
NathanQingyangXu: overlooked. definitively `tableUpdate.getNumberOfReturningColumns() > 0` is better.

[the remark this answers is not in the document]
stIncMale: `AbstractMqlTranslator.visitStandardTableInsert` wasn't changed.

stIncMale (commented):
The last reviewed commit is 3c4e08ff07ce96f9a5a5f557cc62552f099ff933.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
NathanQingyangXu: changed

stIncMale (approved):
The last reviewed commit is 8803d302e73c484b59e158f88b85866b2b4c9e77.

PR #62, opened 2025-03-05
introduce a MongoExtension to simplify integration testing

NathanQingyangXu:
A simple extension providing the following two benefits:

- all the lifecycle events would be handled by the extension per se
- `MongoDatabase` could be injected following `hibernate-testing`'s `SessionFactoryAware` injection mechanism

Its example usage is the simplified `BasicCrudIntegrationTests` class. We could do more later (like command listener which bulk writing test requires), but for now this seems enough.


stIncMale (changes requested):

stIncMale:
The last reviewed commit is 973fda06e9875892175dbed5ed735d8c6c6f8c7b.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,60 @@
...
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.hibernate.internal.cfg.MongoConfigurationBuilder;
+import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public final class MongoExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback {
```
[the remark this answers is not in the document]
NathanQingyangXu: I made some changes as following:

- relies on `TestInstancePreConstructCallback` and `TestInstancePreDestroyCallback` to detect the timing of creating `MongoClient` and `MongoDatabase` using the former's callback parameter
- create `MongoClient` and `MongoDatabase` whenever the extension is used

I suggest not coupling with `MongoDatabaseAware` for it is only one of the ways to use this extension; it might be planned to inject Mongo stuff using other means in the future, e.g. by method parameter injection as `SessionFactoryScope` or even home-grown injection annotation on static and non-static fields.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,81 @@
...
+
+            var mongoConfig = new MongoConfigurationBuilder(hibernateProperties).build();
+            mongoClient = MongoClients.create(mongoConfig.mongoClientSettings());
+            mongoDatabase = mongoClient.getDatabase(mongoConfig.databaseName());
+        }
+    }
+
+    @Override
+    public void preDestroyTestInstance(ExtensionContext context) {
+        if (context == rootContext) {
```
stIncMale, line 60: Let's use equality, as formally there is no reason for us to explicitly rely on reference equality here.

```suggestion
        if (context.equals(rootContext)) {
```

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private ExtensionContext rootContext;
+
+    private MongoClient mongoClient;
+    private MongoDatabase mongoDatabase;
+
+    @Override
+    public void preConstructTestInstance(TestInstanceFactoryContext factoryContext, ExtensionContext context) {
+        if (factoryContext.getOuterInstance().isEmpty()) {
+            rootContext = context;
```
stIncMale, line 46: Let's use a name different from `rootContext` because of the `ExtensionContext.getRoot` method, which returns the actual root context that is unfortunately different from the outermost context we are interested in. Maybe `outermostContext`?

stIncMale (changes requested):

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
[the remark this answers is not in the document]
stIncMale: > relies on `TestInstancePreConstructCallback` and `TestInstancePreDestroyCallback`

👍 

> I suggest not coupling with `MongoDatabaseAware`

Agreed.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,81 @@
...
+            var mongoConfig = new MongoConfigurationBuilder(hibernateProperties).build();
+            mongoClient = MongoClients.create(mongoConfig.mongoClientSettings());
+            mongoDatabase = mongoClient.getDatabase(mongoConfig.databaseName());
+        }
+    }
+
+    @Override
+    public void preDestroyTestInstance(ExtensionContext context) {
+        if (context == rootContext) {
+            mongoDatabase = null;
```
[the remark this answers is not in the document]
NathanQingyangXu: My original thought is there might be edge case when `mongoDatabase.close()` is called (exception is thrown due to the reason `mongoClient` has been closed) for whatever reason (I did encountered such issue but I guess that is because other bugs). I deleted it then.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
NathanQingyangXu: currently I can't think of better name than `outermostContext`. Let us use it for now and we can change it when we know better.

NathanQingyangXu:
I added the second injection option by a home-grown `@InjectMongoCollection` annotation (whose `name` element denotes collection name), which could be applied to both static and non-static field. Its usage is examplified by `BasicCrudIntegraationTest`.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,101 @@
...
+
+        Predicate<Field> predicate = field -> MongoCollection.class.isAssignableFrom(field.getType());
+        findAnnotatedFields(context.getRequiredTestClass(), InjectMongoCollection.class, predicate)
+                .forEach(field -> {
+                    var annotation = field.getDeclaredAnnotation(InjectMongoCollection.class);
+                    String collectionName = annotation.name();
+                    var mongoCollection = mongoDatabase.getCollection(collectionName, BsonDocument.class);
+                    try {
+                        field.setAccessible(true);
+                        field.set(isStatic(field) ? null : context.getRequiredTestInstance(), mongoCollection);
```
NathanQingyangXu, line 88: the above `context.getRequiredTestInstance()` won't throw exception for the precondition that some field is of instance field has implied that test instance has been available.

src/integrationTest/java/com/mongodb/hibernate/junit/InjectMongoCollection.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target(FIELD)
+@Retention(RUNTIME)
+public @interface InjectMongoCollection {
+    String name();
```
stIncMale, line 28: Let's rename it to `value` to be able to use `@InjectMongoCollection("books")`, which seems better than `@InjectMongoCollection(name = "books")`.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext context) {
+        if (context.getTestInstance().orElse(null) instanceof MongoDatabaseAware mongoDatabaseAware) {
+            mongoDatabaseAware.injectMongoDatabase(mongoDatabase);
+        }
+
+        Predicate<Field> predicate = field -> MongoCollection.class.isAssignableFrom(field.getType());
+        findAnnotatedFields(context.getRequiredTestClass(), InjectMongoCollection.class, predicate)
```
stIncMale, line 81: The predicate you suggest hides mistakes. If we declare

```java
String collection;
```

Then with the current predicate the field stays `null`, but the test run does not fail. It is better for the test run to fail, which we can achieve by not filtering out felds:

```suggestion
        findAnnotatedFields(context.getRequiredTestClass(), InjectMongoCollection.class, field -> true)
```

src/integrationTest/java/com/mongodb/hibernate/junit/MongoDatabaseAware.java
stIncMale, line 1: Now that there is `@InjectMongoCollection`, we don't seem to need `MongoDatabaseAware`. Let's remove it to avoid it hanging forever unused. And if we ever actually need it, we can trivially add it back.

stIncMale (changes requested):

stIncMale:
The last reviewed commit is b66aa21d73702b1ddc8a11a47533fd366897c33c.

src/integrationTest/java/com/mongodb/hibernate/junit/InjectMongoCollection.java
NathanQingyangXu: My first iteration is to use `value` as you thought, but my second thought lead me to think explicit element name spec might be more accurate and less confusing. The `name` conveys that the value provided is collection name, not other things.

A perfect solution would be Spring's `@AliasFor`, but we don't have such luxurary.

So let us use `value` then. I over thought previously.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoDatabaseAware.java
NathanQingyangXu: Yeah, currently `@InjectMongoCollection` suffices.

stIncMale (approved):
The last reviewed commit is 8847b1c18f9688259f4203a43700277f2ea6945b.

PR #63, opened 2025-03-09
simplify JakartaPersistenceBootstrappingIntegrationTests

NathanQingyangXu:
using `hibernate-testing` is beneficial to simplify either Hibernate native or JPA testing. Another benefit is to remove the global `persistence.xml` resource which is uesd for container purpose.


src/integrationTest/java/com/mongodb/hibernate/JakartaPersistenceBootstrappingIntegrationTests.java
```diff
@@ -16,67 +16,48 @@
...
+import org.hibernate.testing.orm.junit.Setting;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 
+@Jpa(
+        exportSchema = false,
+        integrationSettings =
+                @Setting(
+                        name = AvailableSettings.CONNECTION_PROVIDER,
+                        value = "com.mongodb.hibernate.jdbc.MongoConnectionProvider"),
```
NathanQingyangXu, line 42: the other special config is mandatory; the other properties would come from `hibernate.properties` automatically

src/integrationTest/java/com/mongodb/hibernate/JakartaPersistenceBootstrappingIntegrationTests.java
```diff
@@ -16,67 +16,48 @@
...
-                            .getCollection("items")
-                            .countDocuments());
-        }
+    void smoke(EntityManagerFactoryScope scope) {
+        scope.inTransaction(em -> {
+            var item = new Item();
+            item.id = 1;
+            em.persist(item);
+        });
+        assertThat(newArrayList(collection.find())).containsOnly(BsonDocument.parse("{_id: 1}"));
```
NathanQingyangXu, line 57: `AssertJ` has util methods to convert `iterable` to `Collection`. I think its usage simplifies.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -70,18 +70,20 @@ public void preDestroyTestInstance(ExtensionContext context) {
...
+                        String collectionName = annotation.value();
+                        var mongoCollection = mongoDatabase.getCollection(collectionName, BsonDocument.class);
+                        try {
+                            field.setAccessible(true);
+                            field.set(isStatic(field) ? null : context.getRequiredTestInstance(), mongoCollection);
+                        } catch (Exception ex) {
+                            throw new RuntimeException(ex);
+                        }
+                    });
+        }
```
NathanQingyangXu, line 86: Just found if our `MongoExtension` is used without PER_CLASS lifecycle, it would be broken. Currently just add a simple sanitfy check to avoid NPE, in the future we might introduce an annotation (analogous to `@SessionFactory` and `@Jpa` in `hibernate-testing`) in lieu of direct extension usage, so we could not only bind `@TestInstance( TestInstance.Lifecycle.PER_CLASS )` together, but also enjoy annotation elements and configurability.

stIncMale: Just adding the `if (mongoDatabase != null)` does not make the current code correct when `Lifecycle.PER_CLASS` is used.

I propose https://github.com/stIncMale/mongo-hibernate/commit/10ebd61558d4455ac22b3c9c716846a26d41aff1, where I not only changed `MongoExtension`, but also made all other changes I am proposing, regardless of whether I made a comment about them or not).

Not only the proposed `MongoExtension` works regardless of the `Lifecycle` chosen, but it also uses the same `MongoClient` for all integration tests, which is something I wanted us to have from the get go, because this should save us noticeable time when running integration tests.

stIncMale (changes requested):

src/integrationTest/java/com/mongodb/hibernate/JakartaPersistenceBootstrappingIntegrationTests.java
stIncMale: See https://github.com/mongodb/mongo-hibernate/pull/63#discussion_r1988224528.

stIncMale (changes requested):
The last reviewed commit is cd262b8b7b42021b8d52846c47de8de4c7195515.

src/integrationTest/java/com/mongodb/hibernate/JakartaPersistenceBootstrappingIntegrationTests.java
NathanQingyangXu: good simplification! My only concern is maybe in the future different integration testing has different `MongoClient` settings, then more flexibility is desired. Perf issue is early opt currently to me. But for now I have no objection to your approach.

NathanQingyangXu: My only change to your code is to use `https://junit.org/junit5/docs/current/user-guide` in lieu of `https://junit.org/junit5/docs/snapshot/user-guide`. Snapshot version is less desirable for it is unstable.

NathanQingyangXu: I didn't touch other integration testings to avoid PR conflicts and labourious conflict resolving. We could simplify consistently in one single tech improvement PR in the future when all the PRs have been merged.

NathanQingyangXu: > good simplification! My only concern is maybe in the future different integration testing has different `MongoClient` settings, then more flexibility is desired. Perf issue is early opt currently to me. But for now I have no objection to your approach.

nvm. I think this is a generic default approach. If some testing has special requirement, it simply avoid using this extension and go about programmatic configuration in ad-hoc way.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
stIncMale: Do you have objections to the change to `InjectMongoCollection.java`, or was it not applied accidentally?

src/integrationTest/java/com/mongodb/hibernate/JakartaPersistenceBootstrappingIntegrationTests.java
stIncMale: > We could simplify consistently in one single tech improvement PR in the future when all the PRs have been merged.

OK, I'll create it later.

stIncMale (commented):
The last reviewed commit is 0ce83ec34f3a7f27728625a58ebdaec70aa16f82.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
NathanQingyangXu: I overlooked. Added.

stIncMale (approved):

PR #65, opened 2025-03-12
add @DynamicInsert testing case in BasicCrudIntegrationTests

NathanQingyangXu:
a tech debt when we have tested `@DynamicUpdate` in the basic update translation PR; also created a new `crud` package to avoid dumping everything into root package. There could be many more integrtation testing classes under `crud` feature in the future.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/crud/BasicCrudIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/crud/BasicCrudIntegrationTests.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.mongodb.hibernate;
+package com.mongodb.hibernate.crud;
 
 import static com.mongodb.hibernate.internal.MongoConstants.ID_FIELD_NAME;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -31,6 +31,7 @@ import jakarta.persistence.Table;
 import java.util.ArrayList;
 import java.util.List;
 import org.bson.BsonDocument;
+import org.hibernate.annotations.DynamicInsert;
 import org.hibernate.annotations.DynamicUpdate;
 import org.hibernate.testing.orm.junit.DomainModel;
 import org.hibernate.testing.orm.junit.SessionFactory;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
```diff
@@ -45,13 +46,14 @@
...
             BasicCrudIntegrationTests.BookWithEmbeddedField.class,
+            BasicCrudIntegrationTests.BookDynamicallyInserted.class,
             BasicCrudIntegrationTests.BookDynamicallyUpdated.class
         })
 @ExtendWith(MongoExtension.class)
 class BasicCrudIntegrationTests implements SessionFactoryScopeAware {
 
     @InjectMongoCollection("books")
-    private static MongoCollection<BsonDocument> collection;
+    private static MongoCollection<BsonDocument> mongoCollection;
```
NathanQingyangXu, line 56: to maintain consistency with other integration testing classes (both `MongoStatementIntegrationTests` and `MongoPreparedStatementIntegrationTests` use `mongoCollection`)

src/integrationTest/java/com/mongodb/hibernate/crud/BasicCrudIntegrationTests.java
```diff
@@ -83,6 +85,25 @@ void testSimpleEntityInsertion() {
...
+                session.persist(book);
+            });
+            var expectedDocument = BsonDocument.parse(
+                    """
+                    {
+                        _id: 1,
+                        title: "War and Peace",
+                        publishYear: 1867
+                    }""");
+            assertCollectionContainsExactly(expectedDocument);
```
NathanQingyangXu, line 104: `author` won't be inserted for it is null, thus ending up with a more compact storage mechanism.
The troublesome thing is the annotations of `@DynamicInsert` and `@DynamicUpdate` are both deprecated.

stIncMale: They are not deprecated. However, it is not clear whether we should support `@DynamicInsert` at the moment or not, so let's not introduce such a test for now.

P.S. I'll reflect the renaming done here in https://github.com/mongodb/mongo-hibernate/pull/64, so that's not lost.

PR #67, opened 2025-03-14
some improvement to gradle usage

NathanQingyangXu:
got some inspiration from reviewing Ross's gradle modernization PR. 
Changes include:

- moved tripple quote indentation spotless into a customized gradle plugin "spotless-java-extention", so our main build.gradle.kts won't lose focus on big picture
- changed `BuildConfig` java's package from `com.mongo.hibernate` to `com.mongo.hibernate.internal`
- enriched the kts spotless formatting configuration
- exploited `bundles` toml feature to decrease common test dependencies duplication
- misc minor improvements


NathanQingyangXu:
closed for now for it is not urgent and could include more improvements later.

build.gradle.kts
```diff
@@ -163,28 +119,26 @@ tasks.withType<JavaCompile>().configureEach {
...
 dependencies {
-    listOf(libs.junit.jupiter, libs.assertj, libs.logback.classic).forEach {
-        testImplementation(it)
-        integrationTestImplementation(it)
-    }
-
+    testImplementation(libs.bundles.test.common)
     testImplementation(libs.mockito.junit.jupiter)
-    testRuntimeOnly(libs.junit.platform.launcher)
+    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
```
rozza, line 134: I'd be interested in the reasons to remove this dependency from libs.versions.toml.

buildSrc/src/main/kotlin/spotless-java-extension.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    java {
+        addStep(
+            FormatterStep.create(
+                "multilineFormatter",
+                MultilineFormatter(),
+                { formatter -> FormatterFunc { input -> formatter.format(input) } }))
+    }
+}
+
+class MultilineFormatter : Serializable {
```
rozza, line 35: I'm interested in the custom multiline formatting - what made you go with it?

rozza (approved):
Just a couple of questions - but nothing that blocks a LGTM.

build.gradle.kts
NathanQingyangXu: Since v9, Gradle would require explicit spec of the platform otherwise a forward compatibility warning will be issued (you can check your branch to see).

So in the current user guide, the following pattern always shows up (e.g. https://docs.gradle.org/current/userguide/java_testing.html#compiling_and_executing_junit_jupiter_tests):
```
dependencies {
    testImplementation("org.junit.jupiter:junit-jupiter:5.7.1")
    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}
```
Given it is not using version, it seems inappropriate to config the launcher in toml file. That is why I excluded it.

buildSrc/src/main/kotlin/spotless-java-extension.gradle.kts
NathanQingyangXu: It is because we require the usage of multiline a lot in our project, especially for the following integration test pattern:
```
var expectedDocument = BsonDocument.parse(
            """
            {
                _id: 1,
                title: "War and Peace",
                author: "Leo Tolstoy",
                publishYear: 1867
            }""");
assertCollectionContainsExactly(expectedDocument);
```
previously the ugly alignment issue troubles coder and code reviewers a lot. Finally we create a Spotless step to make it hassle-free once for all.

I noticed the multiline alignment issue seems present in your branch as well (not in Java but in Kotlin), so this spotless customization might be useful to you as well, though not as necessary as in our usage.

rozza: I like that - thanks for sharing.

PR #68, opened 2025-03-17
introduce jmh gradle usage for local benchmarking purpose

NathanQingyangXu:
We have encountered the need to objectively compare implementation alternatives in terms of perf by local benchmarking. JMH is such handy tool. An example benchmark is included.

The current implementation in main is based on stream usage, and its local (on my laptop) is as follows:

```
Benchmark                                           Mode  Cnt        Score        Error  Units
MongoStatementBenchmark.getProjectStageFieldNames  thrpt   25  7436777.432 ± 237277.758  ops/s
```

When I changed it to non-stream alternative, JMH helps provided that the throughput almost doubles:

```
Benchmark                                           Mode  Cnt         Score        Error  Units
MongoStatementBenchmark.getProjectStageFieldNames  thrpt   25  14637872.456 ± 879754.679  ops/s
```

However, changing the `MongoStatementBenchmark.getProjectStageFieldNames()` is out of the scope of this PR for it has other more issues which are more important than the speed doubling. A followup PR might be created to solve the business logic issue later.

There might be issue on IDE integration but that is out of the scope of this PR as well.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ plugins {
     alias(libs.plugins.spotless)
     alias(libs.plugins.errorprone)
     alias(libs.plugins.buildconfig)
+    alias(libs.plugins.jmh)
 }
 
 repositories { mavenCentral() }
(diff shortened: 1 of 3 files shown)


stIncMale (commented):
I don't think we encountered the need to test performance. Nor do I think that `MongoStatement.getFieldNamesFromProjectStage` would be worth benchmarking even if we were to work on performance, because the performance of this method almost certainly has no measurable effect on the performance of the read/write action involving execution of this method.

Let's not spend time on things we definitely not need at the moment.

PR #72, opened 2025-03-17
Fix `MongoStatement.getFieldNamesFromProjectStage`

stIncMale:


stIncMale:
Both issues addressed in this PR were discovered by @NathanQingyangXu.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -99,9 +99,21 @@ static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
...
+                    specification.getKey(), specification.getValue()));
+        }
+        var include = (value.isBoolean() && value.asBoolean().getValue())
                 || (value.isNumber() && value.asNumber().intValue() == 1);
+        if (!include && !specification.getKey().equals(ID_FIELD_NAME)) {
+            throw new RuntimeException(format(
+                    "Exclusions are not allowed in `$project` specifications, except for the [%s] field: [%s, %s]",
+                    ID_FIELD_NAME, specification.getKey(), specification.getValue()));
+        }
+        return include;
```
NathanQingyangXu, line 116: It seems we might well only focus on `ZeroOrfalse` as follows:
```
static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
        var fieldNames = projectStage.entrySet().stream()
                .filter(field -> ! falseOrZero(field.getValue()))
                .map(Map.Entry::getKey)
                .collect(toCollection(ArrayList::new));
        if (!projectStage.containsKey(ID_FIELD_NAME)) {
            // MongoDB includes this field unless it is explicitly excluded
            fieldNames.add(ID_FIELD_NAME);
        }
        return fieldNames;
    }

    private static boolean falseOrZero(BsonValue value) {
        return (value.isBoolean() && ! value.asBoolean().getValue())
                || (value.isNumber() && value.asNumber().intValue() == 0);
    }
```

src/test/java/com/mongodb/hibernate/jdbc/MongoStatementTests.java
```diff
@@ -183,15 +183,23 @@ void testExecuteBatch() throws SQLException {
...
-                () -> asserter.accept("{title: 1, _id: 0}", List.of("title")),
-                () -> asserter.accept("{title: 1, _id: false}", List.of("title")),
-                () -> asserter.accept("{_id: 1, title: 1}", List.of("_id", "title")));
+                () -> successAsserter.accept("{title: 1, publishYear: 1}", List.of("title", "publishYear", "_id")),
+                () -> successAsserter.accept("{title: 1, _id: 0}", List.of("title")),
+                () -> successAsserter.accept("{title: 1, _id: false}", List.of("title")),
+                () -> successAsserter.accept("{_id: 1, title: 1}", List.of("_id", "title")),
+                () -> failureAsserter.accept("{title: 1, publishYear: 0}", "Exclusions are not allowed"),
+                () -> failureAsserter.accept("{title: 1, publishYear: false}", "Exclusions are not allowed"),
+                () -> failureAsserter.accept("{title: '$fieldName'}", "Expressions are not allowed"),
```
NathanQingyangXu, line 201: should we use double quotes for the string field value?

stIncMale: No, we don't need to use double quotes. Given that you are asking, do you see reasons to use them? If yes, what are those?

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -99,9 +99,21 @@ static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
...
 
-    private static boolean trueOrOne(BsonValue value) {
-        return (value.isBoolean() && value.asBoolean().getValue())
+    private static boolean isIncludeProjectSpecification(Map.Entry<String, BsonValue> specification) {
+        var value = specification.getValue();
+        if (value.isString() || value.isDocument()) {
+            throw new RuntimeException(format(
+                    "Expressions are not allowed in `$project` specifications: [%s: %s]",
+                    specification.getKey(), specification.getValue()));
+        }
```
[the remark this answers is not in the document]
stIncMale: 1. I am not sure if expressions (string or document ones) are going to work in application-specified native queries, as I am not sure Hibernate ORM will always know the data type for them. Maybe we will have to write some more code to support them in all cases. So for now, I suggest us to forbid them until we implement support for application-supplied native queries and add tests showing that they work.
2. If we need to allow expressions for some other features we plan to implement, let's allow the expressions we need when we need them, and, again, have corresponding tests.

Instead of throwing `RuntimeException`, we may throw `FeatureNotSupportedException` if you are sure we'll need to support expressions in `$project` specifications. Will we need only string expressions, or also document expressions? (it's not like those are the actual terms, but I don't think there are any special terms for those, so I came up with ones to distinguish them here)

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
stIncMale: I explained the checks for expressions in the `$project` specifications in https://github.com/mongodb/mongo-hibernate/pull/72#discussion_r2000101085.

> It seems we might well only focus on falseOrZero as follows:

I don't think so. You mention that we do not handle exclusion specifications correctly, which is true: if this `$project` stage with exclusion specifications comes from an application via a native query (which, as far as I understand, is the only way it may happen), then we won't be able to correctly construct a list of the field names for the result set, because doing so based on exclusion specifications requires us to know the schema, which we don't know. And that is why we should explicitly forbid exclusion specifications (except for the one for `_id`).

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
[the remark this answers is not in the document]
stIncMale: Btw, you can see that I am not checking for other possibilities, like a `$project` specification containing a timestamp, or binary. And that is because those are not allowed in a `$project` specification, and the server will to the validation for us. This means I know that expressions are allowed in `$project` specifications, and informing me on that is a waste of time, especially given that you had already told me that in Slack.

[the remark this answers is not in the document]
jyemin: We'll eventually have to support any aggregate expression for $project as part of milestone 4.  For example, in an HQL query like `select x = y from Foo`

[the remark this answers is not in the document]
NathanQingyangXu: I agree. It could be any expression (even including array), so to me we simply treat 0 or false as special case. When we invoke this method, it means the cursor has been returned together with the stage so we only need to check the exclusion case.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
NathanQingyangXu: I raised the issue mainly due to the other expression reason. We should allow it for we used it in POC project. The other issue you mentioned above is relatively less important and we have no disagreement on it.

src/test/java/com/mongodb/hibernate/jdbc/MongoStatementTests.java
NathanQingyangXu: It is fine to use single or double quotes. I asked for previously I always used double quotes, now I know single quote is fine after local experimenting.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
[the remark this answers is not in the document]
stIncMale: Based on the knowledge from the PoC, will we need changes in our codebase to support `$project` specifications with expressions (beyond those required solely to support native queries)?

src/test/java/com/mongodb/hibernate/jdbc/MongoStatementTests.java
stIncMale: The reason single quotes are supported for values of the string type, and no quotes are required for keys, is that our parsers are lenient, and can parse invalid JSON or extended JSON.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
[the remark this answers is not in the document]
stIncMale: I replaced `RuntimeException` with `FeatureNotSupportedException` for expressions and literals.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
stIncMale: Turned out, for all types but BSON boolean and number (whether integral or not), literals are allowed in `$project` specification even without the `$literal` expression. Furthermore, any number but 0 is interpreted as "include". Because of this, checking whether the specification excludes a field is simpler, than checking whether it is included. So I changed this part.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -99,9 +99,21 @@ static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
...
-                || (value.isNumber() && value.asNumber().intValue() == 1);
+    private static boolean isExcludeProjectSpecification(Map.Entry<String, BsonValue> specification) {
+        var key = specification.getKey();
+        var value = specification.getValue();
+        var exclude = (value.isBoolean() && !value.asBoolean().getValue())
+                || (value.isNumber() && value.asNumber().intValue() == 0);
+        if (exclude && !key.equals(ID_FIELD_NAME)) {
+            throw new RuntimeException(format(
+                    "Exclusions are not allowed in `$project` specifications, except for the [%s] field: [%s, %s]",
+                    ID_FIELD_NAME, key, value));
```
stIncMale, line 110: This is hardly to ever change, unless we start requiring a schema or start inferring it.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -99,9 +99,21 @@ static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
...
+        }
+        if (!value.isBoolean() && !value.isInt32()) {
+            throw new FeatureNotSupportedException(format(
+                    "Expressions and literals are not supported in `$project` specifications: [%s: %s]", key, value));
```
stIncMale, line 114: This is likely to change. When it does, we should try to allow only the bare minimum we need or know works (for example, because we have tests), and keep forbidding the rest.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
NathanQingyangXu: well, it is native query related. Currently let us keep it in mind and then move on.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
```diff
@@ -99,9 +99,21 @@ static List<String> getFieldNamesFromProjectStage(BsonDocument projectStage) {
...
+        var key = specification.getKey();
+        var value = specification.getValue();
+        var exclude = (value.isBoolean() && !value.asBoolean().getValue())
+                || (value.isNumber() && value.asNumber().intValue() == 0);
+        if (exclude && !key.equals(ID_FIELD_NAME)) {
+            throw new RuntimeException(format(
+                    "Exclusions are not allowed in `$project` specifications, except for the [%s] field: [%s, %s]",
+                    ID_FIELD_NAME, key, value));
+        }
+        if (!value.isBoolean() && !value.isInt32()) {
```
NathanQingyangXu, line 112: should we use `!value.isNumber()` to align with the above logic?

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
stIncMale: Could you please elaborate on why this is related to native queries?

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
stIncMale: Absolutely. Sorry for this nonsense. Done in https://github.com/mongodb/mongo-hibernate/pull/72/commits/075bef0bdd5b5a0aa3f23f6d3b4e284a2d5e738e.

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
NathanQingyangXu: because for non-native query we have 100% control of the MQL generation, including the `$project` stage. If we saw no need to use `exclusion list` (from my understanding, explicit inclusion list suffices and should be relies upon) during MQL translation, this issue won't bother us at all.

NathanQingyangXu (approved):

src/main/java/com/mongodb/hibernate/jdbc/MongoStatement.java
stIncMale: That still does not explain your _"well, it is native query related"_ comment. Do you think native queries will have to use exclustions? If they do and we allow that, how can we possibly know the list of the columns for `ResultSet` without inferring that list from the returned documents, which we don't have plans to do?

jyemin: Sorry to jump in here.  The way I understand it, native queries will not _have_ to use exclusions.  The question is whether we _allow_ them to.  Since I also can't see a way to make it work, I don't think we will allow them to.

stIncMale: Right. And if we are not going to allow them to use exclusions, then the code this thread is about will likely not change, which is what I said in the first clarifying comment for reviewers. And that makes me wonder, what is it that we need to keep in mind in relation to this code and native queries (I am trying to understand the comment https://github.com/mongodb/mongo-hibernate/pull/72#discussion_r2003415926, assuming it expressed or implied something that I am missing).

NathanQingyangXu: > That still does not explain your _"well, it is native query related"_ comment. Do you think native queries will have to use exclustions? If they do and we allow that, how can we possibly know the list of the columns for `ResultSet` without inferring that list from the returned documents, which we don't have plans to do?

I didn't imply that. Sorry for misunderstanding. I meant the exclusion list could only happen in theory in native query. Native query has many edge cases and this is another one.

PR #76, opened 2025-03-22
Add `MongoConfigurationContributor` tests and document it better

stIncMale:
Given that `MongoConfigurationContributor` is used by the code in `StandardServiceRegistryScoped`, `StandardServiceRegistryScopedStateTests` seems to be the best place for the new tests


src/test/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedStateTests.java
```diff
@@ -36,4 +40,40 @@ void differentStandardServiceRegistriesHaveDifferentStateInstances() {
...
+                    .addService(MongoConfigurationContributor.class, mongoConfigurators::add);
+            try (var standardServiceRegistry1 = standardServiceRegistryBuilder.build();
+                    var standardServiceRegistry2 = standardServiceRegistryBuilder.build()) {
+                standardServiceRegistry1.requireService(StandardServiceRegistryScopedState.class);
+                standardServiceRegistry2.requireService(StandardServiceRegistryScopedState.class);
+                for (var i = 0; i < mongoConfigurators.size(); i++) {
+                    for (var j = 0; j < mongoConfigurators.size(); j++) {
+                        if (i != j) {
+                            assertNotSame(mongoConfigurators.get(i), mongoConfigurators.get(j));
+                        }
```
NathanQingyangXu, line 58: `org.assertj.core.api.assertThat(mongoConfigurators).assertDoesNotHaveDuplicates()` might be more readable.

NathanQingyangXu: further digging shows that the above suggestion is based on `equalness` contract, not on object reference. Didn't find good way in assertj.

I understand the reason you didn't go about as belows:
```
assertThat(mongoConfigurators).hasSize(2);
assertNotSame(mongoConfigurators.get(0), mongoConfigurators.get(1));
```
to decouple logic from the other unit testing case.

But given that `assertNotSame()` is order agnostic, we might well do as the normal pair iterating boilerplate as below, to avoid unnecessary assertions (and a little bit less confusing from code reading perspective):

```
for (var i = 0; i < mongoConfigurators.size() - 1; i++) {
      for (var j = i; j < mongoConfigurators.size(); j++) {
          assertNotSame(mongoConfigurators.get(i), mongoConfigurators.get(j));
      }
}
```

NathanQingyangXu: the above comments are optional and minor. So I approved your PR.

NathanQingyangXu (approved):

src/test/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedStateTests.java
NathanQingyangXu: I furhter explored a little bit. Seems the following clumsy assertj works by comparing based on object reference:
```
assertThat(mongoConfigurators).usingElementComparator((v1, v2) -> v1 == v2 ? 0 : -1).doesNotHaveDuplicates();
```

but my preference would be to combine the two unit testing cases into one without the unnecessary complexity. but I am ok with your approach.

stIncMale: > my preference would be to combine the two unit testing cases into one

I combined them in 77b8a5f8c85a78055322be9898eba442bb94da77.

NathanQingyangXu: I approved, though not fully convinced that we need to consider multiple `StandardServiceRegistry` scenario.

NathanQingyangXu (approved):

src/test/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedStateTests.java
stIncMale: Can we prevent this scenario? I think, we can't. But also, we should not, because it seems totally fine for an application to build and use multiple `SessionFactory`s, which likely results in them building multiple `StandardServiceRegistry`s. It this incorrect?

NathanQingyangXu: That is true. But for one single `SessionFactory`, there seems no reason to create multiple `StandardServiceRegistry`, right? 
The following test code seems confusing to me. Obviously the same `StandardServiceRegistryBuilder` is used twice within one single `SessionFactory`. I doubt this is a typical usage scenario at all.
```
try (var standardServiceRegistry1 = standardServiceRegistryBuilder.build();
                    var standardServiceRegistry2 = standardServiceRegistryBuilder.build()) {
```

Usually Hibernate focuses on single `SessionFactory` usage (there are almost zero testing cases related to multiple `SessionFactory` in Hibernate ORM's codebase). For different `SessionFactory`s, it seems obvious that they won't share services, unless JNI based container-managed resource? But that is exception.

 In one single `SessionFactory`, a single `StandardServiceRegistry` and its builder suffices, to my best knowledge.

NathanQingyangXu: Usually different SessionFactory has its own separate scope and it is very rare for them to share something (container-managed JNDI might be an exception but it is low-level and its wrapper should be different). Usually they have separate `BootstrapServiceRegistry` and `StandardServiceRegistry` pair. 

If we want to simulate the scenario of multiple SessionFactories, we might need to provide separate pairs of the two registries. However, it seems obvious we will end up with different stuff and the testing seems unnecessary.

stIncMale: > for one single SessionFactory, there seems no reason to create multiple `StandardServiceRegistry`, right?

Yes. I have never said or suggested otherwise.

> The following test code seems confusing to me. Obviously the same `StandardServiceRegistryBuilder` is used twice within one single `SessionFactory`.

The test code at the moment when you commented did not create a `SessionFactory` at all, nor did it restrict the number of `SessionFactory`s one could build, so the claim is not only not obvious, but is also false.

> I doubt this is a typical usage scenario at all.
> ...
> Usually Hibernate focuses on single `SessionFactory` usage

I don't think it matters whether it's a typical scenario or not. The important part is that it's a technically possible scenario:
- it is possible for an application to build and use more than one `SessionFactory`;
- it is possible for multiple `SessionFactory`s to share the same instance of `StandardServiceRegistry`, which means, we cannot say that `MongoConfigurationContributor.configure` is called once per `SessionFactory`.

> If we change the testing code to create multiple SessionFactories, it would make more sense and less confusing.

I did that, no problem. Though now the test simply has the previous code as is, plus more code.

> Usually they (`SessionFactory`s) have separate `BootstrapServiceRegistry` and `StandardServiceRegistry` pair.

Yes. But I fail to see why or how that is relevant. If you think that as a result the proposed documentation change is incorrect, or can be improved, please share how.

NathanQingyangXu: Saw you have changed your testing code and your doc is much better.

NathanQingyangXu (approved):

PR #78, opened 2025-03-26
Use "?" for parameter markers in both native query and query translation

jyemin:
WIP

HIBERNATE-61

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
@@ -237,6 +237,37 @@ class BasicCrudIntegrationTests implements SessionFactoryScopeAware {
         }
     }
 
+    @Nested
+    class NativeQueryTests {
+
+        @Test
+        void testNative() {
+            var book = new Book();
+            book.id = 1;
+            book.title = "In Search of Lost Time";
+            book.publishYear = 1913;
+
+            sessionFactoryScope.inTransaction(session -> session.persist(book));
+
+            var nativeQuery =
+                    """
+                    {
+                        aggregate: "books",
+                        pipeline: [
+                            { $match :  { _id: { $eq: :id } } },
+                            { $project: { _id: 1, publishYear: 1, title: 1, author: 1 } }
+                        ]
+                    }
+                    """;
+            sessionFactoryScope.inTransaction(session -> {
+                var query = session.createNativeQuery(nativeQuery, Book.class)
+                        .setParameter("id", book.id);
+                var queriedBook = query.getSingleResult();
+                assertThat(queriedBook).usingRecursiveComparison().isEqualTo(book);
+            });
+        }
+    }
+
     private static void assertCollectionContainsExactly(BsonDocument expectedDoc) {
         assertThat(mongoCollection.find()).containsExactly(expectedDoc);
     }
--- a/src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
+++ b/src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2025-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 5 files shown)


src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.jdbc;
+
+class MongoParameterRecognizer {
```
jyemin, line 19: Adopted from JsonScanner in Java driver.

src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return i - 1;
+    }
+
+    private static int scanString(final char quoteCharacter, final int startIndex, final String json, final StringBuilder builder) {
+        int i = startIndex;
+        builder.append(quoteCharacter);
+        while (i < json.length()) {
+            char c = json.charAt(i++);
+            if (c == '\\') {
```
jyemin, line 84: I didn't include explicit support for escapes like `\uFEBA`.  It will still work as is so long as the string is not malformed, but could add back what is in JsonScanner.

src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
```diff
@@ -0,0 +1,99 @@
...
+                case ',':
+                case ' ':
+                    builder.append(c);
+                    break;
+                case '\'':
+                case '"':
+                    i = scanString(c, i, json, builder);
+                    break;
+                case '?':
+                    builder.append("{$undefined: true}");
```
NathanQingyangXu, line 42: do we need to take into consideration the possibility that `?` shows up in a quoted string? E.g.
```
{
    aggregate: "books",
    pipeline: [
        { $match :  { title: { $eq: "Who's Afraid of Virginia Woolf?" } } },
        { $project: { _id: 1, publishYear: 1, title: 1, author: 1 } }
    ]
}
```

src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
```diff
@@ -0,0 +1,99 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.jdbc;
+
+class MongoParameterRecognizer {
+
+    static String replace(String json) {
+        StringBuilder builder = new StringBuilder(json.length());
```
NathanQingyangXu, line 22: unless `json` doesn't contain `?`, this `StringBuilder` will grow for the replacement length is longer than `?`. How to assign the optimal size is unclear here

src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
jyemin: This is already account for via the `scanString` method.

Needless to say, we should add comprehensive unit testing for this.

src/main/java/com/mongodb/hibernate/jdbc/MongoParameterRecognizer.java
jyemin: Yeah, not sure it matters too much though

src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java
```diff
@@ -237,6 +237,37 @@ void testGetByPrimaryKeyWithNullValueField() {
...
+                    """;
+            sessionFactoryScope.inTransaction(session -> {
+                var query = session.createNativeQuery(nativeQuery, Book.class)
+                        .setParameter("id", book.id);
+                var queriedBook = query.getSingleResult();
+                assertThat(queriedBook).usingRecursiveComparison().isEqualTo(book);
+            });
+        }
+    }
+
```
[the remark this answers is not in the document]
jyemin: I was just doing something easy.  Whoever picks this up should feel free to move the tests to wherever makes sense.

PR #79, opened 2025-03-26
Fail if the product is used with an unsupported MongoDB version

stIncMale:


src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
```diff
@@ -18,16 +18,15 @@
...
+import static com.mongodb.hibernate.internal.MongoConstants.MONGO_JDBC_DRIVER_NAME;
 
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 
 final class MongoDatabaseMetaData implements DatabaseMetaDataAdapter {
 
-    public static final String MONGO_DATABASE_PRODUCT_NAME = "MongoDB";
-    public static final String MONGO_JDBC_DRIVER_NAME = "MongoDB Java Driver JDBC Adapter";
```
stIncMale, line 29: I needed to use `MONGO_DATABASE_PRODUCT_NAME` in `MongoDialect`, but did not want to make `MongoDatabaseMetaData` public, so I moved the constants to `MongoConstants`, which seems appropriate.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -39,11 +42,40 @@ public MongoDialect(DialectResolutionInfo info) {
         super(info);
     }
 
+    /**
+     * This constructor is called only if Hibernate ORM falls back to it due to a failure of
+     * {@link MongoDialect#MongoDialect(DialectResolutionInfo)}.
+     *
+     * @deprecated Exists only to avoid the confusing {@link NoSuchMethodException} thrown by Hibernate ORM when
```
NathanQingyangXu, line 49: I am happy for this change for yeah, it confuses everybody when mongodb is not up (like first-time code reviewers lik Maxim).

src/main/java/com/mongodb/hibernate/internal/MongoConstants.java
```diff
@@ -20,5 +20,7 @@ public final class MongoConstants {
 
     private MongoConstants() {}
 
+    public static final String MONGO_DBMS_NAME = "MongoDB";
+    public static final String MONGO_JDBC_DRIVER_NAME = "MongoDB Java Driver JDBC Adapter";
```
NathanQingyangXu, line 24: yeah, we should move them here. Good move!

src/test/java/com/mongodb/hibernate/dialect/MongoDialectTests.java
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    @SuppressWarnings("deprecation")
+    void constructorFailsIfVersionIsNotSupported(@Mock DialectResolutionInfo info) {
+        when(info.makeCopyOrDefault(any())).thenCallRealMethod();
+        when(info.makeCopy()).thenCallRealMethod();
+        when(info.getDatabaseMajorVersion()).thenReturn(5);
+        when(info.getDatabaseMinorVersion()).thenReturn(3);
+        assertThat(assertThrows(RuntimeException.class, () -> new MongoDialect(info))
+                        .getMessage())
+                .isEqualTo("The minimum supported version of MongoDB is 6.0, but you are using 5.3");
```
NathanQingyangXu, line 41: There is more idiomatic assertj usage when it comes to exception assertion, e.g.
```
assertThatThrownBy(() -> new MongoDialect(info)).isInstanceOf(RuntimeException.class)
  .hasMessage("The minimum supported version of MongoDB is 6.0, but you are using 5.3");
```
see https://www.baeldung.com/assertj-exception-assertion

NathanQingyangXu (approved):

src/test/java/com/mongodb/hibernate/dialect/MongoDialectTests.java
stIncMale: Thank you! This is much better.

Done in 41402bf09e7bb458f5c0ee716751b7aee0c0f523.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -39,11 +42,40 @@ public MongoDialect(DialectResolutionInfo info) {
...
 
+    @Override
+    protected void checkVersion() {
+        var version = getVersion();
+        if (version == null) {
+            return;
+        }
+        var minimumVersion = getMinimumSupportedVersion();
+        if (version.isBefore(minimumVersion)) {
+            throw new RuntimeException(format(
```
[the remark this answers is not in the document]
stIncMale: > Not sure whether we can throw better runtime exception than `RuntimeException`.

Creating/throwing exception types more specific than `RuntimeException`/`Exception` makes sense only if they are expected to be handled automatically in a way specific to that type. That is not the case here.

> add an integration testing case to prove that this RuntimeException would work as expected

Done in d5a6d9d7df6a6bd0fba7a4a23f6955bb4ac5329c.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -39,11 +42,40 @@ public MongoDialect(DialectResolutionInfo info) {
...
+     *
+     * @deprecated Exists only to avoid the confusing {@link NoSuchMethodException} thrown by Hibernate ORM when
+     *     {@link MongoDialect#MongoDialect(DialectResolutionInfo)} fails.
+     * @throws RuntimeException Always.
+     */
+    @Deprecated()
+    public MongoDialect() {
+        throw new RuntimeException(format(
+                "Could not instantiate [%s], see the earlier exceptions to find out why",
+                MongoDialect.class.getName()));
```
NathanQingyangXu, line 57: So from the integration testing code, I got the following stack trace output:
```
08:51:34 [Test worker] WARN  o.h.t.cache.CachingRegionFactory - org.hibernate.testing.cache.CachingRegionFactory should be only used for testing.
08:51:34 [Test worker] WARN  o.h.e.j.e.i.JdbcEnvironmentInitiator - HHH000339: Could not obtain connection metadata: java.sql.SQLException: Failed to get metadata
08:51:34 [Test worker] WARN  o.h.e.j.e.i.JdbcEnvironmentInitiator - HHH000342: Could not obtain connection to query metadata
org.hibernate.boot.registry.selector.spi.StrategySelectionException: Could not instantiate named strategy class [com.mongodb.hibernate.dialect.MongoDialect]
... ...
Caused by: java.lang.RuntimeException: Could not instantiate [com.mongodb.hibernate.dialect.MongoDialect], see the earlier exceptions to find out why
```
So it seems the output is not very helpful as well. The real root cause ends up with WARN level logging only not as exception.

From my understanding, the clean solution might be:

- throw exception which would fail bootstrapping when metadata fetching failed, as opposed to only a WARN level logging; this behaviour aligns with our aggressive defensive development paradigm as well

but I am fine with the current approach, just wondering how to make it less confusing; currently it is better than previous status but still has room of improvement.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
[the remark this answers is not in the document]
NathanQingyangXu: My concern is how to verify the `RuntimeException` could really end up with Hibernate bootstrapping failure when version mismatches. So my original suggestion is not on `buildInfo` or metadta fetching failure, but when an unsupported version is returned. Could we work on that instead or additionally?

I think one reliable way is to change evergreen to create some integration test variant targetting unsupported mongodb and verify that the testing does fail. But I am not sure it is worth while.

Currently the only reliable way is to launch unsupported mongo service locally.

NathanQingyangXu (commented):
I am ok with the current status; just left some comments as food of thoughts

NathanQingyangXu (approved):

src/integrationTest/java/com/mongodb/hibernate/FailedBootstrappingIntegrationTests.java
```diff
@@ -0,0 +1,83 @@
...
+        @Override
+        public void close() {
+            mongoClient
+                    .getDatabase("admin")
+                    .runCommand(new BsonDocument()
+                            .append("configureFailPoint", failPointDocument.getString("configureFailPoint"))
+                            .append("mode", new BsonString("off")));
+        }
+    }
+}
```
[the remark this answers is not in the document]
stIncMale: > create a boot package

Done in eb17ca2ba1d88a21a34597998050e2fe32775917.

> say something in README.md

Done in 907a0c86f4ce5bfd4ae470a8b4677a6424154daa.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
stIncMale: We cannot change that, at least not without severe efforts, because that unhelpful behavior is dictated by Hibernate ORM, specifically by the following two pieces of code:

- https://github.com/hibernate/hibernate-orm/blob/00e552eb99a85c537c27938456f93fc039b89746/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java#L387-L392
- https://github.com/hibernate/hibernate-orm/blob/00e552eb99a85c537c27938456f93fc039b89746/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/internal/JdbcEnvironmentInitiator.java#L417-L421

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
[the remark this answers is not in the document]
stIncMale: > Just wondering whether we have some clean way to test it.

Not without deploying an unsupported MongoDB version via Evergreen, and then running a specific test against it. I don't want to do that, because the efforts are not worth the benefit of having such a test. But if someone does such a change, I don't have objections in principle.

NathanQingyangXu (approved):

PR #69, opened 2025-03-17
implement loading by primary key

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-22

The first "select statement" translation ticket, though the simplest one (where clause only contains primary key equalness comparison). Very business logic intensitve PR. 

I added comments to explain the various edge cases of business logic and hope that would help.

PR #80, opened 2025-03-26
Add internal docs and do other improvements

stIncMale:
- This PR contains a follow-up to https://github.com/mongodb/mongo-hibernate/pull/69#discussion_r2008516721, but not only that.
- See also https://github.com/mongodb/mongo-hibernate/pull/80#discussion_r2493998451, which explains the proposed `@mongoCme` tag.
- Updated dependencies/plugins. Notably, this allows us to build with Java SE 25, which wasn't working before the PR.


src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -46,6 +46,6 @@ protected DatabaseVersion getMinimumSupportedVersion() {
 
     @Override
     public SqlAstTranslatorFactory getSqlAstTranslatorFactory() {
-        return new MongoTranslatorFactory();
+        return MongoTranslatorFactory.INSTANCE;
```
stIncMale, line 49: This way we work around the issue fixed in https://github.com/hibernate/hibernate-orm/pull/9900, because we are not getting that fix in Hibernate ORM 6.x.

src/main/java/com/mongodb/hibernate/internal/translate/MongoTranslatorFactory.java
```diff
@@ -26,7 +26,14 @@
 import org.hibernate.sql.model.ast.TableMutation;
 import org.hibernate.sql.model.jdbc.JdbcMutationOperation;
 
+/**
+ * Tread-safe.
+ */
 public final class MongoTranslatorFactory implements SqlAstTranslatorFactory {
```
stIncMale, line 32: Hibernate ORM does not document that implementations of `SqlAstTranslatorFactory` must be thread-safe, so we should document that for our `MongoTranslatorFactory`.

NathanQingyangXu: For such bootstrapping stuff, usually no need to sepcify type safety for everything will be done sequentially with well-defined dependency or order. There is no need of concurrency for bootstrapping only happens once and its  perf could be compromised a little bit.

NathanQingyangXu (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/MongoTranslatorFactory.java
NathanQingyangXu: sorry, I made a mistake here. Yeah, I think `SqlAstTranslatorFactory` should be thread-safe for it will be invoked concurrently.

You are right. This doc is good to have.

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
```diff
@@ -23,13 +23,22 @@
...
 import org.hibernate.boot.spi.AdditionalMappingContributor;
 import org.hibernate.boot.spi.InFlightMetadataCollector;
 import org.hibernate.boot.spi.MetadataBuildingContext;
 import org.hibernate.mapping.PersistentClass;
 
+/**
+ * The instance methods like {@link #getContributorName()}, {@link #contribute(AdditionalMappingContributions,
+ * InFlightMetadataCollector, ResourceStreamLocator, MetadataBuildingContext)} are called multiple times if multiple
+ * {@link Metadata} instances are {@linkplain MetadataSources#buildMetadata() built} using the same
+ * {@link BootstrapServiceRegistry}.
```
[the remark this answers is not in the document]
stIncMale: > This Javadoc is visible to public even though it is within internal package, right?

Yes, our whole codebase is publicly visible.

> To my best knowledge, I don't expect any Hibernate contributor methods would be called more than once in one SessionFactory.

`SessionFactory` has nothing to do with the invocations of `MongoAdditionalMappingContributor.contribute`. This method is called neither by `SessionFactory`, nor even when a `SessionFactory` instance is built by the `SessionFactoryBuilder.build` method.

> Putting it here seems to hurt our product for they might clarify some doubts to us but would confuse others (including me).

* This is not our public API documentation, because the documented class is not part of our API. Thus, it does not apply to "others".
* What do you mean exactly by "confuse"? (Does the proposed documentation say something that is false? Does it make one to believe something that is false because it is ill-formulated?)

src/main/java/com/mongodb/hibernate/internal/translate/MongoTranslatorFactory.java
stIncMale: Can this be resolved?

NathanQingyangXu (commented):
I would suggest closing this PR for it only ends up with confusion.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
NathanQingyangXu: Thanks for your raising the issue and it is a quick fix. On the other hand, it seems trivial to me for we only avoid one unnecessary duplicated object creation (which usually is not a big deal for the creation is not costly).

NathanQingyangXu: The other id generator defect is more important and hard to fix from Hibernate side. I created a PR at https://github.com/hibernate/hibernate-orm/pull/9965. Let us see.

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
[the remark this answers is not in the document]
NathanQingyangXu: > This is not our public API documentation, because the documented class is not part of our API. Thus, it does not apply to "others".

Given our codebase is public, there is still possibility that the doc could be read by others (e.g. `community contributor` or Hibernate developer's browsing out of curiosity). If the project is private, I would have no such concern.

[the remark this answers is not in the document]
NathanQingyangXu: > What do you mean exactly by "confuse"? (Does the proposed documentation say something that is false? Does it make one to believe something that is false because it is ill-formulated?)

Let me elaborate on why I got confused.

Hibernate's contributor is to be created by default constructor (via JDK's ServiceLoader mechanism or `ClassLoaderService` in Hibernate), so it is a typical one-use object in the sense that once it is loaded (or created), its `contribute()` method is invoked and then that is it and the contributor object is subject to garbage collecting. Next time a new instance will be created and loaded, but I don't think it will be reused more than once.

Such contributor hooking point is hard-coded in Hibernate code as below:
```

		final Collection<AdditionalMappingContributor> additionalMappingContributors = classLoaderService.loadJavaServices( AdditionalMappingContributor.class );
		additionalMappingContributors.forEach( (contributor) -> {
			contributions.setCurrentContributor( contributor.getContributorName() );
			try {
				contributor.contribute(
						contributions,
						metadataCollector,
						classLoaderService,
						rootMetadataBuildingContext
				);
			}
			finally {
				contributions.setCurrentContributor( null );
			}
		} );
```

As you can see, for each contributor, its two methods will be called only once after it is loaded or created. So it seems a typical one-use scenario as our various mql translators.

src/main/java/com/mongodb/hibernate/internal/extension/service/StandardServiceRegistryScopedState.java
```diff
@@ -49,6 +51,11 @@ public MongoConfiguration getConfiguration() {
         return config;
     }
 
+    /**
+     * The instance methods like {@link #contribute(StandardServiceRegistryBuilder)} are called multiple times if
+     * multiple {@link StandardServiceRegistry} instances are {@linkplain StandardServiceRegistryBuilder#build() built}
+     * using the same {@link BootstrapServiceRegistry}.
+     */
```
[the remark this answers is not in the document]
NathanQingyangXu: So from my understanding, Hibernate's `SessionFactory` is pretty heavy-weighted (it is super slow for it tries to prepare everything so runtime perf could be as fast as possible; one example is it tries to finish `static` SQL translations as much as possible so they could be reused). For that reason, for typical usage, it is seldom created multiple times (in typical Spring JPA usage, one single `SessionFactory` is bound to the global Spring Bean Context), unless there are good reasons.

If multiple `SessionFactory` is justified (maybe each one uses a subset of the `META-INF/services` so they could share the same jar library without conflict; then it requires customized `BootstrapServiceRegistry`, one of whose components is the `ClassLoaderService` which could be customized), different `BootstrapServiceRegistry` is expected to be created for each of them. Different `BootstrapServiceRegistry` leads to different `StandardServiceRegistry` (as explained above, maybe due to different class loaders), otherwise why create different `SessionFactory` in the first place?

During runtime, one `SessionFactory` has one single `ServiceRegistry` and its default implementation is `StandardServiceRegistry` as its API below (in `org.hibernate.engine.spi.SessionFactoryImplementor`):
```
ServiceRegistryImplementor getServiceRegistry();
```
So I am struggling to understand why we need to create multiple `StandardServiceRegistry` in one `SessionFactory`. Yeah, API didn't disallow, but if it is too rare to happen, should we keep from bothering our user who is ofen tech-savy and would get puzzled?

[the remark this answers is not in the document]
NathanQingyangXu: The above comment also explains why I got puzzled by your previous PR; but you modified it to make it aligned with reality so I approved that.

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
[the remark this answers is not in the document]
vbabanin: > Next time a new instance will be created and loaded, but I don't think it will be reused more than once.

JDK ServiceLoader does perform some caching, as described in the "Timing of provider discovery" section of the [official documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ServiceLoader.html).

Hibernate also caches loaded services and reuses the same instance of `MongoAdditionalMappingContributor`. Relevant code snippet from `AggregatedServiceLoader`:

```
@Override
public Collection<S> getAll() {
    if ( cache == null ) {
        /*
         * loadAll() uses ServiceLoader.Provider.get(), which doesn't cache the instance internally,
         * unlike ServiceLoader.iterator().
         * Looking at how https://hibernate.atlassian.net/browse/HHH-8363 was solved,
         * waiting for Hibernate ORM to shut down before clearing the service caches,
         * it seems caching of service instances is important, or at least used to be important.
         * Thus we cache service instances ourselves to avoid any kind of backward-incompatibility.
         * If one day we decide caching isn't important, this cache can be removed...
         */
        cache = Collections.unmodifiableCollection( loadAll() );
    }
    return cache;
}
```
As long as it reflects how things actually work and isn't misleading, which is not in this case - I think there is no harm in documenting it.

src/main/java/com/mongodb/hibernate/internal/translate/MongoTranslatorFactory.java
```diff
@@ -26,7 +26,14 @@
 import org.hibernate.sql.model.ast.TableMutation;
 import org.hibernate.sql.model.jdbc.JdbcMutationOperation;
 
+/**
+ * Tread-safe.
```
vbabanin, line 30:
```suggestion
 * Thread-safe.
```

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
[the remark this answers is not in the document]
NathanQingyangXu: Good point and food for thoughts.
I reconsidered the specific service provider in question, i.e. `MongoAdditionalMetadataContributor`. Yeah, once it is loaded into cache the same service provider could be reused later. That is something new I got from your feedback and I appreciate it.

However, from my knowledge, for one specific `SessionFactory`, all the `AdditionalMetadataContributor` service provider instances would be ONLY used for once. Suppose they (we could create multiple `AdditionalMetadataContributor` service providers per good practice; currently we reuse `MongoAdditionalMetadataContributor` for all concerns, including `_id` hardcoding and supportability checking) are cached internally, the cached providers would not have opportunity to be reused again.

`AggregatedServiceLoader` has internal cache but again, we don't reuse `AdditionalMetadataContributor` for one Hibernate bootstrapping, so whether its service providers are cached or not seems not important for our concern.

NathanQingyangXu (approved):

src/main/java/com/mongodb/hibernate/service/spi/MongoConfigurationContributor.java
```diff
@@ -54,12 +54,13 @@
...
 
     /**
-     * Configures the MongoDB Extension for Hibernate ORM. This method is called once per instance of
-     * {@link StandardServiceRegistry} that has this {@link MongoConfigurationContributor}
-     * {@linkplain StandardServiceRegistryBuilder#addService(Class, Service) added}.
+     * Configures the MongoDB Extension for Hibernate ORM.
      *
      * @param configurator The {@link MongoConfigurator} pre-configured with {@linkplain Configurable#configure(Map)
      *     configuration properties}.
+     * @mongoCme This method is called once per instance of {@link StandardServiceRegistry} that has this
```
stIncMale: Given that `MongoConfigurationContributor` is part of our API, you can see how the new tag looks in the generated docs.

src/main/java/com/mongodb/hibernate/internal/extension/service/StandardServiceRegistryScopedState.java
```diff
@@ -33,13 +33,17 @@
...
+import org.hibernate.boot.registry.BootstrapServiceRegistry;
 import org.hibernate.boot.registry.StandardServiceInitiator;
+import org.hibernate.boot.registry.StandardServiceRegistry;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.service.Service;
 import org.hibernate.service.UnknownServiceException;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
 import org.jspecify.annotations.Nullable;
 
+/** @mongoCme Thread-safe. */
```
stIncMale: In most other places I said "Must be thread-safe", but here it is just "Thread-safe". The difference is that here the behavior is something we provide based on what we need. In other places we must provide thread-safety because of the Hibernate ORM API parts we implement/extend.

src/integrationTest/java/com/mongodb/hibernate/embeddable/StructAggregateEmbeddableIntegrationTests.java
stIncMale: This file was reformatted because of the `plugin-palantir` version bump.

vbabanin (approved):
LGTM!

PR #83, opened 2025-04-04
Add `MongoIdFieldNameIntegrationTests.testEntityWithIdColumnAnnotationWithInvalidNameElement`

stIncMale:


src/integrationTest/java/com/mongodb/hibernate/id/ObjectIdAsIdIntegrationTests.java
```diff
@@ -58,8 +58,7 @@ void insert() {
         var item = new Item();
         item.id = new ObjectId(1, 0);
         sessionFactoryScope.inTransaction(session -> session.persist(item));
-        assertThat(mongoCollection.find())
-                .containsExactly(new BsonDocument().append(ID_FIELD_NAME, new BsonObjectId(item.id)));
+        assertThat(mongoCollection.find()).containsExactly(new BsonDocument(ID_FIELD_NAME, new BsonObjectId(item.id)));
```
NathanQingyangXu, line 61: Optional. We could consider removing the verboase `hashCode` and `equals` implemention of `Item` class in another `ObjectIdIntegrationTests` testing case and use assertJ's `assertThat(..).usingRecursiveComparison().isEqualTo(...)`, which improves both readability and maintainability.

NathanQingyangXu (approved):

src/integrationTest/java/com/mongodb/hibernate/id/ObjectIdAsIdIntegrationTests.java
stIncMale: I initially did not want to use `usingRecursiveComparison` because it ignores the overridden `Object.equals` method. But then I discovered `usingOverriddenEquals`.

My understanding is that we have to use

```java
assertThat(actual)
        .usingRecursiveComparison()
        .usingOverriddenEquals()
        .withStrictTypeChecking()
        .isEqualTo(expected);
```

to correctly use `usingRecursiveComparison` in most if not all cases. This is quite wordy, and one may almost guarantee that this incantation won't be spelled out correctly in all cases. That's why I introduced `MongoTestAssertions.assertEquals`: 63cf56b8c7eea218758170a3f02ded1eead9d4f2.

NathanQingyangXu: from my understanding, in 99% of cases the following
```
assertThat(actual)
        .usingRecursiveComparison()
        .isEqualTo(expected);
```
is enough.
From the Javadoc of `usingRecursiveComparison()`, `usingOverrideEquals()` is the default behaviour.

`withStrictTypeChecking` might only be needed when polymorphism is involved (it seems never used up to now), so in most of cases we simply not need to bother with its usage.

but using it is optional, not mandatory.

NathanQingyangXu: after further experimentation, you are right. The overriden equals is not invoked, though 
Screenshot 2025-04-10 at 10 50 31 AM

PR #84, opened 2025-04-04
Improve how identifiers are treated

stIncMale:


src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -91,4 +92,9 @@ public void contribute(TypeContributions typeContributions, ServiceRegistry serv
         typeContributions.contributeJavaType(ObjectIdJavaType.INSTANCE);
         typeContributions.contributeJdbcType(ObjectIdJdbcType.INSTANCE);
     }
+
+    @Override
+    @Nullable public String toQuotedIdentifier(@Nullable String name) {
+        return name;
```
stIncMale, line 98: This single change makes all the new tests pass, and without it, they all fail. Nothing else related to quoting both in `Dialect` (including `Dialect.buildIdentifierHelper`) and in `DatabaseMetaData` had any observable effect on the tests added.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -91,4 +92,9 @@ public void contribute(TypeContributions typeContributions, ServiceRegistry serv
         typeContributions.contributeJavaType(ObjectIdJavaType.INSTANCE);
         typeContributions.contributeJdbcType(ObjectIdJdbcType.INSTANCE);
     }
+
+    @Override
+    public @Nullable String toQuotedIdentifier(@Nullable String name) {
+        return name;
```
stIncMale, line 98: This single change makes all the new tests pass, and without it, they all fail. Nothing else related to quoting both in `Dialect` (including `Dialect.buildIdentifierHelper`) and in `DatabaseMetaData` had any observable effect on the tests added.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+        int v;
+    }
+
+    @Entity
+    @Table(name = InDoubleQuotesMixedCase.COLLECTION_NAME)
+    static class InDoubleQuotesMixedCase {
+        static final String COLLECTION_NAME = "\"collection in double quotes Mixed Case\"";
+        static final String FIELD_NAME = "\"field in double quotes Mixed Case\"";
+        static final String ACTUAL_COLLECTION_NAME = "collection in double quotes Mixed Case";
+        static final String ACTUAL_FIELD_NAME = "field in double quotes Mixed Case";
```
stIncMale, line 216: Unfortunately, it seems impossible to make identifiers like `\"collection in double quotes Mixed Case\"` to not be stripped from quotes by Hibernate ORM. Apparently, it has logic somewhere that ignores any configuration/overrides related to quoting, and strips double and single quotes anyway.

stIncMale: That place seems to be here https://github.com/hibernate/hibernate-orm/blob/0a4d551b32801ae15e99d2f8ac0750303ecda2b6/hibernate-core/src/main/java/org/hibernate/boot/model/naming/Identifier.java#L159-L173.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+        int v;
+    }
+
+    @Entity
+    @Table(name = InSingleQuotesMixedCase.COLLECTION_NAME)
+    static class InSingleQuotesMixedCase {
+        static final String COLLECTION_NAME = "`collection in single quotes Mixed Case`";
+        static final String FIELD_NAME = "`field in single quotes Mixed Case`";
+        static final String ACTUAL_COLLECTION_NAME = "collection in single quotes Mixed Case";
+        static final String ACTUAL_FIELD_NAME = "field in single quotes Mixed Case";
```
NathanQingyangXu, line 175: yeah, Hibernate has its own quotes so the above ACTUAL IDENTIFIER makes sense, as long as the quote pairs surround the whole identifier. For completeness, we could cover `[]` as well.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+
+        @Id
+        int id;
+
+        @Column(name = WithSpaceAndDotMixedCase.FIELD_NAME)
+        int v;
+    }
+
+    @Entity
+    @Table(name = InSingleQuotesMixedCase.COLLECTION_NAME)
```
NathanQingyangXu, line 170: technically speaking, ``` is not the same as `'`. So more accurate verbiage is `backtick`, not `single quote`.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
NathanQingyangXu: yeah, it seems so. If the Hibernate quoting at the start and end is hard-coded or not configurable, I think we could remove those `MixedCases` testing cases for we don't need to test Hibernate if its behaviour is working as expected.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+        static final String COLLECTION_NAME = "collection ending with double quote\"";
+        static final String FIELD_NAME = "field ending with double quote\"";
+
+        @Id
+        int id;
+
+        @Column(name = EndingWithDoubleQuote.FIELD_NAME)
+        int v;
+    }
+}
```
NathanQingyangXu, line 250: seems `EndingWithDoubleQuoteOnly` might be more accurate for the `MixedCase` ends in the same way as well, right?

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
NathanQingyangXu: It is a great news! Thanks for your PR. Withotu some minor changes it could be merged.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+
+    @Test
+    void endingWithDoubleQuote() {
+        var item = new EndingWithDoubleQuote();
+        sessionFactoryScope.inTransaction(session -> session.persist(item));
+        assertThat(mongoCollectionEndingWithDoubleQuote.find())
+                .containsExactly(new BsonDocument()
+                        .append(ID_FIELD_NAME, new BsonInt32(item.id))
+                        .append(EndingWithDoubleQuote.FIELD_NAME, new BsonInt32(item.v)));
+        sessionFactoryScope.inTransaction(session -> session.get(EndingWithDoubleQuote.class, item.id));
```
NathanQingyangXu, line 148: the above testing cases seem to follow similar pattern and is paramerized by two inputs: 1) entity class; 2) identifier name supplier, so it is doable to avoid the duplication by JUnit 5's feature.

but this is optional and over-abstracting in testing code not necessarily helps on code maintaining.

NathanQingyangXu (changes requested):

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
stIncMale: I don't think `EndingWithDoubleQuoteOnly` is better than `EndingWithDoubleQuote`. It is true that `EndingWithDoubleQuote` does not say explicitly that it does not start with a double quote, but `EndingWithDoubleQuoteOnly` does not say that either. Instead, it says that the name ends only with a double quote, as if there are other tests where it ends with not just a double quote, but with a double quote and something else.

To me it the meaning of `EndingWithDoubleQuote` seems quite clear, given that there is `StartingWithDoubleQuote` and `InDoubleQuotesMixedCase`.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
stIncMale: Indeed, thank you. Done in fb7bd7d39e818e375a70e40c6e05cc455758d3db.

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
stIncMale: I don't expect there to be a meaningful improvement if we try and replace the current simple code by extracting the common code and introducing parameters. Given that you marked this thread as optional, I am resolving it (the PR author, despite not being able to resolve other conversations started by a reviewer, is free to resolve optional conversations - that's what optional comments are).

src/integrationTest/java/com/mongodb/hibernate/IdentifierIntegrationTests.java
```diff
@@ -0,0 +1,250 @@
...
+    @Override
+    public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
+        this.sessionFactoryScope = sessionFactoryScope;
+    }
+
+    @Entity
+    @Table(name = WithSpaceAndDotMixedCase.COLLECTION_NAME)
+    static class WithSpaceAndDotMixedCase {
+        static final String COLLECTION_NAME = "collection with space and .dot Mixed Case";
+        static final String FIELD_NAME = "field with space Mixed Case";
```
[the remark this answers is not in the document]
stIncMale: > Seems nontrivial to adhere to defensitive programming to provide relevant exception when field name contains `.`.

I tried doing that in 00e94961a3eeb67c223e064bce6d9caaaa65581a, and also added tests.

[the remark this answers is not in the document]
stIncMale: > it seems more sensible to separate collection name identifier from field identifier classes, though I am ok with current status.

I fail to see a benefit of doing so.

NathanQingyangXu (approved):

PR #86, opened 2025-04-14
add assertion that AstProjectStage must have at least one specification

NathanQingyangXu:
Seems a minor loophole and tech debt from previous PR. I checked our mongoast classes accepting collection constructor parameters and found only `AstProjectStage` requires this assertion that the collection should be non-empty.

In the future PRs (including the current `simple select query translation`), such collection emptiness assertion would be added whenever applied. Just found this bad-case from previous PRs and tried to fix it.

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
+++ b/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
@@ -88,21 +88,6 @@ public final class MongoAssertions {
     }
 
     /**
-     * Asserts that {@code value} is {@code true} with message.
-     *
-     * @param message the message to explain details.
-     * @param value A value to check.
-     * @return {@code true}.
-     * @throws AssertionError If {@code value} is {@code false}.
-     */
-    public static boolean assertTrue(String message, boolean value) throws AssertionError {
-        if (!value) {
-            throw new AssertionError(message);
-        }
-        return true;
-    }
-
-    /**
      * Asserts that {@code value} is {@code false}.
      *
      * @param value A value to check.
@@ -112,21 +97,6 @@ public final class MongoAssertions {
     public static boolean assertFalse(boolean value) throws AssertionError {
         if (value) {
             throw new AssertionError();
-        }
-        return false;
-    }
-
-    /**
-     * Asserts that {@code value} is {@code false} with message.
-     *
-     * @param message the message to explain details.
-     * @param value A value to check.
-     * @return {@code false}.
-     * @throws AssertionError If {@code value} is {@code true}.
-     */
-    public static boolean assertFalse(String message, boolean value) throws AssertionError {
-        if (value) {
-            throw new AssertionError(message);
         }
         return false;
     }
--- a/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
+++ b/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
@@ -128,6 +128,6 @@ public final class MongoAssertions {
         if (value) {
             throw new AssertionError(message);
         }
-        return true;
+        return false;
     }
 }
--- a/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
+++ b/src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
@@ -115,4 +115,19 @@ public final class MongoAssertions {
         }
         return false;
     }
+
+    /**
+     * Asserts that {@code value} is {@code false} with message.
+     *
+     * @param message the message to explain details.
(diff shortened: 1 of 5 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstProjectStage.java
```diff
@@ -19,7 +19,14 @@
...
 import org.bson.BsonWriter;
 
+import static com.mongodb.hibernate.internal.MongoAssertions.assertFalse;
+
 public record AstProjectStage(List<? extends AstProjectStageSpecification> specifications) implements AstStage {
+
+    public AstProjectStage {
+        assertFalse("Projection stage must have at least one specification", specifications().isEmpty());
+    }
+
```
NathanQingyangXu, line 29: the `compact constructor` of Java record seems to be the place to insert our assertion logic. See https://mikemybytes.com/2022/02/16/java-records-and-compact-constructors/

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstProjectStage.java
```diff
@@ -16,10 +16,17 @@
...
 
+import static com.mongodb.hibernate.internal.MongoAssertions.assertFalse;
+
 import java.util.List;
 import org.bson.BsonWriter;
 
 public record AstProjectStage(List<? extends AstProjectStageSpecification> specifications) implements AstStage {
+
+    public AstProjectStage {
+        assertFalse("Projection stage must have at least one specification", specifications.isEmpty());
```
stIncMale, line 27: This static message provides no information that is not already clear from the code. Given that assertions can be violated only when there are bugs in the product, `AssertionError`s do not have a requirement to be understood without looking at the product code. Thus, having static error messages in `AssertionError`s is completely useless.

Let's remove the message and have only `assertFalse(specifications.isEmpty());`

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
```diff
@@ -115,4 +115,19 @@ public static boolean assertFalse(boolean value) throws AssertionError {
...
+     * @param message the message to explain details.
+     * @param value A value to check.
+     * @return {@code false}.
+     * @throws AssertionError If {@code value} is {@code true}.
+     */
+    public static boolean assertFalse(String message, boolean value) throws AssertionError {
+        if (value) {
+            throw new AssertionError(message);
+        }
+        return true;
```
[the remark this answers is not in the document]
NathanQingyangXu: changed in the sole commit including the other comment resolving.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstProjectStage.java
NathanQingyangXu: done

src/main/java/com/mongodb/hibernate/internal/MongoAssertions.java
```diff
@@ -115,4 +115,19 @@ public static boolean assertFalse(boolean value) throws AssertionError {
...
+
+    /**
+     * Asserts that {@code value} is {@code false} with message.
+     *
+     * @param message the message to explain details.
+     * @param value A value to check.
+     * @return {@code false}.
+     * @throws AssertionError If {@code value} is {@code true}.
+     */
+    public static boolean assertFalse(String message, boolean value) throws AssertionError {
```
stIncMale, line 127: Actually, now we don't need this method anymore, do we?

NathanQingyangXu: we don't need it in this PR. Just wonder whether we can retain it for future PR. We had counterpart of `assertTrue` already, so I think it is fine to retain it? Let me know I could delete it without issue.

stIncMale: We do have `MongoAssertions.assertTrue(String message, boolean value)`, but it's not used. And it is unlikely that that method or the new `assertFalse` method it to be used, as that requires us to need to accompany the assertion error with runtime data in its message.

I propose removing both methods, because having them without a legitimate usage only increases the chances that they will be used to supply a useless static assertion message.

NathanQingyangXu: removed both

stIncMale (approved):

PR #89, opened 2025-04-23
Support `@Struct` aggregate embeddable mapping

stIncMale:
HIBERNATE-59


NathanQingyangXu (commented):
Looks great! It even finished recursive embedding!
Only left some minor comments.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -0,0 +1,214 @@
...
+    public <X> ValueBinder<X> getBinder(JavaType<X> javaType) {
+        return new Binder<>(javaType);
+    }
+
+    @Override
+    public <X> ValueExtractor<X> getExtractor(JavaType<X> javaType) {
+        return new Extractor<>(javaType);
+    }
+
+    /** Thread-safe. */
```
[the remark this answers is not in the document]
stIncMale: If a new instance were indeed created for each use, then `Binder`/`Extractor` would not need to be thread-safe. But that is not necessarily the case, see `org.hibernate.type.AbstractStandardBasicType`, for example, that remembers and reuses instances of `Binder`/`Extractor`. Therefore, it seem like generally speaking, the implementations of these interfaces must be thread-safe.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Provides conversion methods between {@link BsonValue}s, which our {@link PreparedStatement}/{@link ResultSet}
+ * implementation uses under the hood and rarely exposes, and domain values we usually use when setting parameter values
+ * on our {@link PreparedStatement}, or retrieving column values from a {@link ResultSet}.
+ */
+public final class ValueConversions {
+    private ValueConversions() {}
+
+    static BsonValue toBsonValue(Object value) {
+        assertNotNull(value);
```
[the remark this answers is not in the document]
stIncMale: I don't know, no necessarily, as it is not impossible that `null`s will be handled outside of `ValueConversions.toBsonValue`. We'll see how things are and what's better once we implement that ticket.

However, I merged `main`, and replaced `AbstractMqlTranslator.toBsonValue` with `ValueConversions.toBsonValue`. Now, there are two different reasons why `null` can't be passed to `ValueConversions.toBsonValue`, and it's not even technically possible handle both of them in `ValueConversions.toBsonValue`.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
```diff
@@ -0,0 +1,71 @@
...
+            String template,
+            String placeholder,
+            String aggregateParentReadExpression,
+            String columnExpression,
+            AggregateColumn aggregateColumn,
+            Column column) {
+        var aggregateColumnType = aggregateColumn.getTypeCode();
+        if (aggregateColumnType == JDBCType.STRUCT.getVendorTypeNumber()) {
+            return format(
+                    "unused from %s.aggregateComponentCustomReadExpression",
```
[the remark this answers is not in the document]
stIncMale: The method name is `MongoAggregateSupport.aggregateComponentCustomReadExpression`, not `MongoAggregateSupport.aggregateComponentCustomReadExpression()`. Furthermore `()` implies that the method has no parameters, which is not the case.

stIncMale:
Added @vbabanin as the secondary reviewer.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,272 @@
...
+                        .append("nested1", new BsonDocument("a", new BsonInt32(2)))
+                        .append(
+                                "nested2",
+                                new BsonDocument()
+                                        .append("a", new BsonInt32(3))
+                                        .append(
+                                                "nested",
+                                                new BsonDocument()
+                                                        .append("a", new BsonInt32(4))
+                                                        .append("b", new BsonInt32(5)))));
```
katcharov, line 112: Is it possible to cleanly parse multiline json strings for tests?

NathanQingyangXu:
I finished the first-round review. Will go deeper to inspect more tech details, particularly on `MongoStructJdbcType`'s internal details.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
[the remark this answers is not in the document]
NathanQingyangXu: thanks for refactoring `AbstractMqlTranslator.toBsonValue`. That is a tech debt and I am happy you fixed it.

NathanQingyangXu (commented):
just some minor comments

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    @Override
+    public EmbeddableMappingType getEmbeddableMappingType() {
+        return assertNotNull(embeddableMappingType);
+    }
+
+    @Override
+    public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options) {
+        var embeddableMappingType = assertNotNull(this.embeddableMappingType);
```
NathanQingyangXu, line 94: maybe we should throw exception when `embeddableMappingType.isPolymorphic()` return true:
```
        if (embeddableMappingType.isPolymorphic()) {
        }
```

Here is an example testing case which would be broken currently:
```
    @Test
    void test(SessionFactoryScope scope) {
        var structHolder = new StructHolder();
        structHolder.id = 1;
        structHolder.animal = new Cat();
        structHolder.animal.name = "Lucy";

        scope.inTransaction(session -> session.persist(structHolder));

        var fetchedStructHolder = scope.fromTransaction(session -> session.get(StructHolder.class, 1));
        MongoTestAssertions.assertEquals(structHolder, fetchedStructHolder);
    }

    @Entity
    @Table(name = "structHolder")
    static class StructHolder {
        @Id int id;
        @Embedded
        Animal animal;
    }

    @Embeddable
    @Struct(name = "Animal")
    static class Animal {
        String name;
    }

    @Embeddable
    @Struct(name = "Cat")
    static class Cat extends Animal {
        String color;
    }

    @Embeddable
    @Struct(name = "Dog")
    static class Dog extends Animal {
        String breed;
    }
}
```

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -0,0 +1,213 @@
...
+        var result = new BsonDocument();
+        var jdbcValueCount = embeddableMappingType.getJdbcValueCount();
+        for (int columnIndex = 0; columnIndex < jdbcValueCount; columnIndex++) {
+            var jdbcValueSelectable = embeddableMappingType.getJdbcValueSelectable(columnIndex);
+            if (jdbcValueSelectable.isFormula()) {
```
NathanQingyangXu, line 99: Could we add some testing case to cover this logic branch? I experimented using the following test case and it throws exception during model scanning and it seems Formula is not allowed in aggregated columns.

```
    @Test
    void test(SessionFactoryScope scope) {
        Person person = new Person();
        person.name = "John";
        person.age = 17;

        var structHolder = new StructHolder();
        structHolder.id = 1;
        structHolder.person = person;

        scope.inTransaction(session -> session.persist(structHolder));

        var fetchedStructHolder = scope.fromTransaction(session -> session.get(StructHolder.class, 1));
        MongoTestAssertions.assertEquals(structHolder, fetchedStructHolder);
    }

    @Entity
    @Table(name = "structHolder")
    static class StructHolder {
        @Id int id;
        Person person;
    }

    @Embeddable
    @Struct(name = "Person")
    static class Person {
        String name;
        int age;

        @Formula("name + ' ' + age")
        String nameWithAge;
    }
}
```
so it seems the above logic branch won't be hit at all.

NathanQingyangXu (changes requested):
I finished exploring `MongoStructJdbcType` using various kinds of feature combination and found some edge cases and left comments with reproducible entity and testing code.

Seems its logic is good enough for now, but when later on more advanced features (e.g. table joining or `@ManytoOne`) are implemented, its logic might need to tweak accordingly.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,272 @@
...
+    @Test
+    void flattenedValues() {
+        var item = new ItemWithFlattenedValues();
+        item.id = new EmbeddableValue();
+        item.id.a = 1;
+        item.flattened1 = new EmbeddableValue();
+        item.flattened1.a = 2;
+        item.flattened2 = new EmbeddablePairValue1(3, new EmbeddablePairValue2(4, 5));
+        sessionFactoryScope.inTransaction(session -> session.persist(item));
+        assertThat(mongoCollection.find())
```
stIncMale, line 70: Hibernate ORM 7.0 deprecated more Hibernate-specific methods on `SessionFactory` in favour of those from Jakarta `EntityManager`. We should start using them now, instead of producing more code that will have to be changed once we upgrade to Hibernate ORM 7.0.

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
```diff
@@ -83,8 +83,8 @@ private static void setIdentifierColumnName(PersistentClass persistentClass) {
         var idColumns = identifier.getColumns();
         if (idColumns.size() > 1) {
             throw new FeatureNotSupportedException(format(
-                    "%s: %s does not support [%s] field spanning multiple columns %s",
-                    persistentClass, MONGO_DBMS_NAME, ID_FIELD_NAME, idColumns));
+                    "%s: %s does not support primary key spanning multiple columns %s",
```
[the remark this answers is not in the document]
NathanQingyangXu: it is not very weird from Hibernate perspective for traditional `@Embeddable` id field does mapped to multiple columns (e.g. see https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#identifiers-composite-aggregated), but I agree new change is better.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,272 @@
...
+@DomainModel(
+        annotatedClasses = {
+            EmbeddableIntegrationTests.ItemWithFlattenedValues.class,
+            EmbeddableIntegrationTests.ItemWithNestedValues.class,
+            EmbeddableIntegrationTests.Unsupported.ItemWithNestedValueHavingNonInsertable.class,
+            EmbeddableIntegrationTests.Unsupported.ItemWithNestedValueHavingAllNonInsertable.class,
+            EmbeddableIntegrationTests.Unsupported.ItemWithNestedValueHavingNonUpdatable.class
+        })
+@ExtendWith(MongoExtension.class)
+class EmbeddableIntegrationTests implements SessionFactoryScopeAware {
```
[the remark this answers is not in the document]
NathanQingyangXu: This is optional. Feel free to close it. We could always refactor testing packages when it is a better timing.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
NathanQingyangXu: replacing `get` with `find` method is fine, but given we are still using v6.6, it is not urgent to embrace Jakarta for now; when we switch to v7, it seems pretty straightforward to refactor for I don't think we use Hibernate-specific methods (which have no Jakarta counterparts) a lot.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: > it seems Formula is not allowed in aggregated columns
> ...
> it seems the above logic branch won't be hit at all.

Then we should replace the current check with an assertion.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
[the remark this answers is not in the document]
stIncMale: Java arrays and Java SE API collections (`Collection`, `Map`), cannot be mapped to types that Hibernate ORM calls ["aggregate types"](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#embeddable-mapping-aggregate). Thus, it wouldn't be correct to put the tests for arrays and collections in the `aggregate` package. Furthermore, as far as I can see, Hibernate ORM also does not see arrays and collections as ["embeddable values"](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#embeddables).

Arrays (except for `byte[]`/`Byte[]` and `char[]`/`Character[]`, which are considered ["basic values"](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#basic-mapping-array)) and collections (except for [situations when they are considered "basic values"](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#basic-mapping-collection)) seem to form a separate kind of values - ["collections"](https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#collections), which is neither embeddable, nor aggregate (aggregate is a subset of embeddable).

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: This is not a correct way to configure polymorphism in Hibernate ORM. With the code as proposed, Hibernate ORM ignores any persistent attributes of the `Cat`, and stores/reads `Animal`s. The proposed code works without any changes to the PR, but the document stored is

```
{
  "_id": 1,
  "animal": {
    "name": "Lucy"
  }
}
```

and `embeddableMappingType.isPolymorphic()` is `false`. I played with this, but not enough to figure out how to configure polymorphism correctly. However, while doing so I found another issue, and will make a change to fix it in this PR (the issue is not specific to this PR, but it's easy to reproduce with an `@Embeddable`).

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,310 @@
...
+        item.id = 1;
+        item.omitted = new StructAggregateEmbeddableEmptyValue();
+        sessionFactoryScope.inTransaction(session -> session.persist(item));
+        assertThat(mongoCollection.find()).containsExactly(
+                // Hibernate ORM does not store `item.omitted` despite it being non-`null`
+                new BsonDocument(ID_FIELD_NAME, new BsonInt32(item.id)));
+        var loadedItem =
+                sessionFactoryScope.fromTransaction(session -> session.find(ItemWithNestedEmptyValue.class, item.id));
+        // the entity we stored and the entity we loaded are not equal because Hibernate ORM omits `item.omitted`
+        assertNotEquals(item, loadedItem);
```
stIncMale, line 139: The behavior mentioned in the comments looks like another Hibernate ORM bug.

src/main/java/com/mongodb/hibernate/internal/translate/ModelMutationMqlTranslator.java
```diff
@@ -40,7 +41,13 @@ public O translate(@Nullable JdbcParameterBindings jdbcParameterBindings, QueryO
...
-        var mutationCommand = acceptAndYield(tableMutation, COLLECTION_MUTATION);
-        return tableMutation.createMutationOperation(renderMongoAstNode(mutationCommand), getParameterBinders());
+        String sql;
+        if ((TableMutation<?>) tableMutation instanceof TableUpdateNoSet) {
+            sql = "";
+        } else {
+            var mutationCommand = acceptAndYield(tableMutation, COLLECTION_MUTATION);
+            sql = renderMongoAstNode(mutationCommand);
+        }
+        return tableMutation.createMutationOperation(sql, getParameterBinders());
```
stIncMale, line 51: Turned out, Hibernate ORM has an implementation of `TableMutation`, whose `accept` method does nothing. Passing an instance of `TableUpdateNoSet` to `acceptAndYield` results in `AssertionError`, because the `accept` method does nothing, which means, `AstVisitorValueHolder.yield` is not called, which we don't expect.

Without this change, the new `nestedEmptyValue` test fails.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: The reason I observed behavior different from the one observed by you is that whether you add `Animal`, `Cat`, `Dog` to `DomainModel.annotatedClasses` or not makes a difference. If they are not added, the test you posted succeeds, despite not storing the animal correctly (what I said previously). If they are added, then the test fails.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: The reason I observed behavior different from the one observed by you is that whether you add `Animal`, `Cat`, `Dog` to `DomainModel.annotatedClasses` or not makes a difference. If they are not added, the test you posted succeeds, despite not storing the animal correctly (what I said previously). If they are added, then the test fails.

stIncMale: I added the proposed check: aff8089c15c8c399aed379f1f2bcee444a2196c7.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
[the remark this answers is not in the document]
stIncMale: > This is optional. Feel free to close it.

Resolving for the reason specified in the previous comment.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: I checked in b6c9fdf66597ef00b5519610f1b02229842c3ee4 that `@Parent` mapping is not considered a formula, and replaced the formula check with an assertion in 1da9b205637a852305dab4c9efc2e608f4a7e9ab.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,272 @@
...
+                            // Hibernate ORM does not persist `item.omitted`, because all its persistent attributes are
+                            // non-insertable.
+                            new BsonDocument(ID_FIELD_NAME, new BsonInt32(item.id)));
+            assertThatThrownBy(() -> sessionFactoryScope.fromTransaction(
+                            session -> session.get(ItemWithNestedValueHavingAllNonInsertable.class, item.id)))
+                    .isInstanceOf(Exception.class);
+        }
+
+        @Test
+        void nonUpdatable() {
```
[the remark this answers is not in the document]
stIncMale: @katcharov

My thoughts on the "test" prefix:

1. The names of all top-level test classes end with `Tests` or `IntegrationTests`.
2. Every test method is annotated with `@Test`/`@ParameterizedTest`.
3. Now every test method name also starts with `test`. So we have "EmbeddableIntegration**Tests**.@**Test** **test**FlattenedValues". How many more times should we write "test" to make sure it's clear that the method is a test?
4. I don't buy the argument that without the "test" prefix the names of the test methods become weird, as introducing the "test" prefix does not affect what I write in the rest of the method name.

I made the change in a0e0ecfdf82c940a1ede242420c6d857e66cd72c to avoid spending more time arguing.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: To work in all situations, strings must use the canonical extended JSON format (for example, to represent `ObjectId`, or `int64`), and that format is neither particularly writeable, nor particularly readable. I am prefer writing and reading code that uses `BsonDocument` API instead.

NathanQingyangXu (dismissed):

src/main/java/com/mongodb/hibernate/jdbc/MongoPreparedStatement.java
```diff
@@ -193,24 +185,23 @@ public void setObject(int parameterIndex, Object x, int targetSqlType) throws SQ
...
+            value = toBsonValue(v);
+        } else if (targetSqlType == JDBCType.STRUCT.getVendorTypeNumber()) {
+            if (!(x instanceof BsonDocument v)) {
+                throw fail();
+            }
+            value = v;
         } else {
-            throw new SQLFeatureNotSupportedException("To be implemented in scope of Array / Struct tickets");
+            throw new SQLFeatureNotSupportedException(format(
+                    "Parameter value [%s] of SQL type [%s] with index [%d] is not supported",
```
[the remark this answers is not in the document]
stIncMale: Thank you, fixed in 63702d6240bef11639c642e0481e92e0bb9da872.

Your comment is no visible on the https://github.com/mongodb/mongo-hibernate/pull/89 page 🤷

NathanQingyangXu (dismissed):

NathanQingyangXu (commented):
LGTM. Great job.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
NathanQingyangXu: This seems not a Hibernate ORM bug. It has a property (`false` by default) to configure whether to persist empty embeddable: `hibernate.create_empty_composites.enabled`. 

For instance, when I added the following `hibernate-testing` setting to this class:
```
@ServiceRegistry( settings = @Setting( name = AvailableSettings.CREATE_EMPTY_COMPOSITES_ENABLED, value = "true" ) )
```
you can see all the empty embeddables are working as expected.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,411 @@
...
+                                                "nested",
+                                                new BsonDocument()
+                                                        .append("a", new BsonInt32(4))
+                                                        .append("b", new BsonInt32(5)))));
+        var loadedItem =
+                sessionFactoryScope.fromTransaction(session -> session.find(ItemWithNestedValues.class, item.id));
+        assertEquals(item, loadedItem);
+        var updatedItem = sessionFactoryScope.fromTransaction(session -> {
+            var result = session.find(ItemWithNestedValues.class, item.id);
+            result.nested1.a = -result.nested1.a;
```
katcharov, line 160: Let's modify all possible values? For example, at least:
```
            result.nested1.a = -result.nested1.a;
            result.nested2.nested = new StructAggregateEmbeddablePairValue2(-4, -5);
```

I am unsure of how recursion works, so perhaps there is some way to miss deeply nested changes.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,411 @@
...
+                                new BsonDocument()
+                                        .append("a", new BsonInt32(3))
+                                        .append(
+                                                "nested",
+                                                new BsonDocument()
+                                                        .append("a", new BsonInt32(4))
+                                                        .append("b", new BsonInt32(5)))));
+        var loadedItem =
+                sessionFactoryScope.fromTransaction(session -> session.find(ItemWithNestedValues.class, item.id));
+        assertEquals(item, loadedItem);
```
katcharov, line 157: I think we should rename this to something like `assertSameValues` - it is not at all equals, but seems to be something much more complex, involving reflection (though it is true that it would ultimately behave like equals under certain circumstances).

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
katcharov, line 1: Could we add tests for a self-referential class, like `Node { Node child; String value; }`?

katcharov (changes requested):

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
NathanQingyangXu: could we remove those `emptyValue` testing cases for they are working as expected? If we retain them, at least we'd better add more accurate comments not to mislead code readers into thinking that they are Hibernate ORM's bug.

For instance, the following comment:
```
// Hibernate ORM does not store `item.omitted` despite it being non-`null`
```
at least could be changed to
```
// By default Hibernate ORM does not store `item.omitted` despite it being non-`null`
```

stIncMale: This configuration property is deprecated, and the feature it controls was removed in Hibernate ORM 7.0:
- https://hibernate.atlassian.net/browse/HHH-11936
- https://lists.jboss.org/archives/list/hibernate-issues@lists.jboss.org/message/6LEDPQL2UI7HPPIQHLQV4UMB2MDGBWEC/

The current behavior is what Hibernate ORM considers the right one, i.e., it's not considered to be a bug.

NathanQingyangXu: but given currently we are still using v6.6, could we at least mentioned such detail in comment (like since v7, ....) to avoid confusion and be more accurate?

stIncMale: Done in fe6190f7bf71108d2b7aeef16b23c7607c4e1c5a. I changed the wording and linked to https://hibernate.atlassian.net/browse/HHH-11936, in case the reader needs more details. I don't want to link to the 6.6 documentation in this case, because if we change the links after upgrading to 7.0, the link won't work. And if we don't change the links, then linking to the documentation of 6.6 from a product that requires 7.0 is going to be weird.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,413 @@
...
+        sessionFactoryScope.inTransaction(session -> session.persist(item));
+        assertThat(mongoCollection.find())
+                .containsExactly(
+                        // Hibernate ORM does not store/read the empty `item.omitted` value.
+                        // See https://hibernate.atlassian.net/browse/HHH-11936 for more details.
+                        new BsonDocument(ID_FIELD_NAME, new BsonInt32(item.id)));
+        var loadedItem =
+                sessionFactoryScope.fromTransaction(session -> session.find(ItemWithFattenedEmptyValue.class, item.id));
+        // the entity we stored and the entity we loaded are not equal because Hibernate ORM omits `item.omitted`
+        assertNotEquals(item, loadedItem);
```
NathanQingyangXu, line 116: The current usage of `aspectj` is fine, though not idematic.
If only one field is the blocker to equalness, usually assertj's `ignoringFields()` method is used (see https://www.baeldung.com/assertj-ignore-fields-comparison). That way, we convey the two assertion intentions in one go:

1. one field is different
2. all the other fields are equals

which is a little bit better than `assertNotEquals`. Furthermore, we don't need to explicitly set the field to null simply to be able to run `assertEquals`.

Optional; only for discussion purpose.

stIncMale: Thank you. I changed this in e138a818c97ad390a75e2b77c4f2889935ff6950.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,415 @@
...
+        assertThat(mongoCollection.find())
+                .containsExactly(
+                        // Hibernate ORM does not store/read the empty `item.omitted` value.
+                        // See https://hibernate.atlassian.net/browse/HHH-11936 for more details.
+                        new BsonDocument(ID_FIELD_NAME, new BsonInt32(item.id)));
+        var loadedItem =
+                sessionFactoryScope.fromTransaction(session -> session.find(ItemWithFattenedEmptyValue.class, item.id));
+        assertUsingRecursiveComparison(item, loadedItem, (assertion, actual) -> assertion
+                .ignoringFields("omitted")
+                .isEqualTo(actual));
```
stIncMale, line 117: The `actual` parameter, despite always being the same as `loadedItem`, is needed to work around the effectively `final` requirement for a variable to be allowed to be put in a lambda expression. The Java SE API uses the same approach in some places, for example in `Map.merge`.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: We have all of the values "modified" when we insert. The only reason I test update is to have only some of them "modified".

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: The current name is intentionally chosen to be what it is. All this method does is saves us from overriding `equals` in the test code.

As for "same" when talking about reference types, I prefer using it only to express [reference equality](https://docs.oracle.com/javase/specs/jls/se24/html/jls-15.html#jls-15.21.3). The Java SE specification attempts to have the distinction between "same" and "equal" in some places, while failing to have it in others 🤷‍♂️

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: I did not mean it won't work for these specific tests, I meant the approach can't work generally speaking. This approach is also inconvenient in situations when the primary key is generated.  Which is why regardless of how tests are written in `BasicCrudIntegrationTests`, we do have tests written with `BsonDocument` API.

Done in f4e5391bbab07e2972836c09e548d07dd7d79774.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,411 @@
...
+        })
+@ExtendWith(MongoExtension.class)
+class EmbeddableIntegrationTests implements SessionFactoryScopeAware {
+    @InjectMongoCollection("items")
+    private static MongoCollection<BsonDocument> mongoCollection;
+
+    private SessionFactoryScope sessionFactoryScope;
+
+    @Test
+    void testFlattenedValues() {
```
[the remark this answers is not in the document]
stIncMale: Done in 12f3bf6b81f24870ec5a2b4f0e8cb5522700db1f.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,411 @@
...
+        {
+            item.id = new EmbeddableValue();
+            item.id.a = 1;
+            item.flattened1 = new EmbeddableValue();
+            item.flattened1.a = 2;
+            item.flattened2 = new EmbeddablePairValue1();
+            item.flattened2.a = 3;
+            item.flattened2.flattened = new EmbeddablePairValue2(4, 5);
+            item.flattened2.parent = item;
+        }
```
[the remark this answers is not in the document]
stIncMale: > EmbeddablePairValue1, which is not a pair but a value and a pair

"a value and a pair" is still a pair of two values. It does not matter what the type of those values is. It is, though, no longer a pair, because now it also has the `parent` field.

> "book", "author", etc.

I find those names very unhelpful, as they convey nothing useful in the context of tests, and are meant for the business logic, which does not even exist.

I did rename the types in 7091bb44f51756370f774d6e93f2d1da87ce10a0, hopefully that makes the naming better.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: That does not work in Hibernate ORM even for `@Embeddable` values that are flattened, and work without our intervention.

NathanQingyangXu (approved):

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -0,0 +1,191 @@
...
+        } else if (value instanceof BigDecimal v) {
+            return toBsonValue(v);
+        } else if (value instanceof String v) {
+            return toBsonValue(v);
+        } else if (value instanceof byte[] v) {
+            return toBsonValue(v);
+        } else if (value instanceof ObjectId v) {
+            return toBsonValue(v);
+        } else {
+            throw new SQLFeatureNotSupportedException(format(
```
stIncMale, line 65: Methods from `ValueConversions` are used in `MongoResultSet` and `MongoPreparedStatements`, as well as similar other places, where we convert domain values to JDBC values. When we call `MongoResultSet.get...`, and it fails because we don't support a specific type, the exception should be `SQLFeatureNotSupportedException`, not `FeatureNotSupportedException`.

The rest of the changes in f89c344e04d4eab453aac3eb6e8804b7ca236a40 is caused by this change.

vbabanin:
LGTM overall! Just a couple of minor suggestions.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
NathanQingyangXu: I think Hibernate might support such self-referential case but it is based on `@ManyToOne` or `@OneToOne` which requires table joining feature in m4; for now, we can't support entity type field and only focus on basic non-entity types our JDBC layer supports.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
NathanQingyangXu: One caveat is this `ValueConversions` ia also used outside of JDBC context, e.g. in `AbstractMqlAstTranslator` we also reuse it to transfer java value to AST literals. It would be great to move JDBC-specific concern out of this generic class (we could do that by refactoring our JDBC classes easily) so it could be resued universally.

Optional though.

src/main/java/com/mongodb/hibernate/internal/translate/ModelMutationMqlTranslator.java
```diff
@@ -40,7 +41,13 @@ public O translate(@Nullable JdbcParameterBindings jdbcParameterBindings, QueryO
         assertNull(jdbcParameterBindings);
         checkQueryOptionsSupportability(queryOptions);
 
-        var mutationCommand = acceptAndYield(tableMutation, COLLECTION_MUTATION);
-        return tableMutation.createMutationOperation(renderMongoAstNode(mutationCommand), getParameterBinders());
+        String sql;
```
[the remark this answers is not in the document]
NathanQingyangXu: I agree. `mql` better conveys our intention.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
NathanQingyangXu: `ValueConversions` is a little bit vague for value converions are everywhere even in Hibernate code logic. Not sure about how to better convey its intention. The salient feature here seems to focus on `BsonValue` (conversion source or target), so to me even `BsonValueConversions` is a little bit better, though it might mislead people to anticipate it is for conversion between two `BsonValue`s.

Discussion only; no need to change.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -0,0 +1,191 @@
...
+import org.bson.BsonValue;
+import org.bson.types.Decimal128;
+import org.bson.types.ObjectId;
+
+/**
+ * Provides conversion methods between {@link BsonValue}s, which our {@link PreparedStatement}/{@link ResultSet}
+ * implementation uses under the hood and rarely exposes, and domain values we usually use when setting parameter values
+ * on our {@link PreparedStatement}, or retrieving column values from a {@link ResultSet}.
+ */
+public final class ValueConversions {
```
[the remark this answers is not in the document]
stIncMale: Done in e72228bf7e0ddda5a2d36b67f3e0a5f186f249e7.

src/main/java/com/mongodb/hibernate/internal/translate/ModelMutationMqlTranslator.java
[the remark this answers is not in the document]
stIncMale: Done in e72228bf7e0ddda5a2d36b67f3e0a5f186f249e7.

NathanQingyangXu (dismissed):

src/main/java/com/mongodb/hibernate/internal/type/ValueConverter.java
```diff
@@ -40,8 +40,8 @@
  * implementation uses under the hood and rarely exposes, and domain values we usually use when setting parameter values
  * on our {@link PreparedStatement}, or retrieving column values from a {@link ResultSet}.
  */
-public final class ValueConversions {
-    private ValueConversions() {}
+public final class ValueConverter {
```
NathanQingyangXu, line 43: I prefer the original `ValueConversions` for it is a util class with only static methods. `ValueConverter` implies it is a non-util class and requires creating an object to go about converions.

but anyway, I am not picky and what really matters is it is not hard to understand its purpose.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
stIncMale: > It would be great to move JDBC-specific concern out of this generic class (we could do that by refactoring our JDBC classes easily)

The methods of `ValueConverter` are currently called from methods that are declared as `throws SQLException`, and from methods that are not. We have two approaches:

1. Throw `FeatureNotSupportedException`, then in methods declared as `throws SQLException`, catch it and wrap in `SQLFeatureNotSupportedException`.
2. Throw `SQLFeatureNotSupportedException`, then in methods not declared as `throws SQLException`, catch it and wrap in `FeatureNotSupportedException`.

If we use the second approach, which is currently used in this PR, we are forced by the compiler to follow it through. If we use the first approach, there is nothing but our conscience and attentiveness that can make us to follow it through. I think, we'll have better luck with the compiler forcing us, which is why I chose the current approach.

Another alternative is to have otherwise equivalent methods in `ValueConverter` such that half of them throws `FeatureNotSupportedException`, and another half throws `SQLFeatureNotSupportedException`. However, once we have them, it will again be only our conscience and attentiveness that can make us to void using the ones that throw `FeatureNotSupportedException` in contexts where `SQLException` can and should be thrown. And that is why I don't think this approach is helpful.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
NathanQingyangXu: Sorry, I misunderstood. We are talking about self-referencing another `@Embeddable`, not `@Entity`, so my above comment doesn't apply at all.
It might be possible to do that in Hibernate ORM, but not sure.

NathanQingyangXu: I tested Hibernate ORM by running the following testing case connecting with local Postgres:

```
@SessionFactory
@DomainModel(annotatedClasses = {
        WriteAndThrow.Book.class,
        WriteAndThrow.Comment.class
})
class WriteAndThrow {

    @BeforeEach
    void setUp(SessionFactoryScope scope) {
        scope.inTransaction(session -> {
            for (int i = 0; i < 5; i++) {
                var book = new Book(1 + i, "Book " + i);
                session.persist(book);
            }
        });
    }

    @Test
    void test(SessionFactoryScope scope) {
        scope.inTransaction(session -> {
            session.createSelectionQuery("select b from Book b", Book.class)
                    .getResultList()
                    .forEach(System.out::println);
        });
    }

    @Entity(name = "Book")
    @Table(name = "books")
    static class Book {
        @Id
        int myId;
        Comment comment;

        Book() {}
        Book(int myId, String text) {
            this.myId = myId;
            this.comment = new Comment();
            this.comment.text = text;
            this.comment.selfReference = this.comment;
        }

        @Override
        public String toString() {
            return "Book{" +
                    "myId=" + myId +
                    ", comment=" + comment +
                    '}';
        }
    }

    @Embeddable
    @Struct(name = "Comment")
    static class Comment {
        String text;
        Comment selfReference;
    }
}
```
Note the `Component` STRUCT contains `selfReference` poiniting to itself. Unfortunately, Hibernate ORM is not robust for this edge case for it shows the following error:

```
java.lang.StackOverflowError
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
	at org.hibernate.boot.model.internal.AbstractPropertyHolder.getExactOverriddenColumn(AbstractPropertyHolder.java:257)
```
So obviously currently Hibiernate ORM's STRUCT lacks circular fetching protection as in entity case, so we can do almost nothing here but report this defect to Hibernate team (I'll do this later).

@katcharov , I think we can do nothing here if Hibernate doesn't support this `self-referencing` STRUCT feature.

stIncMale: I believe, you are talking about self-referencing entities, not embeddables, which allow modelling schemas where the referencing and the referenced table in a foreign key constraint is the same table (or conceptually similar situations, despite them not involving the foreign key constraints). But self-referencing embeddables, which is what @katcharov asked for, are just not a thing in Hibernate ORM, as far as I can see. Am I wrong?

NathanQingyangXu: The STRUCT is SQL's UDT type, which might not allow embedding itself as its field. So Hibernate ORM might be defective only in the sense that it should error more elegantly.

Suppose we allow circular reference in an embedded doc (by emulating SQL's STRUCT feature in Hibernate so we have to live with its limitation), it is still an open question how to do that, right? We can't store itself by recursively embedding, right? So it seems we can skip this concern for now. If Hibernate does allow STRUCT self-referencing, we might need to store in other way (e.g. doc reference?). That defintively goes beyond this milestone or even the scope of the initial release. It is fine to create a ticket to keep track if we confirmed that Hibernate ORM supports this feature.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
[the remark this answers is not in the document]
stIncMale: @NathanQingyangXu's response: https://github.com/mongodb/mongo-hibernate/pull/89#discussion_r2081714669.

src/main/java/com/mongodb/hibernate/internal/type/ValueConverter.java
stIncMale: I believe, this response belongs to https://github.com/mongodb/mongo-hibernate/pull/89#discussion_r2079119461.

vbabanin (dismissed):

vbabanin:
LGTM! Reviewed as a secondary reviewer within the allocated timebox.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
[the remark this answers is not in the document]
NathanQingyangXu: I prefer the plural form for it is consistent with our existing naming patterns in this repo, e.g.:

- MongoAssertions
- MongoChecks

but I am fine with `ValueConverter` without problem. There is some discussion in https://stackoverflow.com/questions/2676704/naming-convention-for-utility-classes-in-java, and some people prefer plural forms as well.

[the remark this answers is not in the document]
NathanQingyangXu: JDK seems to follow the same pattern. E.g. `Collections`, with the exception that if the static method is factory method, singular form is used, e.g. `List.of(...)`.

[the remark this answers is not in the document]
stIncMale: I did not care much previously, but I like Nathan's arguments. @vbabanin, what do you think?

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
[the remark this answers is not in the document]
katcharov: Let's introduce constructors, and consider calling it Pair.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
katcharov: This looks fine to me, thanks for the explanations.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
katcharov: `assertEq`?

katcharov (approved):
Couple of minor comments above. LGTM.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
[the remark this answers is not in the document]
stIncMale: Done in 166e517c797a1024c360f8bbd4639adf9bf3ce90, 956cc1a7e9d6e7a62b0ed85058ccbe7d68846f2f.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
stIncMale: Done in 09d49217b2530d41463e15c8050e25a3e4bd93aa.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -0,0 +1,218 @@
...
+    @Override
+    public EmbeddableMappingType getEmbeddableMappingType() {
+        return assertNotNull(embeddableMappingType);
+    }
+
+    @Override
+    public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options) throws SQLException {
+        var embeddableMappingType = assertNotNull(this.embeddableMappingType);
+        if (embeddableMappingType.isPolymorphic()) {
+            throw new FeatureNotSupportedException("Polymorphic mapping is not supported");
```
NathanQingyangXu, line 96: Do we plan to support it in the future? Seems the decision is not based on "there is no MongoDB counterpart, so we decided to drop the support tottally). If yes, should we need to create a ticket to avoid loosing track of it? Ideally we also could create testing case for such `FeatureNotSupportedException` as @vbabanin emphasized (and I adopted since his participation of Hibernate PR review).

We did create a related ticket at https://jira.mongodb.org/browse/HIBERNATE-69, though not sure whether we need to reuse it or create a new one.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+        var result = new BsonDocument();
+        var jdbcValueCount = embeddableMappingType.getJdbcValueCount();
+        for (int columnIndex = 0; columnIndex < jdbcValueCount; columnIndex++) {
+            var jdbcValueSelectable = embeddableMappingType.getJdbcValueSelectable(columnIndex);
+            assertFalse(jdbcValueSelectable.isFormula());
```
NathanQingyangXu, line 102: I think another important edge case is the `toOneAttributeMapping`(see https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/metamodel/mapping/EmbeddableMappingType.java#L166) similar to the above `DiscriminatedAssociationAttributeMapping` which we have covered by exception throwing. We might need to do the same thing here to guard against a totally confusing and misleading excepttion at run-time

I create a testing case below for your reference, which succeeds in Hibernate ORM + PG, but it fails on our project with misleading exception deeply within our MQL translator code.

```
@DomainModel(
        annotatedClasses = {
            StructEdgeCaseTests.StructHolder.class,
            StructEdgeCaseTests.Comment.class,
            StructEdgeCaseTests.User.class
        })
class StructEdgeCaseTests {

    @Test
    void test(SessionFactoryScope scope) {
        User user = new User();
        user.id = 1;
        user.name = "John Smith";
        scope.inTransaction(session -> session.persist(user));

        var structHolder = new StructHolder();
        structHolder.id = 1;
        Comment comment = new Comment();
        comment.content = "it sucks";
        comment.user = user;
        structHolder.comment = comment;
        scope.inTransaction(session -> session.persist(structHolder));

        var fetchedStructHolder = scope.fromTransaction(session -> session.get(StructHolder.class, 1));
        MongoTestAssertions.assertEquals(structHolder, fetchedStructHolder);
    }

    @Entity
    @Table(name = "structHolder")
    static class StructHolder {
        @Id int id;
        Comment comment;
    }

    @Embeddable
    @Struct(name = "comment")
    static class Comment {
        String content;
        @ManyToOne
        User user;
    }

    @Entity
    @Table(name = "users")
    static class User {
        @Id int id;
        String name;
    }
}
```

We do have a ticket for simple `ToOne` translation at https://jira.mongodb.org/browse/HIBERNATE-65, but again, I am not sure whether we could reuse it. But at least we should avoid losing track of this important feature, similar to the above polymorphism attribute mapping case.

NathanQingyangXu (changes requested):

katcharov (approved):
Couple of minor comments above are resolved. LGTM.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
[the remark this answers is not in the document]
NathanQingyangXu: If we agree on plural form, the original `ValueConversions` is preferable to `ValueConverts`, which implies its static method(s) will work on some `ValueConvert` object as static method parameter.

src/integrationTest/java/com/mongodb/hibernate/EmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,180 @@
...
+        item.id = new StructAggregateEmbeddableValue();
+        item.id.a = 1;
+        item.nested1 = new StructAggregateEmbeddableValue();
+        item.nested1.a = 2;
+        item.nested2 = new StructAggregateEmbeddablePairValue1(3, new StructAggregateEmbeddablePairValue2(4, 5));
+        sessionFactoryScope.inTransaction(session -> session.persist(item));
+        assertThat(mongoCollection.find())
+                .containsExactly(new BsonDocument()
+                        // Hibernate ORM flattens `item.id` despite it being of an aggregate type
+                        .append(ID_FIELD_NAME, new BsonInt32(item.id.a))
```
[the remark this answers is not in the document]
stIncMale: The discussion is here [#hibernate-orm-dev > Using STRUCT as id](https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/Using.20STRUCT.20as.20id/with/514429822).

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: I included embeddables in [HIBERNATE-69](https://jira.mongodb.org/browse/HIBERNATE-69).

Done in 48d5171c19d805b0edbd69c781f91cc04f97ec67.

NathanQingyangXu (approved):

vbabanin (approved):

PR #90, opened 2025-04-23
sorting MQL translation

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-68

Since last PR, null value supporting was disabled, so this PR excludes null value sorting out of the scope.

Note that MQL only supports sorting specification based on field path, not ad-hoc function invocation result on field as the following HQL:

```
from Book order by length(title)
```
However, by tapping into `$addField` (or `$set` as its alias) aggregate stages, we could emulate such HQL. A separate ticket was created at https://jira.mongodb.org/browse/HIBERNATE-79 to track this unfinished feature (not complete solution for it depends on whether the SQL function has its MQL operator counterpart).

HQL's sorting doc is at https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#hql-order-by. 

This PR focuses on basic sorting semantic; the next followup ticket will tackle the `Limits/Offsets` (https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#hql-limit-offset).

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
+++ b/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
@@ -855,7 +855,7 @@ abstract class AbstractMqlTranslator<T extends JdbcOperation> implements SqlAstT
 
     static String renderMongoAstNode(AstNode rootAstNode) {
         try (var stringWriter = new StringWriter();
-                var jsonWriter = new JsonWriter(stringWriter, JSON_WRITER_SETTINGS)) {
+                var jsonWriter = new JsonWriter(stringWriter, EXTENDED_JSON_WRITER_SETTINGS)) {
             rootAstNode.render(jsonWriter);
             jsonWriter.flush();
             return stringWriter.toString();
--- a/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortFieldTests.java
+++ b/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortFieldTests.java
@@ -16,7 +16,7 @@
 
 package com.mongodb.hibernate.internal.translate.mongoast.command.aggregate;
 
-import static com.mongodb.hibernate.internal.translate.mongoast.AstNodeAssertions.assertElementRender;
+import static com.mongodb.hibernate.internal.translate.mongoast.AstNodeAssertions.assertElementRendering;
 import static com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstSortOrder.ASC;
 
 import org.junit.jupiter.params.ParameterizedTest;
@@ -29,9 +29,9 @@ class AstSortFieldTests {
     void testRendering(AstSortOrder sortOrder) {
         var sortField = new AstSortField("field", sortOrder);
         var expectedJson = """
-                           {"field": %d}\
-                           """
+                {"field": {"$numberInt": "%d"}}\
+                """
                 .formatted(sortOrder == ASC ? 1 : -1);
-        assertElementRender(expectedJson, sortField);
+        assertElementRendering(expectedJson, sortField);
     }
 }
--- a/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortStageTests.java
+++ b/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortStageTests.java
@@ -16,7 +16,7 @@
 
 package com.mongodb.hibernate.internal.translate.mongoast.command.aggregate;
 
-import static com.mongodb.hibernate.internal.translate.mongoast.AstNodeAssertions.assertRender;
+import static com.mongodb.hibernate.internal.translate.mongoast.AstNodeAssertions.assertRendering;
 
 import java.util.List;
 import org.junit.jupiter.api.Test;
@@ -29,9 +29,10 @@ class AstSortStageTests {
         var astSortField2 = new AstSortField("field2", AstSortOrder.DESC);
         var astSortStage = new AstSortStage(List.of(astSortField1, astSortField2));
 
-        var expectedJson = """
-                {"$sort": {"field1": 1, "field2": -1}}\
+        var expectedJson =
+                """
+                {"$sort": {"field1": {"$numberInt": "1"}, "field2": {"$numberInt": "-1"}}}\
                 """;
-        assertRender(expectedJson, astSortStage);
+        assertRendering(expectedJson, astSortStage);
     }
 }


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -498,6 +548,11 @@ public <N extends Number> void visitUnparsedNumericLiteral(UnparsedNumericLitera
                 FIELD_VALUE, new AstLiteralValue(toBsonValue(unparsedNumericLiteral.getLiteralValue())));
     }
 
+    @Override
+    public void visitSqlSelectionExpression(SqlSelectionExpression sqlSelectionExpression) {
+        sqlSelectionExpression.getSelection().getExpression().accept(this);
+    }
+
```
NathanQingyangXu, line 556: The above visitor implementation deserves some explanation. It is used for the following HQL involving sorting using "select expression" as sort key name:
```
select b.title as title, b.publishYear as year from Book as b order by year desc, title asc
```
without implementing the visitor method, the above valid HQL won't work. The corresponding testing case is `SortingIntegrationTests#testSortFieldByAlias()`.

The implementation is inspired by `AbstractSQLTranslator` in Hibernate as below (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java#L7158):
```
	@Override
	public void visitSqlSelectionExpression(SqlSelectionExpression expression) {
		if ( dialect.supportsOrdinalSelectItemReference() ) {
			appendSql( expression.getSelection().getJdbcResultSetIndex() );
		}
		else {
			expression.getSelection().getExpression().accept( this );
		}
	}
```
the first logic branch is about using ordinal  reference as below:
```
select title, address from Contact order by 1 ASC, 2 DEC;
```
Given we have to translate to MQL which doesn't support native ordinal reference, so we have to use the second logic branch. Luckily regardless of whether oridnal preference is used or not, the current implementation works invariably.

I added a testing case (`SortingIntegrationTests#testSortFieldByOrdinalReference`) to prove that ordinal reference works without further ado (as long as we chose the second logic branch):

stIncMale (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +369,70 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
-        var stages = filter == null
-                ? List.of(new AstProjectStage(projectStageSpecifications))
-                : List.of(new AstMatchStage(filter), new AstProjectStage(projectStageSpecifications));
         astVisitorValueHolder.yield(COLLECTION_AGGREGATE, new AstAggregateCommand(collection, stages));
     }
 
+    private Optional<AstSortStage> getOptionalAstSortStage(QuerySpec querySpec) {
+        if (querySpec.hasSortSpecifications()) {
+
+            if (querySpec.getSortSpecifications().size() > SORT_KEY_MAX_NUM) {
```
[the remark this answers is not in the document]
NathanQingyangXu: removed. thanks.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +367,68 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
+        if (querySpec.hasSortSpecifications()) {
+            var sortFields = new ArrayList<AstSortField>(
+                    querySpec.getSortSpecifications().size());
+
+            var encounteredSortFieldPaths = new HashSet<String>();
+            for (SortSpecification sortSpecification : querySpec.getSortSpecifications()) {
+                if (!isFieldPathExpression(sortSpecification.getSortExpression())) {
+                    throw new FeatureNotSupportedException(
+                            format("%s does not support sort key not of field path type", MONGO_DBMS_NAME));
+                }
```
NathanQingyangXu, line 405: The above logic is to ensure sort key is denoting a field path, for Hibernate supports ad-hoc sort key expression, e.g.
```
from Book order by upper(b.title)
```

As in the case of find operator case, we need to narrow down the scope (it requires one field path expression and  one literal value or parameter expression; sorting requires only one field path expression).

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -867,7 +914,7 @@ private static AstComparisonFilterOperator getAstComparisonFilterOperator(Compar
     }
 
     private static boolean isFieldPathExpression(Expression expression) {
-        return expression instanceof ColumnReference || expression instanceof BasicValuedPathInterpretation;
+        return expression.getColumnReference() != null;
```
NathanQingyangXu, line 917: Seems `expression.getColumnReference()` denotes whether the expression in question corresponds to a column (or field path in our case). By default its value is null, and the only few child classes that return non-null value includes:
Screenshot 2025-04-24 at 4 42 37 PM

So as you can see, the previous `ColumnReference` and `BasicValuedPathInterpretation` all belong to the group.
In this new sorting scenario, the `SqlSelectionExpression` is required in case of using selection item alias as sort keys in ORDER BY (corresponding to the testing cases of both `testSortFieldByAlias()` and `testSortFieldByOrdinalReference()` in `SortingIntegrationTests`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +369,57 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
+                if (sortSpecification.getNullPrecedence() != null
+                        && sortSpecification.getNullPrecedence() != NullPrecedence.NONE) {
+                    throw new FeatureNotSupportedException(format(
+                            "%s does not support nulls precedence: NULLS %s",
+                            MONGO_DBMS_NAME, sortSpecification.getNullPrecedence()));
+                }
+                if (sortSpecification.isIgnoreCase()) {
+                    throw new FeatureNotSupportedException(
+                            format("%s does not support case-insensitive sort key", MONGO_DBMS_NAME));
+                }
```
NathanQingyangXu, line 415: the above `isIgnoreCase` is an edge case Gavin King introduced from v6 without HQL counterpart (HQL always ends up with `false` config); the only way to reproduce the `true` case is through Criteria. I can create one if needed.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +369,57 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
+            for (SortSpecification sortSpecification : querySpec.getSortSpecifications()) {
+                if (!isFieldPathExpression(sortSpecification.getSortExpression())) {
+                    throw new FeatureNotSupportedException(
+                            format("%s does not support sort key not of field path type", MONGO_DBMS_NAME));
+                }
+                if (sortSpecification.getNullPrecedence() != null
+                        && sortSpecification.getNullPrecedence() != NullPrecedence.NONE) {
+                    throw new FeatureNotSupportedException(format(
+                            "%s does not support nulls precedence: NULLS %s",
+                            MONGO_DBMS_NAME, sortSpecification.getNullPrecedence()));
```
NathanQingyangXu, line 410: MongoDB puts `null` to the top in ASC case (only after `MinKey `) as per https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#std-label-bson-types-comparison-order

However, there is no counterpart of HQL's `NULLS LAST` and I am not sure whether the `NULLS FIRST` corresponds to Mongo's behaviour.  My understanding is we could simply throw exception to avoid confusion.

Another alternative is to only throw exception when `NULLS LAST` is specified in HQL, provided `NULLS FIRST` is just what Mongo is doing. Not sure whether the precondition holds or not.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Currently, the syntax `select ... order by (column1, column2)` throws an exception `does not support sort key not of field path type.` In this case, `sortSpecification.getSortExpression()` returns a `SqlTuple`.

From what I see, Hibernate’s `AbstractSqlAstTranslator#visitSortSpecification` handles it by treating the tuple as comma-separated fields. Should we implement the same approach?

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingIntegrationTests.java
```diff
@@ -0,0 +1,164 @@
...
+        testCommandListener.clear();
+    }
+
+    @ParameterizedTest
+    @EnumSource(AstSortOrder.class)
+    void testOrderBySingleFieldWithoutTies(AstSortOrder sortOrder) {
+        assertSelectionQuery(
+                "from Book as b ORDER BY b.publishYear " + sortOrder,
+                Book.class,
+                null,
```
vbabanin, line 64: Cross-referencing from another PR that is currently in review: https://github.com/mongodb/mongo-hibernate/pull/91#discussion_r2069313045

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: A high-quality code review comment!
Similar HQL I've found is `select b from Book b order by 1`, which will be treated just as `select b from Book b order by (b.id, b.title, b.price ...)`.

Another edge case is that `SqlTuple` could contain nested `SqlTuple`, as `from Book order by (id, (title, price))`.

Do we need to cover `SqlTuple` case in this ticket? A good question.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: Seems mongodb's `collation` is not the same as `SortSpecification#isIgnoreCase()` in the sense that:

- `SortSpecification#ignoreCase()` is more finer-grained for it is targeting one single sort key, not the whole sort key set or sorting criterion.
- MongoDB's collation config in aggregate command seems on the whole aggregate command scope
- MongoDB supports case-insensitive indexing, but again, that is global config for one specific column; but ``SortSpecification#isIgnoreCase()` is run-time config so the same field could have different case-sensitiveness requirements for different Hibernate queries.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +369,57 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
         var projectStageSpecifications = acceptAndYield(querySpec.getSelectClause(), PROJECT_STAGE_SPECIFICATIONS);
+        stages.add(new AstProjectStage(projectStageSpecifications));
 
-        var stages = filter == null
-                ? List.of(new AstProjectStage(projectStageSpecifications))
-                : List.of(new AstMatchStage(filter), new AstProjectStage(projectStageSpecifications));
         astVisitorValueHolder.yield(COLLECTION_AGGREGATE, new AstAggregateCommand(collection, stages));
     }
 
+    private Optional<AstSortStage> getOptionalAstSortStage(QuerySpec querySpec) {
```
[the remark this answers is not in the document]
NathanQingyangXu: good point! Removed.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -360,28 +369,57 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
+            var filter = acceptAndYield(whereClauseRestrictions, FILTER);
+            stages.add(new AstMatchStage(filter));
+        }
+
+        // $sort stage
+        getOptionalAstSortStage(querySpec).ifPresent(stages::add);
 
+        // $project stage
         var projectStageSpecifications = acceptAndYield(querySpec.getSelectClause(), PROJECT_STAGE_SPECIFICATIONS);
+        stages.add(new AstProjectStage(projectStageSpecifications));
```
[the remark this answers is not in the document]
NathanQingyangXu: I adopted, but I think `whereStage` is unnatural for `where` is SQL thing and `Stage` is Mongo thing.
I chose the following method names:

- createMatchStage()
- createSortStage()

Let me know whether they are ok

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingIntegrationTests.java
NathanQingyangXu: I introduced an overl-loaded `assertSelectionQueryFailure()` method without query postprocessor and pushed.
Given the similar concern could be solved in the other PR, let us avoid duplicated changing in two contending PRs and centralize the concern in the other simple PR. That would be more manageable IMHO.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -498,6 +530,64 @@ public <N extends Number> void visitUnparsedNumericLiteral(UnparsedNumericLitera
...
+                    "%s does not support nulls precedence: NULLS %s",
+                    MONGO_DBMS_NAME, sortSpecification.getNullPrecedence()));
+        }
+        if (sortSpecification.isIgnoreCase()) {
+            throw new FeatureNotSupportedException(
+                    format("%s does not support case-insensitive sort key [%s]", MONGO_DBMS_NAME, sortSpecification));
+        }
+        var sortExpression = sortSpecification.getSortExpression();
+        final List<String> fieldPaths;
+        if (sortExpression instanceof SqlTuple sqlTuple) {
```
NathanQingyangXu, line 555: If we compare with the default Hibernate impelementation in `AbstractSqlAstTranslator` (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java#L4260C2-L4278C3):

```
	@Override
	public void visitSortSpecification(SortSpecification sortSpecification) {
		final Expression sortExpression = sortSpecification.getSortExpression();
		final Nulls nullPrecedence = sortSpecification.getNullPrecedence();
		final SortDirection sortOrder = sortSpecification.getSortOrder();
		final boolean ignoreCase = sortSpecification.isIgnoreCase();
		final SqlTuple sqlTuple = getSqlTuple( sortExpression );
		if ( sqlTuple != null ) {
			String separator = NO_SEPARATOR;
			for ( Expression expression : sqlTuple.getExpressions() ) {
				appendSql( separator );
				visitSortSpecification( expression, sortOrder, nullPrecedence, ignoreCase );
				separator = COMMA_SEPARATOR;
			}
		}
		else {
			visitSortSpecification( sortExpression, sortOrder, nullPrecedence, ignoreCase );
		}
	}
```

You can see it uses the following statement to detect `SqlTuple`:
```
SqlTupleContainer.getSqlTuple( sortExpression );
```
The reason is there are many `SqlTupleContainer` child classes (`SqlTuple` is one of them) as follows:
Screenshot 2025-05-02 at 1 45 51 PM

However, even though these expression types are valid tuples, they don't all qualify as the tuples allowable in MongoDB's sort keys, so I avoided intentionally the complexity and only focus on the simple `SqlTuple` class alone.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -498,6 +530,64 @@ public <N extends Number> void visitUnparsedNumericLiteral(UnparsedNumericLitera
...
+            if (expression instanceof SqlTuple) {
+                fieldPaths.addAll(acceptAndYield(expression, FIELD_PATHS));
+            } else {
+                if (!isFieldPathExpression(expression)) {
+                    throw new FeatureNotSupportedException();
+                }
+                fieldPaths.add(acceptAndYield(expression, FIELD_PATH));
+            }
+        }
+        astVisitorValueHolder.yield(FIELD_PATHS, fieldPaths);
```
NathanQingyangXu, line 588: compared with Hibernate's default `visitTuple` implementation, our implementation is less elegant for we need to return value and currently we only support field path tuple, so the following common value type tuple usage is not supported for now:
```
insert into books (id, title) values (1, "War and Peace"), (2, "Crime and Punishment"), ...
```

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingIntegrationTests.java
NathanQingyangXu: I made complete code changes in the other PR (https://github.com/mongodb/mongo-hibernate/pull/91#issuecomment-2848008711)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: I inspected how `AbstractSqlAstTranslator` is using the `isIgnoreCase` info for SQL translation purpose. Below is its logic (at https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java#L4310):
```
	protected void renderSortExpression(Expression sortExpression, boolean ignoreCase) {
		if ( ignoreCase ) {
			appendSql( dialect.getLowercaseFunction() );
			appendSql( OPEN_PARENTHESIS );
		}

		if ( inOverOrWithinGroupClause() || ignoreCase ) {
			resolveAliasedExpression( sortExpression ).accept( this );
		}
		else {
			sortExpression.accept( this );
		}

		if ( ignoreCase ) {
			appendSql( CLOSE_PARENTHESIS );
		}
	}
```
by default, the `dialect.getLowercaseFunction()` will return `lower` function name:
```
	public String getLowercaseFunction() {
		return "lower";
	}
```

So suppose two sort keys are involved, one is `title` and ther other is `address`. If the `title` is case insensitive, by default per the above logic, the following SQL would be generated during the translation:

```
... order by lower(title), address
```
Obviously MQL has no such function-based emulation of the case-insensitive sort key spec, so for that reason I think we'd better throw `FeatureNotSupportedException` in this edge case (which only manifests itself when the query is built via Criterion for there is no HQL counterpart in its BNF grammer: https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/antlr/org/hibernate/grammars/hql/HqlParser.g4).

NathanQingyangXu: I added a testing case (based on `Criteria` for HQL has no counterpart) to cover the logic at `SortingSelectQueryIntegrationTests#testCaseInsensitiveSortSpecNotSupported()`.

To avoid the common confusion, I deleted the message of `FeatureNotSupportedException`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: I refactored the tuple visiting logic to make it generic. Now it will return collection of `Expression`, so it will be solid to work in any scenario.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -512,6 +530,70 @@ public void visitBooleanExpressionPredicate(BooleanExpressionPredicate booleanEx
...
+                astSortFields.add(createAstSortField(expression, astSortOrder));
+            }
+            astVisitorValueHolder.yield(SORT_FIELDS, astSortFields);
+        }
+    }
+
+    private AstSortField createAstSortField(Expression sortExpression, AstSortOrder astSortOrder) {
+        if (!isFieldPathExpression(sortExpression)) {
+            throw new FeatureNotSupportedException(
+                    format("%s does not support sort key not of field path type", MONGO_DBMS_NAME));
```
vbabanin, line 578: As discussed offline, this could be achieved using `$addFields` with MQL operators - for example, converting a field to lowercase or calculating its length before the sort stage, then sorting on the derived field without projecting it. [HIBERNATE-79](https://jira.mongodb.org/browse/HIBERNATE-79) has been created to track this for future consideration, but it’s out of scope for this ticket and PR.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: As discussed offline, this might be achieved using `$addFields` with MQL operators - for example, converting a field to lowercase or calculating its length before the sort stage, then sorting on the derived field without projecting it. [HIBERNATE-79](https://jira.mongodb.org/browse/HIBERNATE-79) has been created to track this for future consideration, but it’s out of scope for this ticket and PR.

vbabanin: https://github.com/mongodb/mongo-hibernate/pull/90#discussion_r2074489615

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -362,28 +362,46 @@ public void visitQuerySpec(QuerySpec querySpec) {
...
-        var filter = whereClauseRestrictions == null || whereClauseRestrictions.isEmpty()
-                ? null
-                : acceptAndYield(whereClauseRestrictions, FILTER);
+        var stages = new ArrayList<AstStage>(3);
+
+        createMatchStage(querySpec).ifPresent(stages::add);
+
+        createSortStage(querySpec).ifPresent(stages::add);
 
         var projectStageSpecifications = acceptAndYield(querySpec.getSelectClause(), PROJECT_STAGE_SPECIFICATIONS);
```
vbabanin, line 377: [nit/optional] Since we created `createMatchStage` and `createSortStage`, we could create a separate one project stage.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,215 @@
...
+        @Test
+        void testCaseInsensitiveSortSpecNotSupported() {
+            getSessionFactoryScope().inTransaction(session -> {
+                var cb = getSessionFactoryScope().getSessionFactory().getCriteriaBuilder();
+                var criteria = cb.createQuery(Book.class);
+                var root = criteria.from(Book.class);
+                criteria.select(root);
+                criteria.orderBy(cb.sort(root.get("title"), SortDirection.ASCENDING, NullPrecedence.NONE, true));
+                assertThatThrownBy(() -> session.createSelectionQuery(criteria).getResultList())
+                        .isInstanceOf(FeatureNotSupportedException.class);
```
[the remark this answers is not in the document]
NathanQingyangXu: done

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: done

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,218 @@
...
+                        new Object[] {"Anna Karenina", 1877},
+                        new Object[] {"Crime and Punishment", 1866},
+                        new Object[] {"The Brothers Karamazov", 1880},
+                        new Object[] {"War and Peace", 2025},
+                        new Object[] {"War and Peace", 1869}));
+    }
+
+    @Nested
+    @DomainModel(annotatedClasses = Book.class)
+    @ServiceRegistry(settings = @Setting(name = DEFAULT_NULL_ORDERING, value = "first"))
```
NathanQingyangXu, line 147: This is an approach I found we could limit Hibernate property configuration in one testing method (the setting won't be applied to the whole class for it would break all the testing cases) using `hibernate-testing` and JUnit5's `@Nested` features.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Leaving this discussion unresolved for visibility, so other reviewers are aware of the offline conversation.

vbabanin (dismissed):

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,218 @@
...
+        assertSelectionQuery(
+                "from Book as b ORDER BY b.publishYear " + sortOrder,
+                Book.class,
+                "{ 'aggregate': 'books', 'pipeline': [ { '$sort': { 'publishYear': " + (sortOrder == ASC ? "1" : "-1")
+                        + " } }, {'$project': {'_id': true, 'discount': true, 'isbn13': true, 'outOfStock': true, 'price': true, 'publishYear': true, 'title': true} } ] }",
+                sortOrder == ASC ? getBooksByIds(2, 1, 3, 4, 5) : getBooksByIds(5, 4, 3, 1, 2));
+    }
+
+    @ParameterizedTest
+    @EnumSource(AstSortOrder.class)
```
vbabanin, line 78: [Optional] Since this is an integration test, we should generally avoid exposing internal implementation details, as class names or specifics might change - causing the test to fail even though functionality remains correct. I suggest using just `ASC` and `DESC` in `ValueSource`

NathanQingyangXu: Good point. I changed both methods which are using the internal enum class to switch to strings.

NathanQingyangXu:
@stIncMale could you take a new look at the PR? It was approved by the primary reviewer already; but its merging still requires secondary reviewer's approval.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,217 @@
...
+        getTestCommandListener().clear();
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"ASC", "DESC"})
+    void testOrderBySingleFieldWithoutTies(String sortDirection) {
+        assertSelectionQuery(
+                "from Book as b ORDER BY b.publishYear " + sortDirection,
+                Book.class,
+                "{ 'aggregate': 'books', 'pipeline': [ { '$sort': { 'publishYear': "
```
katcharov, line 69: Let's use a nicely-formatted multi-line string for all of these.

Since you are using parameters, let's use `.formatted(` to set the parameter in the string (you do this in other tests).

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @BeforeEach
+    void beforeEach() {
+        getSessionFactoryScope().inTransaction(session -> testingBooks.forEach(session::persist));
+        getTestCommandListener().clear();
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"ASC", "DESC"})
```
katcharov, line 64: Maybe use `AstSortOrder`?

katcharov (changes requested):

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
NathanQingyangXu: good point! Committed at https://github.com/mongodb/mongo-hibernate/pull/90/commits/804216351d1df7432e399357a1d04f51f4d0e040

Will create a quick PR without ticket later to fix similar issue in existing `SimpleSelectQueryIntegrationTests`.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
NathanQingyangXu: That was my original usage, but @stIncMale told me that is internal stuff so we'd better avoid using it. see https://github.com/mongodb/mongo-hibernate/pull/90#discussion_r2076033023 for the previous solved comment.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
NathanQingyangXu: Given the refactoring in `SimpleSelectQueryIntegrationTests` is easy to verify, I included it at https://github.com/mongodb/mongo-hibernate/pull/90/commits/a2cef52ffdc86beb04099cdc2f48356ceab95a55

NathanQingyangXu: Note that another `palantir` formatting issue was flushed out (`.formatted()` will be formatted in a new line, not right after the ending triple quotes of the text block). I solved this format issue at another PR: https://github.com/mongodb/mongo-hibernate/pull/97

stIncMale (changes requested):

stIncMale:
The last reviewed commit is a2cef52ffdc86beb04099cdc2f48356ceab95a55.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -512,6 +532,69 @@ public void visitBooleanExpressionPredicate(BooleanExpressionPredicate booleanEx
         astVisitorValueHolder.yield(FILTER, filter);
     }
 
+    @Override
+    public void visitSqlSelectionExpression(SqlSelectionExpression sqlSelectionExpression) {
+        sqlSelectionExpression.getSelection().getExpression().accept(this);
+    }
+
+    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // ORDER BY clause
```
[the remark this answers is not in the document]
NathanQingyangXu: This is subjective comment. I copied such comments from Hibernate and I do think it helps. But if you persist, I have no objection remove them as well.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,371 @@
...
+    @Nested
+    class UnsupportedTests {
+        @Test
+        void testSortFieldNotFieldPathExpressionNotSupported() {
+            assertSelectQueryFailure(
+                    "from Book ORDER BY length(title)",
+                    Book.class,
+                    FeatureNotSupportedException.class,
+                    "TODO-HIBERNATE-%1$d https://jira.mongodb.org/browse/HIBERNATE-%1$d",
+                    79);
```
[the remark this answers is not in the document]
NathanQingyangXu: I used `%1$d` to avoid duplication. I agree it is overkill and I reverted back to `TODO-HIBERNATE-79 https://jira.mongodb.org/browse/HIBERNATE-79`

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortField.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast.command.aggregate;
+
+import com.mongodb.hibernate.internal.translate.mongoast.AstNode;
+import org.bson.BsonWriter;
+
+public record AstSortField(String path, AstSortOrder order) implements AstNode {
```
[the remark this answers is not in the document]
NathanQingyangXu: It is a naming convention in C# codebase. Similar convention applies to other stages like:

- `AstSetStage`: `AstComputedField`
- `AstLookupStage`: `foreignField`, `localField`

to me, there is no confusion to use `AstSortField` and I don't think we need to maintain consistency with `$project` stage.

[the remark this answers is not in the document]
NathanQingyangXu: From the doc, it seems only field is allowed; but I experimented in `Mongo playground` and found path is allowed.
Screenshot 2025-05-21 at 9 54 57 AM

https://mongoplayground.net/p/23v5wVmc3og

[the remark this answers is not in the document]
NathanQingyangXu: I think we definitively could simplify `AstFilterFieldPath` in favor of simple String. I guess the reason we introduced a wrapper is due to the C# code as below:
```
{
    internal sealed class AstFilterField : AstNode
    {
        private string _path;

        public AstFilterField(string path)
        {
            _path = Ensure.IsNotNull(path, nameof(path));
        }

        public string Path => _path;
        public override AstNodeType NodeType => AstNodeType.FilterField;

        public override AstNode Accept(AstNodeVisitor visitor)
        {
            return visitor.VisitFilterField(this);
        }

        public AstFilterField SubField(string subFieldName)
        {
            Ensure.IsNotNull(subFieldName, nameof(subFieldName));

            if (_path == "@<current>")
            {
                return new AstFilterField(subFieldName);
            }
            else
            {
                return new AstFilterField(_path + "." + subFieldName);
            }
        }

        public override BsonValue Render() => _path;
    }
}
```

If I were not wrong, the reason a wrapper was introduced is for the `SubField` method, which seems not totally mandatory. I think we can simplify without much risk. I'll create a quick PR focusing on this.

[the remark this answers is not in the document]
NathanQingyangXu: that being said, if you persist, I have no objection to rename it to `AstSortStageSpecificaiton`. my concern is `$project` might be special so adhering its naming convention prematurely might not be desirable.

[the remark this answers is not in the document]
NathanQingyangXu: A simple PR created at https://github.com/mongodb/mongo-hibernate/pull/99

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
stIncMale: I agree that the first line was subjective, but both item 1 and 2 are objective.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstSortField.java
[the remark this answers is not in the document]
stIncMale: Got it, thank you for verifying!

stIncMale (approved):
The last reviewed commit is 9e9ca6ced584534e692f76fa16b7debcb78cce6b.

stIncMale (approved):
The last reviewed commit is 019d5c0b424a9a1f6715fea0241fbf6d095c7a1e.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
katcharov: I responded in 97. I don't think that we need to change the position of the `.formatted` call, but this is outside of the scope of this PR anyway.

katcharov (approved):
LGTM (my comments are resolved). Did not review full PR.

katcharov:
Semgrep found no issues (as of 7573ef2). Looks good.

PR #91, opened 2025-04-23
Implement visitBooleanExpressionPredicate

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-78

A straightforward ticket

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2025-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.query.select;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import com.mongodb.hibernate.TestCommandListener;
+import com.mongodb.hibernate.junit.MongoExtension;
+import java.util.List;
+import java.util.function.Consumer;
+import org.assertj.core.api.InstanceOfAssertFactories;
+import org.bson.BsonDocument;
+import org.hibernate.query.SelectionQuery;
+import org.hibernate.testing.orm.junit.ServiceRegistryScope;
+import org.hibernate.testing.orm.junit.ServiceRegistryScopeAware;
+import org.hibernate.testing.orm.junit.SessionFactory;
+import org.hibernate.testing.orm.junit.SessionFactoryScope;
+import org.hibernate.testing.orm.junit.SessionFactoryScopeAware;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@SessionFactory(exportSchema = false)
+@ExtendWith(MongoExtension.class)
+abstract class AbstractSelectionQueryIntegrationTests implements SessionFactoryScopeAware, ServiceRegistryScopeAware {
+
+    SessionFactoryScope sessionFactoryScope;
+
+    TestCommandListener testCommandListener;
+
+    @Override
+    public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
+        this.sessionFactoryScope = sessionFactoryScope;
+    }
+
+    @Override
+    public void injectServiceRegistryScope(ServiceRegistryScope serviceRegistryScope) {
+        this.testCommandListener = serviceRegistryScope.getRegistry().requireService(TestCommandListener.class);
+    }
+
+    <T> void assertSelectionQuery(
+            String hql,
+            Class<T> resultType,
+            Consumer<SelectionQuery<T>> queryPostProcessor,
+            String expectedMql,
+            List<T> expectedResultList) {
+        assertSelectionQuery(hql, resultType, queryPostProcessor, expectedMql, resultList -> assertThat(resultList)
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyElementsOf(expectedResultList));
+    }
+
+    <T> void assertSelectionQuery(
+            String hql,
+            Class<T> resultType,
+            Consumer<SelectionQuery<T>> queryPostProcessor,
+            String expectedMql,
+            Consumer<List<T>> resultListVerifier) {
+        sessionFactoryScope.inTransaction(session -> {
+            var selectionQuery = session.createSelectionQuery(hql, resultType);
+            if (queryPostProcessor != null) {
+                queryPostProcessor.accept(selectionQuery);
+            }
+            var resultList = selectionQuery.getResultList();
+
+            assertActualCommand(BsonDocument.parse(expectedMql));
+
+            resultListVerifier.accept(resultList);
+        });
+    }
+
+    <T> void assertSelectQueryFailure(
+            String hql,
+            Class<T> resultType,
+            Consumer<SelectionQuery<T>> queryPostProcessor,
+            Class<? extends Exception> expectedExceptionType,
+            String expectedExceptionMessage,
+            Object... expectedExceptionMessageParameters) {
+        sessionFactoryScope.inTransaction(session -> assertThatThrownBy(() -> {
+                    var selectionQuery = session.createSelectionQuery(hql, resultType);
+                    if (queryPostProcessor != null) {
+                        queryPostProcessor.accept(selectionQuery);
+                    }
+                    selectionQuery.getResultList();
+                })
+                .isInstanceOf(expectedExceptionType)
+                .hasMessage(expectedExceptionMessage, expectedExceptionMessageParameters));
+    }
+
+    void assertActualCommand(BsonDocument expectedCommand) {
+        var capturedCommands = testCommandListener.getStartedCommands();
+
+        assertThat(capturedCommands)
+                .singleElement()
+                .asInstanceOf(InstanceOfAssertFactories.MAP)
+                .containsAllEntriesOf(expectedCommand);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionPredicateTranslatingTests.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@DomainModel(annotatedClasses = Book.class)
+class BooleanExpressionPredicateTranslatingTests extends AbstractSelectionQueryIntegrationTests {
```
vbabanin, line 29: `BooleanExpressionPredicateTranslatingTests` doesn’t feel like the right name here. Since it extends `AbstractSelectionQueryIntegrationTests`, it’s effectively an integration test focused on selection queries. I suggest renaming it to something clearer, like `BooleanExpressionSelectQueryIntegrationTests`, to better reflect its purpose.

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
```diff
@@ -0,0 +1,110 @@
...
+import org.hibernate.testing.orm.junit.SessionFactory;
+import org.hibernate.testing.orm.junit.SessionFactoryScope;
+import org.hibernate.testing.orm.junit.SessionFactoryScopeAware;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@SessionFactory(exportSchema = false)
+@ExtendWith(MongoExtension.class)
+abstract class AbstractSelectionQueryIntegrationTests implements SessionFactoryScopeAware, ServiceRegistryScopeAware {
+
+    SessionFactoryScope sessionFactoryScope;
```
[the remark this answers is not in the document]
stIncMale: `protected` is a more permissive than package-access, and making program elements more accessible than needed is something that we agreed not to do. Does this field and `testCommandListener` have to be `protected`?

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
```diff
@@ -0,0 +1,110 @@
...
+
+    TestCommandListener testCommandListener;
```
stIncMale, line 42: These fields are writeable by code outside of the `AbstractSelectionQueryIntegrationTests` class, but must never be written by such code. We can't solve this issue by making them `final`, but we can solve it by making them `private` and exposing package-access getter method for reading the fields. Let's do that, and not worry about the fields being modified outside of `AbstractSelectionQueryIntegrationTests`.

stIncMale, line 42: These fields are writeable by code outside of the `AbstractSelectionQueryIntegrationTests` class, but must never be written by such code. We can't solve this issue by making them `final`, but we can solve it by making them `private` and exposing package-access getter method for reading the fields. Let's do that, and not worry about the fields being modified outside of `AbstractSelectionQueryIntegrationTests`.

stIncMale (changes requested):

src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionPredicateTranslatingTests.java
```diff
@@ -0,0 +1,75 @@
...
+                        + "}}}, {'$project': {'_id': true, 'discount': true, 'isbn13': true, 'outOfStock': true, 'price': true, 'publishYear': true, 'title': true}}]}",
+                negated ? singletonList(bookInStock) : singletonList(bookOutOfStock));
+    }
+
+    @Test
+    void testNonFieldPathExpressionNotSupported() {
+        assertSelectQueryFailure(
+                "from Book where true",
+                Book.class,
+                null,
```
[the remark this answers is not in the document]
NathanQingyangXu: adopted

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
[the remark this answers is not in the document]
NathanQingyangXu: well, from visibility perspective, `protected` is not mandatory; but it is also a good practice to use `protected` on abstract class to denote it is accessible to any child class, regardless of whether the child class reside in the same package or not. it is more about semantic connotation.

For that reason, I want to follow @vbabanin's suggestions.

src/integrationTest/java/com/mongodb/hibernate/query/select/Book.java
```diff
@@ -16,26 +16,23 @@
...
 import java.math.BigDecimal;
-import org.bson.types.ObjectId;
 
 @Entity(name = "Book")
 @Table(name = "books")
 public class Book {
     @Id
-    @ObjectIdGenerator
-    ObjectId id;
+    int id;
```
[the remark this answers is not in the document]
NathanQingyangXu: This doesn't need to be changed for we don't rely on `id` to go about query result validation (as in ther other [PR](https://github.com/mongodb/mongo-hibernate/pull/90)). I reverted the change back.

src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionPredicateTranslatingTests.java
NathanQingyangXu: Given the package name has been `com.mongodb.hibernate.query.select`, I would claim that no need to include `selectQuery` in the testing classe names. I included it to the first `SimpleSelectQueryIntegrationTests` because it is too simple not to add something to its name.

That being said, I have no objection to rename it. However, `BooleanExpressionSelectQueryIntegrationTests` might be misleading either, because we only use `BooleanExpression` in where clause (not in select clause, for instance), so I renamed it to `BooleanExpressionWhereClauseIntegrationTests` instead (I had to omit `selectQuery` for the class name has been long enough and the containing package has made it clear that we are not working on DML queries like insertion, updating and deletion).

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
NathanQingyangXu: done

NathanQingyangXu:
@vbabanin , I completed the assertion overloaded methods without null query post processor at b420a2d00b0409481e73218b99356c0110b27809

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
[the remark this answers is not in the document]
vbabanin: I initially assumed this class might be reused across packages. However, if it’s meant to be used only within the same package, then `package-private` should be more appropriate based on the earlier discussions.

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+            String expectedMql,
+            List<T> expectedResultList) {
+        assertSelectionQuery(hql, resultType, queryPostProcessor, expectedMql, resultList -> assertThat(resultList)
+                .usingRecursiveFieldByFieldElementComparator()
+                .containsExactlyElementsOf(expectedResultList));
+    }
+
+    <T> void assertSelectionQuery(String hql, Class<T> resultType, String expectedMql, List<T> expectedResultList) {
+        assertSelectionQuery(hql, resultType, null, expectedMql, expectedResultList);
+    }
```
vbabanin, line 75: I suggest format the parameters to match the pattern used in the rest of the class for consistency.

```suggestion
    <T> void assertSelectionQuery(
            String hql, 
            Class<T> resultType, 
            String expectedMql, 
            List<T> expectedResultList) {
        assertSelectionQuery(hql, resultType, null, expectedMql, expectedResultList);
    }
```

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+            assertActualCommand(BsonDocument.parse(expectedMql));
+
+            resultListVerifier.accept(resultList);
+        });
+    }
+
+    <T> void assertSelectionQuery(
+            String hql, Class<T> resultType, String expectedMql, Consumer<List<T>> resultListVerifier) {
+        assertSelectionQuery(hql, resultType, null, expectedMql, resultListVerifier);
+    }
```
vbabanin, line 99: I suggest format the parameters to match the pattern used in the rest of the class for consistency.

```suggestion
    <T> void assertSelectionQuery(
            String hql,
            Class<T> resultType,
            String expectedMql,
            Consumer<List<T>> resultListVerifier) {
        assertSelectionQuery(hql, resultType, null, expectedMql, resultListVerifier);
    }
```

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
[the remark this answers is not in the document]
NathanQingyangXu: I switched to `package-private` for the following two methods:
```
    SessionFactoryScope getSessionFactoryScope() {
        return sessionFactoryScope;
    }

    TestCommandListener getTestCommandListener() {
        return testCommandListener;
    }
```

src/integrationTest/java/com/mongodb/hibernate/query/select/AbstractSelectionQueryIntegrationTests.java
NathanQingyangXu: The formatting is done automatically by `spotlessApply`. There used to be a `static-check` bug (`spotlessApply` is applied automatically) which hides the issue. Now I updated the PR to include the evergreen bug fix, now the above two `Commit suggestion` would result in evergreen building failure.

Had to revert back to satisfy `spotless`.

vbabanin (dismissed):
LGTM!

vbabanin (approved):

stIncMale (approved):
Skimmed through, did not review.

PR #92, opened 2025-04-24
Replace `Session.get` calls with `Session.find` calls, because Hibernate ORM 7.0 deprecated `Session.get`

stIncMale:
https://docs.jboss.org/hibernate/orm/7.0/javadocs/org/hibernate/Session.html#get(java.lang.Class,java.lang.Object)

Where there is no Jakarta alternative, we'll keep using Hibernate-specific methods.


NathanQingyangXu (approved):

vbabanin (approved):
LGTM!

PR #93, opened 2025-05-02
remove cache plan cache forbidding in AbstractMqlTranslator

NathanQingyangXu:
This is a defect I found recently. Initially we went about `defensive programming` by throwing `FeatureNotSupportedException` for any meaningful configuration in `QueryOptions`, including `getQueryPlanCachingEnabled()` as below:

```
        if (Boolean.TRUE.equals(queryOptions.getQueryPlanCachingEnabled())) {
            throw new FeatureNotSupportedException("'queryPlanCaching' in QueryOptions not supported");
        }
```
However, this is insane for whether to enable query plan cache almost has nothing to do with MQL translating and it is mainly meant for upper level code to decide whether to cache or not, so unlike the other query options which are related to query features, we should not forbid the following usage by end-user:

```
var selectQuery = session.createSelectQuery("from Book where id = 1", Book.class);
selectQuery.setQueryPlanCacheable(true);
// get query result like selectQuery.getSingleResult();
```

If you search in Hibernate's default `AbstractSqlAstTranslator`, you can see `getQueryPlanCachingEnabled()` is never used in any way.
An overlooking from my previous PR. Still not too late to fix it though.

Added a testing case to cover our asses.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/SimpleSelectQueryIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/SimpleSelectQueryIntegrationTests.java
@@ -18,6 +18,7 @@ package com.mongodb.hibernate.query.select;
 
 import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.mongodb.hibernate.TestCommandListener;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/com/mongodb/hibernate/query/select/SimpleSelectQueryIntegrationTests.java
```diff
@@ -321,6 +322,15 @@ void testNullParameterNotSupported() {
                     FeatureNotSupportedException.class,
                     "TODO-HIBERNATE-74 https://jira.mongodb.org/browse/HIBERNATE-74");
         }
+
+        @Test
+        void testQueryPlanCacheIsNotForbidden() {
```
stIncMale, line 327: [optional]

Let's use "IsSupported" or "IsAllowed" instead of "IsNotForbidden": 

```suggestion
        void testQueryPlanCacheIsSupported() {
```

stIncMale (approved):
Approved with one optional suggestion.

The last reviewed commit is 08874f107bbb9f060de56e953b57110ce24f2e85.

PR #95, opened 2025-05-09
Support arrays/collections mapping

stIncMale:
This PR introduces tests for `null` values, including tests for empty `@Embeddable` values (ones that have `null` as the value of each of the persistence attributes), and the commit c4986fa8a5f4343a678e85b5740289bfea7dc558 has hacked-in implementation of `null` support to allow running those tests. You may identify such tests by searching for `@Disabled`.

HIBERNATE-58


src/integrationTest/java/com/mongodb/hibernate/CollectionIntegrationTests.java
```diff
@@ -0,0 +1,387 @@
...
+            var item = new ItemWithBoxedBytesArrayValue();
+            {
+                item.id = 1;
+                item.bytes = new byte[] {1};
+                item.boxedBytes = new Byte[] {2};
+            }
+            // this is Hibernate ORM bug, and it goes away if the `ItemWithBoxedBytesArrayValue.bytes` field is removed
+            assertThatThrownBy(() -> sessionFactoryScope.inTransaction(session -> session.persist(item)))
+                    .isInstanceOf(ClassCastException.class);
+        }
```
stIncMale, line 272: There is likely a bug in Hibernate ORM causing this behavior.

src/integrationTest/java/com/mongodb/hibernate/embeddable/StructAggregateEmbeddableIntegrationTests.java
```diff
@@ -0,0 +1,379 @@
...
+        @Id
+        int id;
+
+        MultiValueWithCollections nested;
+    }
+
+    @Embeddable
+    @Struct(name = "MultiValueWithArraysAndCollections")
+    static class MultiValueWithCollections {
+        Collection<Integer> intsCollection;
```
stIncMale, line 282: Replacing the type with `int[]` here does not work.

We extract the value of `intsCollection` from `ResultSet` for Hibernate ORM via `MongoStructJdbcType.Extractor.doExtract`. That boils down to calling `ValueConverter.toDomainValue(BsonValue)` and returning `List<?>`, whose runtime type is `ArrayList<Integer>`, so Hibernate ORM has no problem assigning it to `Collection<Integer> intsCollection`.

However, if we declare `int[] intsCollection`, the  Hibernate ORM does no extra converting work trying to assign `ArrayList<Integer>` to `int[] intsCollection`, and just fails.

Note, however, that this is not how Hibernate ORM behaves when we have array persistent attributes in entities (as opposed to having arrays as fields in embeddable values). There (see `CollectionIntegrationTests.testBasicArrayValues`/`testBasicCollectionValues`) Hibernate ORM extracts `MongoArray` from `ResultSet` by calling `MongoResultSet.getArray` on its own. And despite `MongoArray` containing `Object[]` (if the persistent attribute type is `int[]`/`Collection<Integer>`, then the runtime type of this array is `Integer[]`), Hibernate ORM manages to convert it further and successfully assign the value to a persistent attribute of the `int[]`/`Collection<Integer>` type.

P.S. Us returning `Object[]` from `ValueConverter.toDomainValue(BsonValue)` instead of `List<?>`, or even `MongoArray` does not solve the issue.

src/integrationTest/java/com/mongodb/hibernate/CollectionIntegrationTests.java
NathanQingyangXu: I digged a little bit. The root cause is `BasicTypeRegistry` registries both `byte[]` and `Byte[]` with value of the latter. If we use `legacy` as the `hibernate.type.wrapper_array_handling` property, the issue would be gone.

Not 100% sure but it seems this is as expected. We might need to know every details of the new property introduced since 6.2 to fully grasp.

NathanQingyangXu: The `org.hibernate.mapping.Property#isValid()` provides some helpful context to justify why they converge `byte[]` and `Byte[]` as below (https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/mapping/Property.java#L293):

Screenshot 2025-05-22 at 5 11 32 PM

So from my understanding, if `hibernate.type.wrapper_array_handling ` property is set to `allow`, `Byte[]` instead of `byte[]` is supposed to be used; otherwise `legacy` property might make more sense.

Still kind of defective but mixture of `byte[]` and `Byte[]` seems inconsistent with the `allow` configuration as well.

src/integrationTest/java/com/mongodb/hibernate/embeddable/StructAggregateEmbeddableIntegrationTests.java
stIncMale: I found a way to work around this issue.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
```diff
@@ -38,10 +39,11 @@ public String aggregateComponentCustomReadExpression(
...
             Column column) {
         var aggregateColumnType = aggregateColumn.getTypeCode();
-        if (aggregateColumnType == MongoStructJdbcType.JDBC_TYPE.getVendorTypeNumber()) {
+        if (aggregateColumnType == MongoStructJdbcType.JDBC_TYPE.getVendorTypeNumber()
+                || aggregateColumnType == MongoArrayJdbcType.HIBERNATE_SQL_TYPE) {
             return format(
-                    "unused from %s.aggregateComponentCustomReadExpression",
-                    MongoAggregateSupport.class.getSimpleName());
+                    "unused from %s.aggregateComponentCustomReadExpression for SQL type code [%d]",
+                    MongoAggregateSupport.class.getSimpleName(), aggregateColumnType);
```
stIncMale, line 46: We now know how these returned strings are used, and have a relevant ticket https://jira.mongodb.org/browse/HIBERNATE-93.

src/main/java/com/mongodb/hibernate/internal/type/MongoArrayJdbcType.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public int getJdbcTypeCode() {
+        var result = super.getJdbcTypeCode();
+        assertTrue(result == JDBC_TYPE.getVendorTypeNumber());
+        return result;
+    }
+
+    @Override
+    protected <X> @Nullable X getArray(
```
stIncMale, line 55: `MongoArrayJdbcType` exists solely to make this method accessible to our code.

src/main/java/com/mongodb/hibernate/internal/type/MongoBigDecimalJavaType.java
stIncMale, line 1: All the new `...JavaType` overrides, as well as the `...PrimitiveArrayJavaType` overrides were added to "teach" Hibernate ORM how to convert `BsonValue` to those basic types. Previously we got by without it because we were always converting them explicitly in the `ResultSet.get...` methods, but now that we started relying on Hibernate's `ArrayJdbcType`, it is needed.

src/integrationTest/java/com/mongodb/hibernate/CollectionIntegrationTests.java
stIncMale: @NathanQingyangXu, thank you for asking this question: https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/hibernate.2Etype.2Ewrapper_array_handling.20setting.20issue/with/525325554.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -132,26 +153,60 @@ public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options)
...
+                }
+                if (!(jdbcMapping.getJdbcValueExtractor() instanceof BasicExtractor<?> jdbcValueExtractor)) {
+                    throw fail();
+                }
+                domainValue = arrayJdbcType.getArray(
+                        jdbcValueExtractor,
+                        // `org.hibernate.type.descriptor.jdbc.ArrayJdbcType` expects `java.sql.Array.getArray`
+                        // to return `Object[]`, we cannot put a `BsonArray`, which is a `List<BsonValue>`, inside.
+                        new MongoArray(value.asArray().toArray()),
+                        options);
```
NathanQingyangXu, line 200: from my understanding, we can tap into your `ValueConventions#toArrayDomainValue()` to pass SQL Array with contents already of domain values. Then we don't need to introduce those basic Mongo JavaTyps in the first place.
E.g.
```
                domainValue = arrayJdbcType.getArray(
                        jdbcValueExtractor,
                        ValueConversions.toArrayDomainValue(value.asArray()),
                        options);
```
then your testing case would pass without further ado. The bridging between bason values and domain values has been centralized in `ValueConversions`. The JavaTypes below are for additional ad-hoc domain values transformation:

```
	protected <X> X getArray(BasicExtractor<X> extractor, java.sql.Array array, WrapperOptions options) throws SQLException {
		if ( array != null && getElementJdbcType() instanceof AggregateJdbcType ) {
			 ... ...
			return extractor.getJavaType().wrap( domainObjects, options );
		}
		else {
			return extractor.getJavaType().wrap( array, options );
		}
	}
```

e.g. the following usage is the intention of the above code:
```
class MyEntity {
     ... ...
     @JdbcTypeCode(INTERVAL_SECOND)
     Duration duration;
}
```
then the above code would transform from stored jdbc numeric value to Duration domain object.

NathanQingyangXu: I messed up. The above example is for JdbcType, not JavaType. The classic example is at Hibernate User Guide: https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#basic-bitset
Particularly below is a typical JavaType example:
```
@Entity(name = "Product")
public static class Product {
	@Id
	private Integer id;

	@JavaType(BitSetJavaType.class)
	private BitSet bitSet;

	//Constructors, getters, and setters are omitted for brevity
}
```

src/main/java/com/mongodb/hibernate/internal/type/MongoBigDecimalJavaType.java
NathanQingyangXu: I am concerned about the implication. To me it seems a hack with many undesirable features, though it works on the existing testing cases:

- it is intrusive implementation. Very likely it would interfere with existing JavaType mechnaism (like explicit JavaType specification on entity or struct field). Currently we don't need to cover JavaType feature, but it would be great not to make its support too cumbersome
- It seems unnecessary. As in my other comment points out, we could avoid them totally by simply reusing our `ValueConversions`'s `toArrayDomainValue` which has been used in `MongoResultSet#getArray()` already. It seems natural to reuse it on deeper level recursively
- Yeah, the ugly truth is we have to leak JDBC Driver implementation to higher level Hibernate abstraction. We might strive to create some centralized util method with minimal intrusion

src/integrationTest/java/com/mongodb/hibernate/CollectionIntegrationTests.java
NathanQingyangXu: Christian created a Hibernate ORM ticket at https://hibernate.atlassian.net/browse/HHH-19573

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -72,6 +93,18 @@ public static BsonBoolean toBsonValue(boolean value) {
         return BsonBoolean.valueOf(value);
     }
 
+    /**
+     * <a
+     * href="https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#basic-character">
```
NathanQingyangXu, line 98: It might be a little bit better to avoid Hibernate ORM version in the doc for it is a moving target. Using `current` in lieu of version usually serve the same purpose.
```
https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#basic-character
```

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -166,17 +256,39 @@ private static BigDecimal toDomainValue(BsonDecimal128 value) {
...
+        } else {
+            result = v;
+        }
+        return domainType.cast(result);
+    }
+
+    /** @see #toBsonValue(char) */
+    private static char toDomainValue(String value) {
+        assertTrue(value.length() == 1);
+        return value.charAt(0);
```
NathanQingyangXu, line 282: this method is a little bit strange in the sense that the input is not a BsonValue as in other methods. Also, it is hard to grasp its usage by its usage.

src/main/java/com/mongodb/hibernate/jdbc/MongoArray.java
```diff
@@ -0,0 +1,31 @@
...
+public final class MongoArray implements ArrayAdapter {
+    private final Object contents;
+
+    public MongoArray(Object contents) {
+        this.contents = contents;
+    }
+
+    @Override
+    public Object getArray() {
+        // Hibernate ORM does not call `Connection.getTypeMap`/`setTypeMap`, therefore we are free to ignore it
```
NathanQingyangXu, line 28: the above comment is not necessary from my perspective, for we know we are not developing JDBC Driver, and it is easy to know from `MongoConnection` implementation (no typeMap related method overriding) that we don't need to be concerned about type map.

For tech-savvy developer who is versed in Hibernate code logic, array is a relatively new feature and it is so natural that only basic features are implemented. Maybe in the long future type map is to be considered, but it is highly doubtable.

NathanQingyangXu (changes requested):
Overall another awesome PR from Valentin. Thorough and convincing.
I highly recommended removing those Mongo Java Types for they could be replaced by one-liner code change by reusing existing method in `ValueConversions`. They might be needed in the future, but for now they are not necessary.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -132,26 +153,60 @@ public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options)
...
+                if (!(jdbcMapping.getJdbcValueExtractor() instanceof Extractor<?> structValueExtractor)) {
+                    throw fail();
+                }
+                domainValue = structValueExtractor.getJdbcType().extractJdbcValues(value, options);
+            } else if (jdbcTypeCode == MongoArrayJdbcType.JDBC_TYPE.getVendorTypeNumber()) {
+                if (!(jdbcMapping.getJdbcType() instanceof MongoArrayJdbcType arrayJdbcType)) {
+                    throw fail();
+                }
+                if (!(jdbcMapping.getJdbcValueExtractor() instanceof BasicExtractor<?> jdbcValueExtractor)) {
+                    throw fail();
```
NathanQingyangXu, line 193: seems it is a good timing to introduce an assertion method to `MongoAssertions` to avoid the boilerplate:
```
    public static <T> T assertInstanceOf(Object value, Class<T> tClass) {
        if (!tClass.isInstance(value)) {
            throw new AssertionError();
        }
        return tClass.cast(value);
    }
```

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+        var item = new ItemWithArrayAndCollectionValues(
+                1,
+                // TODO-HIBERNATE-48 sprinkle on `null` array/collection elements
+                new byte[] {2, 3},
+                new char[] {'s', 't', 'r'},
+                new int[] {5},
+                new long[] {Long.MAX_VALUE, 6},
+                new double[] {Double.MAX_VALUE},
+                new boolean[] {true},
+                new Character[] {'s', 't', 'r'},
```
vbabanin, line 83: Hibernate treats `Byte[`] the same as `byte[]`, converting it to `BINARY` only when marked with `@JavaType(ByteArrayJavaType.class)`.

To cover all standard array types, we could add it to a field test set as well. What do you think?

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
vbabanin: Should we throw `FeatureNotSupported` with a link to a ticket as we do in other places until it is implemented?

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: Thank you! It's much nicer this way, and much less code. Done in c13fda99503339f43278bccf727b099152ee4991.

src/main/java/com/mongodb/hibernate/internal/type/MongoBigDecimalJavaType.java
stIncMale: I am not sure I agree with all the expressed arguments, but I did the proposed change because it is nice regardless of the aforementioned arguments.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
```diff
@@ -38,10 +39,11 @@ public String aggregateComponentCustomReadExpression(
             AggregateColumn aggregateColumn,
             Column column) {
         var aggregateColumnType = aggregateColumn.getTypeCode();
-        if (aggregateColumnType == MongoStructJdbcType.JDBC_TYPE.getVendorTypeNumber()) {
+        if (aggregateColumnType == MongoStructJdbcType.JDBC_TYPE.getVendorTypeNumber()
+                || aggregateColumnType == MongoArrayJdbcType.HIBERNATE_SQL_TYPE) {
             return format(
-                    "unused from %s.aggregateComponentCustomReadExpression",
-                    MongoAggregateSupport.class.getSimpleName());
+                    "unused from %s.aggregateComponentCustomReadExpression for SQL type code [%d]",
```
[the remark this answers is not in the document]
stIncMale: Related to https://github.com/mongodb/mongo-hibernate/pull/95#discussion_r2168261149.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
stIncMale: I don't think it does, which is why we have to set `WRAPPER_ARRAY_HANDLING` in `ArrayAndCollectionIntegrationTests`.

The reason `Byte[]` is not present in normal tests is a [bug in Hibernate ORM](https://hibernate.atlassian.net/browse/HHH-19573), which was created because @NathanQingyangXu reported our [observation](https://github.com/mongodb/mongo-hibernate/pull/95#discussion_r2081759710), for which we have the `ArrayAndCollectionIntegrationTests.Unsupported.testBoxedBytesArrayValue` test.

src/integrationTest/java/com/mongodb/hibernate/embeddable/StructAggregateEmbeddableIntegrationTests.java
```diff
@@ -114,38 +160,314 @@ void testNestedValues() {
...
+                        bigDecimalsCollection: null,
+                        objectIdsCollection: null,
+                        structAggregateEmbeddablesCollection: null
+                    }
+                }
+                """);
+        var loadedItem = sessionFactoryScope.fromTransaction(
+                session -> session.find(ItemWithNestedValueHavingArraysAndCollections.class, item.id));
+        // `loadedItem.nested` is `null` despite `item.nested` not being `null`.
+        // TODO-HIBERNATE-48 There is nothing we can do here, such is the Hibernate ORM behavior. - or can we?
```
NathanQingyangXu, line 464: If Hibernate decided to return null, we'd better accept the truth. There is no official Hibernate extension point since v7. We can't change the behaviour without changing the class file (e.g. AOP) which is out of our considersation currently.

There is a Hibernate bug that empty struct won't end up with null when its containing array is loaded. I created the ticket at after confirmation: https://hibernate.atlassian.net/browse/HHH-19575

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -166,17 +256,39 @@ private static BigDecimal toDomainValue(BsonDecimal128 value) {
...
+
+    private static <T> T toDomainValue(BsonString value, Class<T> domainType) {
+        var v = value.getValue();
+        Object result;
+        if (domainType.equals(Character.class)) {
+            result = toDomainValue(v);
+        } else {
+            result = v;
+        }
+        return domainType.cast(result);
```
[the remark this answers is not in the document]
stIncMale: The `toDomainValue` you are suggesting to inline exists as a counterpart to `BsonString toBsonValue(char value)`, and their docs refer to each other. Without this method, the method `<T> T toDomainValue(BsonString value, Class<T> domainType)` would have been the only candidate to be documented as a counterpart to `BsonString toBsonValue(char value)`, which it apparently is not.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+        assertEq(updatedItem, loadedItem);
+    }
+
+    /**
+     * @see EmbeddableIntegrationTests#testFlattenedValueHavingNullArraysAndCollections()
+     * @see StructAggregateEmbeddableIntegrationTests#testNestedValueHavingNullArraysAndCollections()
+     */
+    @Test
+    @Disabled("TODO-HIBERNATE-48 https://jira.mongodb.org/browse/HIBERNATE-48 enable this test")
+    public void testArrayAndCollectionValuesOfEmptyStructAggregateEmbeddables() {
```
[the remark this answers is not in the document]
stIncMale: It has to be `public` as otherwise the Java documentation links to this method become broken.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
stIncMale: Its documentation, or, rather, the method it links to` should explain the situation.

This comment is related to https://github.com/mongodb/mongo-hibernate/pull/95#discussion_r2169285684.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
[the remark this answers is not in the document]
NathanQingyangXu: Given the existence of the other comment, I resolved this comment.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+                ArraysAndCollections[] structAggregateEmbeddables,
+                Collection<ArraysAndCollections> structAggregateEmbeddablesCollection) {
+            this.id = id;
+            this.structAggregateEmbeddables = structAggregateEmbeddables;
+            this.structAggregateEmbeddablesCollection = structAggregateEmbeddablesCollection;
+        }
+    }
+
+    @Nested
+    class Unsupported {
```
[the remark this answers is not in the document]
NathanQingyangXu: I think it is related to our decision not supporting those Java primitive types which have no one-to-one mapping with BsonValues. For the same reason, `float[]` is not tested as well.

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
```diff
@@ -100,6 +102,15 @@ private static void forbidStructIdentifier(PersistentClass persistentClass) {
         }
     }
 
+    private static void forbidEmbeddablesWithoutPersistentAttributes(InFlightMetadataCollector metadata) {
+        metadata.visitRegisteredComponents(component -> {
+            if (component.getProperties().isEmpty()) {
```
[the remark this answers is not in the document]
stIncMale: Yes, it's better to require at least one insertable column to be present. Done in 15737464ac1bd6d2df98e40d4c1d5a8a14399909.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,697 @@
...
+                        objectIdsCollection: [{$oid: "000000000000000000000001"}],
+                        structAggregateEmbeddablesCollection: [{a: 1}]
+                    }],
+                    structAggregateEmbeddablesCollection: []
+                }
+                """);
+        var loadedItem = sessionFactoryScope.fromTransaction(session -> session.find(
+                ItemWithArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndCollections.class, item.id));
+        assertEq(item, loadedItem);
+        ItemWithArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndCollections updatedItem =
```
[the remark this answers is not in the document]
stIncMale: Sorry, those are typos, we agree to use `var` whenever possible. Done in 1ca82017c9a82fce0715647875c71c43d7cd81aa.

As for the class name - it follows the naming convention of the rest of the entity classes, and happens to be so long 🤷‍♂️

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -92,17 +125,58 @@ public static BsonString toBsonValue(String value) {
...
+            contents = value.getArray();
+        } catch (SQLException e) {
+            throw fail(e.toString());
+        }
+        return arrayToBsonValue(contents);
+    }
+
+    private static BsonArray arrayToBsonValue(Object value) throws SQLFeatureNotSupportedException {
+        var length = Array.getLength(value);
+        var elements = new ArrayList<BsonValue>(length);
```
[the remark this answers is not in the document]
stIncMale: We don't use `bsonValue` anywhere in `ValueConversions`, so I am not sure how it makes naming more consistent. I used `element` here and in `toDomainValue` simply to refer to an array element, regardless of whether it's of a domain type or not.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
stIncMale: > ending up with a seemingly working app (no exception) but would fail in mysterious way

We know that's not the case, there will be an exception about invalid MQL thrown as a result of a server error.

> Ellegant error handling requires explict exception instead.

As I previously said, there is no way of throwing an exception from `MongoAggregateSupport.aggregateComponentCustomReadExpression`/`aggregateComponentAssignmentExpression` in the situations when we currently return strings. If you see how to throw an exception from there while passing tests, could you please suggest a code change?

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+                    structAggregateEmbeddablesCollection: []
+                }
+                """);
+        var loadedItem = sessionFactoryScope.fromTransaction(
+                session -> session.find(ItemWithArrayAndCollectionValues.class, item.id));
+        assertEq(item, loadedItem);
+    }
+
+    @Test
+    @Disabled("TODO-HIBERNATE-48 https://jira.mongodb.org/browse/HIBERNATE-48 enable this test")
```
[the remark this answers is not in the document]
stIncMale: The `@JiraKey` annotation cannot be used to replace the `TODO-Jira-ID` notes in all the places were we have them. Therefore, if we start using it, we'll have more than one way of doing the same thing, which is not beneficial.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+                ItemWithArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndCollections.class, item.id));
+        assertEq(item, loadedItem);
+    }
+
+    @Override
+    public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
+        this.sessionFactoryScope = sessionFactoryScope;
+    }
+
+    private static void assertCollectionContainsExactly(String json) {
```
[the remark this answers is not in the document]
stIncMale: I am not sure what you think is not correct in the method name. I clarified the parameter name in 865cf7e4cf99812732f4fffc371a07ae41880563: `json` -> `documentAsJsonObject`.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
NathanQingyangXu: Now I got it. This method will be invoked during Hibernate bootstrap, so exception is not an option. We'd thought it would be invoked at run-time. Sorry about that.

NathanQingyangXu (approved):
From my side, no LGTM blocker for me.
There is an interesting discussion regarding struct fetching robustness issue, but it could be done later, if the issue is valid.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: Done in ed6fb2c4b320eb5aa875a60772ba786ab64a3fcd.

src/main/java/com/mongodb/hibernate/dialect/MongoAggregateSupport.java
stIncMale: Since this thread was started by me, and we all seem to agree now that there is nothing to do, I am resolving it.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,697 @@
...
+                },
+                List.of('s', 't', 'r'),
+                List.of(5),
+                List.of(Long.MAX_VALUE, 6L),
+                List.of(Double.MAX_VALUE),
+                List.of(true),
+                List.of("str"),
+                List.of(BigDecimal.valueOf(10.1)),
+                List.of(new ObjectId("000000000000000000000001")),
+                List.of(new StructAggregateEmbeddableIntegrationTests.Single(1)));
```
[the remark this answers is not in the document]
stIncMale: Let's discuss that separately and, if the team agrees, do separately.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+        annotatedClasses = {
+            ArrayAndCollectionIntegrationTests.ItemWithArrayAndCollectionValues.class,
+            ArrayAndCollectionIntegrationTests
+                    .ItemWithArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndCollections.class,
+            ArrayAndCollectionIntegrationTests.Unsupported.ItemWithBoxedBytesArrayValue.class,
+            ArrayAndCollectionIntegrationTests.Unsupported.ItemWithBytesCollectionValue.class
+        })
+@ServiceRegistry(settings = {@Setting(name = WRAPPER_ARRAY_HANDLING, value = "allow")})
+@ExtendWith(MongoExtension.class)
+public class ArrayAndCollectionIntegrationTests implements SessionFactoryScopeAware {
```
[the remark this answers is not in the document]
stIncMale: "_avoid putting testing case under root package_" has nothing to do with "_better organize our testing classes_". The `com.mongodb.hibernate` package is as good as any other package. We should put code in subpackages when it it makes sense, and should not when it does not make sense. We definitely should not be showeling code in subpackages just to avoid having it in `com.mongodb.hibernate`.

src/main/java/com/mongodb/hibernate/jdbc/MongoArray.java
stIncMale: > the above comment is not necessary

Our implementation of the `java.sql.Array.getArray` method violates its specification. The comment explains why such a violation is fine. I think, this comment is needed.

> we know we are not developing JDBC Driver

Yet we try to follow the specification where possible. If Hibernate ORM used `Connection.getTypeMap`/`setTypeMap`, we would likely have to take those into account. 

"we are not developing JDBC Driver" is not the reason we are diverging from the specification when implementing `java.sql.Array.getArray`. The reason is that Hibernate ORM does not use the methods mentioned in the comment. And this explanation worth to be put in a comment.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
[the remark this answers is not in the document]
stIncMale: `Unsupported.testBoxedBytesArrayValue`/`testBytesCollectionValue` are there only to expose the Hibernate ORM bug that prevents us from supporting arrays/collections of `Byte`.

If we want to make sure we throw a clear exception when types that we don't support at the JDBC level are used, we should add tests for those types to `MongoPreparedStatementTests`/`MongoResultSetTests`, which is unrelated to the current PR. Though I am not sure if such tests are worth the effort, since our approach with `*Adapter` interfaces obviously takes care of that as long as it is used. I am, however, also have no objections to them.

src/integrationTest/java/com/mongodb/hibernate/ArrayAndCollectionIntegrationTests.java
```diff
@@ -0,0 +1,775 @@
...
+                    }]
+                }
+                """);
+        var loadedItem = sessionFactoryScope.fromTransaction(session -> session.find(
+                ItemWithArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndCollections.class, item.id));
+        assertEq(item, loadedItem);
+    }
+
+    @Override
+    public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
```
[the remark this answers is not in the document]
stIncMale: Done in 20f8e8332ce76ce766d7029f1fe15bde35c2398b.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -132,26 +153,60 @@ public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options)
...
             throw fail();
         }
+        var embeddableMappingType = getEmbeddableMappingType();
         var result = new Object[bsonDocument.size()];
         var elementIdx = 0;
         for (var value : bsonDocument.values()) {
-            assertNotNull(value);
-            result[elementIdx++] =
-                    value instanceof BsonDocument ? extractJdbcValues(value, options) : toDomainValue(value);
+            var jdbcMapping =
```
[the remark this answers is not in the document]
stIncMale: The `instanceof` assertions were replaced with the `assertInstanceOf` method. The body of the `MongoStructJdbcType.extractJdbcValues` method currently takes 34 lines, and expected to take 32 lines when we start supporting `null`/missing values. This does not seem to be large, and is less than one screen.

NathanQingyangXu (approved):

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -59,19 +76,33 @@ public static BsonValue toBsonValue(Object value) throws SQLFeatureNotSupportedE
...
+        } else if (value instanceof char[] v) {
+            return toBsonValue(v);
         } else if (value instanceof ObjectId v) {
             return toBsonValue(v);
-        } else {
-            throw new SQLFeatureNotSupportedException(format(
-                    "Value [%s] of type [%s] is not supported",
-                    value, value.getClass().getTypeName()));
+        } else if (value instanceof Object[] v) {
+            return arrayToBsonValue(v);
```
NathanQingyangXu, line 84: There is some subtlety confusing above. So for primitive array (except for `byte[]` and `char[]`), the above `instanceof Object[]` won't catch it, so this method doesn't align with the `mirror` method of `toDomainValue()` in the sense that the latter accepts converntion from primitive array Bson wrapper (e.g. `int[]`) but the former doesn't accept them.

Is it intentional? Currently primitive array like `int[]` won't be passed to the above method, but why such discrepancy? 

For instance, we could change the above to
```
        } else if (value.getClass().isArray()) {
            return arrayToBsonValue(value);
        }
```
and we can make centralize all the upstream logic through this common method, by changing `BsonArray toBsonValue(java.sql.Array value)`
to
```
    public static BsonArray toBsonValue(java.sql.Array value) throws SQLFeatureNotSupportedException {
        Object contents;
        try {
            contents = value.getArray();
        } catch (SQLException e) {
            throw fail(e.toString());
        }
        return (BsonArray) toBsonValue(contents);
    }
```
Is the above normal primitive array refusal by design?

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -188,4 +290,23 @@ public static ObjectId toObjectIdDomainValue(BsonValue value) {
...
     }
+
+    public static MongoArray toArrayDomainValue(BsonValue value) throws SQLFeatureNotSupportedException {
+        return new MongoArray(toDomainValue(value.asArray(), Object.class));
+    }
+
+    private static Object toDomainValue(BsonArray value, Class<?> elementType) throws SQLFeatureNotSupportedException {
+        var size = value.size();
+        var result = Array.newInstance(elementType, size);
+        for (int i = 0; i < size; i++) {
```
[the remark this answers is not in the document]
stIncMale: Sorry for that, done in 2998250535cea56758cde56204c6185dcac8989e.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
NathanQingyangXu: showcased at https://github.com/mongodb/mongo-hibernate/commit/2cbd20bc1f060e32392ada45ab7e8d91f38c54d7

Currently there is a backdoor and the alternative makes everything go through the centralized `toBsonValue()`. More for logic cleanness. Not mandatory change. Up to you to decide.

NathanQingyangXu:
I should have raised this concern in last PR. I think we should create some testing case to showcase that Struct's field annotated with customized `@Column` should take effect. I tested it on main branch and found it works as expected, but it seems not to hurt to provide an integration testing case to cover our asses. An example testing case is like below:

```

@DomainModel(
        annotatedClasses = {
                TestStructWithCustomizedColumnName.Book.class,
                TestStructWithCustomizedColumnName.Author.class
        })
class TestStructWithCustomizedColumnName {

    private static MongoCollection<BsonDocument> mongoCollection;

    @Test
    void testFieldNameIsCustomized(SessionFactoryScope scope) {
        var book = new Book();
        book.id = 1;
        book.title = "Effective Java";
        var author = new Author();
        author.firstName = "Joshua";
        author.lastName = "Bloch";
        book.author = author;
        scope.inTransaction(session -> session.persist(book));
        assertThat(mongoCollection.find()).containsExactly(BsonDocument.parse(
                """
                {
                  "_id": 1,
                  "title": "Effective Java",
                  "author": {
                    "first_name": "Joshua",
                    "last_name": "Bloch"
                  }
                }
                """
        ));
    }

    @Entity(name = "Book")
    @Table(name = "books")
    static class Book {
        @Id
        int id;
        String title;
        Author author;
    }

    @Embeddable
    @Struct(name = "Author")
    static class Author {
        @Column(name = "first_name")
        String firstName;
        @Column(name = "last_name")
        String lastName;

    }
}
```

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
NathanQingyangXu: Now I see your point. Ignore my previous suggestion. We don't need to change for we are sure only `Object[]` will be passed into `ValuleConvertions#toBsonValule(Object)` due to Hibernate's internal mechnism.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -64,10 +67,18 @@ private MongoStructJdbcType() {
 
     private MongoStructJdbcType(
             @Nullable EmbeddableMappingType embeddableMappingType, @Nullable String structTypeName) {
+        if (embeddableMappingType != null && embeddableMappingType.isPolymorphic()) {
+            throw new FeatureNotSupportedException("Polymorphic mapping is not supported");
```
NathanQingyangXu, line 71: Have we decided not to support polymorphic forever for temporarily? Seems it is the latter.
Should we put ticket reference here instead?
Maybe https://jira.mongodb.org/browse/HIBERNATE-69? Or should we create a sub-task or new ticket for it?

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -132,26 +152,52 @@ public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options)
...
-        var elementIdx = 0;
-        for (var value : bsonDocument.values()) {
-            assertNotNull(value);
-            result[elementIdx++] =
-                    value instanceof BsonDocument ? extractJdbcValues(value, options) : toDomainValue(value);
+        var bsonDocument = assertInstanceOf(rawJdbcValue, BsonDocument.class);
+        var embeddableMappingType = getEmbeddableMappingType();
+        var jdbcValueCount = embeddableMappingType.getJdbcValueCount();
+        var result = new Object[jdbcValueCount];
+        for (int columnIndex = 0; columnIndex < jdbcValueCount; columnIndex++) {
```
NathanQingyangXu, line 172: there are still cases that `var` is not used consistently, though I am perfectly fine with it.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
stIncMale: I mentioned in the description of that ticket that "The polymorphism should also be supported for @Embeddable @Struct." We will be able to move this part out to a new ticket if needed when we start working on polymorphims.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -116,14 +136,14 @@ public BsonDocument createJdbcValue(Object domainValue, WrapperOptions options)
...
-                }
-                bsonValue = structJdbcType.createJdbcValue(value, options);
+            var jdbcTypeCode = jdbcMapping.getJdbcType().getJdbcTypeCode();
+            if (jdbcTypeCode == getJdbcTypeCode()) {
+                var structValueBinder = assertInstanceOf(jdbcMapping.getJdbcValueBinder(), Binder.class);
+                bsonValue = structValueBinder.getJdbcType().createBsonValue(value, options);
+            } else if (jdbcTypeCode == MongoArrayJdbcType.JDBC_TYPE.getVendorTypeNumber()) {
+                @SuppressWarnings("unchecked")
+                ValueBinder<Object> valueBinder = jdbcMapping.getJdbcValueBinder();
+                bsonValue = toBsonValue(valueBinder.getBindValue(value, options));
```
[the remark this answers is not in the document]
stIncMale: Another argument, admittedly also not a very strong one in this case, for leaving the code as is consistency: while it's possible to handle structs and arrays using the same code branch in `MongoStructJdbcType.createBsonValue`, a similar approach is not possible in `MongoStructJdbcType.extractJdbcValues`, as there we have to use methods specific to `MongoStructJdbcType`/`MongoArrayJdbcType`.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -113,16 +186,29 @@ static Object toDomainValue(BsonValue value) throws SQLFeatureNotSupportedExcept
...
+            }
         } else if (value instanceof BsonBinary v) {
             return toDomainValue(v);
         } else if (value instanceof BsonObjectId v) {
             return toDomainValue(v);
-        } else {
-            throw new SQLFeatureNotSupportedException(format(
-                    "Value [%s] of type [%s] is not supported",
-                    value, value.getClass().getTypeName()));
+        } else if (value instanceof BsonArray v && domainType.isArray()) {
```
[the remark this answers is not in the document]
stIncMale: When the data being read does not match the expectations based on the domain model, we should throw an exception instead of throwing an `AssertionError`: `AssertionError`s indicate bugs in our code, but encountering data we don't support is not a bug. So there is nothing to change here.

However, there is a different place I found where assertion was used instead of throwing an exception, and I created a followup PR https://github.com/mongodb/mongo-hibernate/pull/112 to fix that.

PR #97, opened 2025-05-15
Improve spotless java formatter usage

NathanQingyangXu:
Previously we create a gradle plugin based on `buildSrc` to go about spotless java config. However, recently some defects are found as below:

- spotless java are configed both in root project and `builsSrc` kotlin dsl; gradle dictates `buildSrc` will be run first. However, some format tweaking might be overridden by `palantir` simply because it is configured in root project and will be run later (example will be given in code review comments where appropriate)
- we previously solved the multiline indentating issue (out of radar of `palantir`) in `buildSrc` by introducing a new format step; however, the following format issue (https://github.com/mongodb/mongo-hibernate/blob/main/src/integrationTest/java/com/mongodb/hibernate/BasicCrudIntegrationTests.java#L106) was flushed out and is out of radar of `palantir` again:

```
var expectedDocument = BsonDocument.parse(
                    """
                    {
                        _id: 1,
                        title: "War and Peace",
                        author: "%s",
                        publishYear: 1867
                    }"""
                            .formatted(author));
```

Note that `palantir` dictates the `.formatted(...)` will be split from the multiline text block, which is not ideal.

This PR will solve the following issues:

1. move spotless java config from root project to `buildSrc`, ending up with easy control to ensure our customized format step would happen AFTER `palantir`
2. a new format step was introduced in `buildSrc` to solve the above parameterized multiline format issue.
3. other minor gradle issues


src/test/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstMatchStageTests.java
```diff
@@ -35,7 +35,8 @@ void testRendering() {
                 new AstComparisonFilterOperation(EQ, new AstLiteralValue(new BsonString("Jane Eyre"))));
         var astMatchStage = new AstMatchStage(astFilter);
 
-        var expectedJson = """
+        var expectedJson =
+                """
                 {"$match": {"title": {"$eq": "Jane Eyre"}}}\
                 """;
         assertRender(expectedJson, astMatchStage);
```
NathanQingyangXu, line 42: the above issue was rectified simply due to the fact that our format steps happen AFTER `palantir`

buildSrc/src/main/kotlin/spotless-java-extension.gradle.kts
```diff
@@ -58,12 +79,34 @@ class MultilineFormatter : Serializable {
...
+            } else {
+                result.append(line)
+            }
+            i++
+            if (i < lines.size) {
+                result.append("\n")
+            }
+        }
+        return result.toString()
+    }
```
NathanQingyangXu, line 111: we might well merge the above logic into existing formatter which mainly aims for indentation, but that would make the code logic difficult to understand. One step one concern makes for much maintainable code.

buildSrc/src/main/kotlin/spotless-java-extension.gradle.kts
```diff
@@ -58,12 +79,34 @@ class MultilineFormatter : Serializable {
             val minIndent =
                 multilineStringLines
                     .filter { it.isNotBlank() }
-                    .map { l -> l.indexOfFirst { ch -> !ch.isWhitespace() }.takeIf { it >= 0 } ?: line.length }
-                    .minOrNull() ?: 0
+                    .minOfOrNull { l -> l.indexOfFirst { ch -> !ch.isWhitespace() }.takeIf { it >= 0 } ?: line.length }
+                    ?: 0
```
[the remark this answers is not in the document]
katcharov: What is the defect?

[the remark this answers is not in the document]
NathanQingyangXu: the defect seems we can combine the two statements into one. Not a big deal. Simply more concise.

buildSrc/src/main/kotlin/spotless-java-extension.gradle.kts
```diff
@@ -24,44 +24,96 @@ plugins {
...
+
+class MultilineWithParametersFormatter : MultilineFormatter() {
+    override fun format(content : String) : String {
+        val lines = content.lines()
+        val result = StringBuilder()
+        var i = 0
+        while (i < lines.size) {
+            val line = lines[i]
+            if (line.trimEnd().endsWith(TRIPLE_QUOTES) && i + 1 < lines.size
+                && lines[i + 1].trimStart().startsWith(".formatted(")) {
```
katcharov, line 108: This code seems too specific - a format should apply to all methods, not just this `formatted` method.

build.gradle.kts
katcharov, line 1: It seems that, if we decide to use a popular off-the-shelf format, we should "undo" that format only when there is something fairly serious. Otherwise, we lose the primary benefit of choosing such a formatter (that it saves us from frequently needing to make formatting decisions).

Separately, if the formatter does not cover something that we need, we can add such a step before the formatting happens. So the order would be:

1. Any custom formatting, like our multiline string formatter, which was not covered by our chosen formatter.
2. Our chosen formatter (`palantir`).
3. Overrides to the formatter.

I do not think that having a method on a separate line after multiline strings is enough to justify a customization of the formatter. I agree it looks a bit odd, but I worry about losing the benefits arising from a consistent application of the formatter.

PR #101, opened 2025-05-27
DML(insertion, updating and deletion) MQL translation

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-46

The scope of the ticket is to unblock the following simple DML HQL mutation queries:

- `insert into Book (id, title) values (1, "War and Peace"), (2, "Crime and Punishmnet")`
- `update Book set title = "War and Peace" where title = "War & Peace"`
- `delete from Book where title = "War and Peace"`

Translation is straightforward. Only one caveat is previously our insertion command mongo ast only allows one document, to fully support HQL's feature, the mongo ast model was refactored to allow multiple documents, consistent with MQL's syntax as well


src/integrationTest/java/com/mongodb/hibernate/query/mutate/AbstractMutateQueryIntegrationTests.java
```diff
@@ -0,0 +1,112 @@
...
+                queryPostProcessor.accept(query);
+            }
+            query.executeUpdate();
+            var mutateTranslator =
+                    ((MutateTranslatorAwareDialect) session.getJdbcServices().getDialect()).getMutateSqlAstTranslator();
+            assertThat(mutateTranslator.getAffectedTableNames()).isEqualTo(expectedAffectedTableNames);
+        });
+    }
+
+    public static final class MutateTranslatorAwareDialect extends Dialect {
```
NathanQingyangXu, line 74: An important feature to test is we need to include the table names involved in DML queries into the result of `SqlAstTranslator#getAffectedTableNames()` which acts as auto-flush purpose in the upper level caller of the translator.

We might need to finish other features (like table joining) to test the auto-flush features in integration test; currently we could only test this important feature by some expedient.

PR #102, opened 2025-05-27
simple DML translation

NathanQingyangXu:
https://jira.mongodb.org/browse/HIBERNATE-46

The scope of the ticket is to unblock the following simple DML HQL mutation queries:

- `insert into Book (id, title) values (1, "War and Peace"), (2, "Crime and Punishmnet")`
- `update Book set title = "War and Peace" where title = "War & Peace"`
- `delete from Book where title = "War and Peace"`

Translation is straightforward. The following new visitor methods would be implemented in this PR:

- void visitDeleteStatement(DeleteStatement deleteStatement);
- void visitUpdateStatement(UpdateStatement updateStatement);
- void visitInsertStatement(InsertSelectStatement insertSelectStatement);

Both of the first two methods share a lot in common in the sense that they are both based on where clause, with various AST filters as the backbones, which has been finished in previous PR. 

The scope of this PR is modest and only focuses on the following basic stuff:

- no new AST filter will be introduced
- when updating, only `$set` updating operator is supported (with the constraint that `$set` focuses on field path and value pair) 

One caveat is previously our insertion command mongo ast only allows one document, to fully support HQL's feature, the mongo ast model was refactored to allow multiple documents, consistent with MQL's syntax as well. There is a separate ticket for this requirment: https://jira.mongodb.org/browse/HIBERNATE-44, so we can finish it in this PR additionally.


src/integrationTest/java/com/mongodb/hibernate/query/mutate/AbstractMutateQueryIntegrationTests.java
```diff
@@ -0,0 +1,112 @@
...
+                queryPostProcessor.accept(query);
+            }
+            query.executeUpdate();
+            var mutateTranslator =
+                    ((MutateTranslatorAwareDialect) session.getJdbcServices().getDialect()).getMutateSqlAstTranslator();
+            assertThat(mutateTranslator.getAffectedTableNames()).isEqualTo(expectedAffectedTableNames);
+        });
+    }
+
+    public static final class MutateTranslatorAwareDialect extends Dialect {
```
NathanQingyangXu, line 74: An important feature to test is we need to include the table names involved in DML queries into the result of `SqlAstTranslator#getAffectedTableNames()` which acts as auto-flush purpose in the upper level caller of the translator.

We might need to finish other features (like table joining) to test the auto-flush features in integration test; currently we could only test this important feature by some expedient.

src/integrationTest/java/com/mongodb/hibernate/query/AbstractQueryIntegrationTests.java
```diff
@@ -36,17 +38,17 @@
 
 @SessionFactory(exportSchema = false)
 @ExtendWith(MongoExtension.class)
-abstract class AbstractSelectionQueryIntegrationTests implements SessionFactoryScopeAware, ServiceRegistryScopeAware {
+public abstract class AbstractQueryIntegrationTests implements SessionFactoryScopeAware, ServiceRegistryScopeAware {
```
NathanQingyangXu, line 41: this class is reusable not only for `select` query but also for `mutate` query, so it is renamed and moved to more general location. The `Book` class was changed accordingly to go together with it hand in hand.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -580,17 +577,93 @@ public void visitTuple(SqlTuple sqlTuple) {
...
     }
 
     @Override
-    public void visitInsertStatement(InsertSelectStatement insertSelectStatement) {
-        throw new FeatureNotSupportedException("TODO-HIBERNATE-46 https://jira.mongodb.org/browse/HIBERNATE-46");
+    public void visitInsertStatement(InsertSelectStatement insertStatement) {
+        checkMutationStatement(insertStatement);
+        if (insertStatement.getConflictClause() != null) {
+            throw new FeatureNotSupportedException();
+        }
```
NathanQingyangXu, line 627: HQL has an interesting ON CONFLICT during insertion feature (see https://www.baeldung.com/hibernate-insert-query-on-conflict-clause). It is sort of similar to MongoDB's `upsert` feature but it seems there is a big difference between them:

- HQL mainly focuses on conflict during insertion (w.r.t. unique constraint or fields combination); MongoDB's `upsert` focuses on conflict during updating (w.r.t. some query or filter,, e.g. comparison between a field and a value)

So I guess we can't support this HQL feature. HQL's update statement is simpler and provides no possibility to emulate MongoDB's `upsert` feature.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -580,17 +577,93 @@ public void visitTuple(SqlTuple sqlTuple) {
...
+        if (insertStatement.getSourceSelectStatement() != null) {
+            throw new FeatureNotSupportedException();
+        }
```
NathanQingyangXu, line 630: HQL has two flavours of insertion statement:

- `insert into Book (id, title) values (1, "War and Peace"), (2, "Crime and Punishment")`
- `insert into Book (id, title) select c.identifier, c.name from Catalog as c`

This PR will focus on the first variant and throw exception for the latter.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -580,17 +577,93 @@ public void visitTuple(SqlTuple sqlTuple) {
...
+                if (!isValueExpression(fieldValueExpression)) {
+                    throw new FeatureNotSupportedException();
+                }
+                var fieldValue = acceptAndYield(fieldValueExpression, FIELD_VALUE);
+                astElements.add(new AstElement(fieldName, fieldValue));
+            }
+            documents.add(new AstDocument(astElements));
+        }
+
+        astVisitorValueHolder.yield(COLLECTION_MUTATION, new AstInsertCommand(collection, documents));
```
NathanQingyangXu, line 660: there are 3 assertion usages above as follows:

- insertion field list is not empty
- insertion values list is not empty
- field list length is the same as values length

The above constraints were either enforced from HQL grammar level or semantic checking phase in Hibernate (a prior step when the final SQL AST translation step is reached), so we can rest assured these constraints are enforced already.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -392,16 +395,9 @@ private AstProjectStage createProjectStage(SelectClause selectClause) {
...
+        checkFromClauseSupportability(fromClause);
         var tableGroup = fromClause.getRoots().get(0);
-
-        if (!(tableGroup.getModelPart() instanceof EntityPersister entityPersister)
-                || entityPersister.getQuerySpaces().length != 1) {
-            throw new FeatureNotSupportedException();
-        }
-
+        var entityPersister = (EntityPersister) tableGroup.getModelPart();
         affectedTableNames.add(((String[]) entityPersister.getQuerySpaces())[0]);
```
NathanQingyangXu, line 401: the above `visitFromClause` will only be invoked for `selection query`, not for `mutation query`, though some common validation logic applies to both. For that reason, the logic is centralized in a new `checkFromClauseSupportability()` method

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: similarly, various assertions are used in updating case as well. Common constraints have been carefully centralized into `checkMutationStatementSupportability()` method. Only one checking is required to ensure the insertion field value or udpating field value expressions should be either `valueExpression` (literal value or parameter expressions).

src/integrationTest/java/com/mongodb/hibernate/query/mutate/AbstractMutateQueryIntegrationTests.java
NathanQingyangXu: It is. related to auto-flushing criteria No..2 (see https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html#flushing-auto):

- prior to executing a JPQL/HQL query that overlaps with the queued entity actions

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
NathanQingyangXu: Definitively it deserves a ticket. I think all of the following HQL criteria needs to be satisfied to allow for MQL translation:

- ON CONFLICT is specified
- ON CONSTRAINT uniqueIndexName is not used as the criteria (MQL's filter can't be based on unique index name, correct?)
- CONFLICT ACTION is not NOTHING

so it seems the ticket is an interesting tech one but could be used very rarely in reality; native query seems more flixible and powerful.

src/integrationTest/java/com/mongodb/hibernate/query/mutation/AbstractMutationQueryIntegrationTests.java
```diff
@@ -0,0 +1,101 @@
...
+
+@ServiceRegistry(
+        settings =
+                @Setting(
+                        name = DIALECT,
+                        value =
+                                "com.mongodb.hibernate.query.mutation.AbstractMutationQueryIntegrationTests$MutationTranslateResultAwareDialect"))
+public class AbstractMutationQueryIntegrationTests extends AbstractQueryIntegrationTests {
+
+    protected void assertExpectedAffectedCollections(String... expectedAffectedfCollections) {
```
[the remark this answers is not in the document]
NathanQingyangXu: good point. changed in a separate commit

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -956,4 +1021,34 @@ private static BsonValue toBsonValue(Object value) {
...
+        if (!cteContainer.getCteStatements().isEmpty()
+                || !cteContainer.getCteObjects().isEmpty()) {
+            throw new FeatureNotSupportedException("CTE not supported");
+        }
+    }
+
+    private static void checkMutationStatementSupportability(AbstractMutationStatement mutationStatement) {
+        checkCteContainerSupportability(mutationStatement);
+        if (!mutationStatement.getReturningColumns().isEmpty()) {
+            throw new FeatureNotSupportedException();
```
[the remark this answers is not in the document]
NathanQingyangXu: Updated as suggested.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -576,17 +571,87 @@ public void visitTuple(SqlTuple sqlTuple) {
...
+            fieldNames.add(fieldReference.getColumnExpression());
+        }
+
+        var valuesList = insertStatement.getValuesList();
+        assertFalse(valuesList.isEmpty());
+
+        var documents = new ArrayList<AstDocument>(valuesList.size());
+        for (var values : valuesList) {
+            var fieldValueExpressions = values.getExpressions();
+            assertTrue(fieldNames.size() == fieldValueExpressions.size());
```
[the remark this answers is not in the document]
NathanQingyangXu: DynamicInsert simply omits both field name and field value expression IF field value is bound to null. Also, `@DynamicInsert` only applies to model mutation case whose translation was done in m1 and it won't involve ad-hoc explict HQL insertion statement.

src/integrationTest/java/com/mongodb/hibernate/query/mutation/InsertionIntegrationTests.java
```diff
@@ -0,0 +1,142 @@
...
+import com.mongodb.hibernate.junit.InjectMongoCollection;
+import com.mongodb.hibernate.query.Book;
+import java.util.List;
+import org.bson.BsonDocument;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@DomainModel(annotatedClasses = Book.class)
+class InsertionIntegrationTests extends AbstractMutationQueryIntegrationTests {
```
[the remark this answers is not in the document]
vbabanin: Yes, we can get more details from the root cause returned by the driver. The `BulkWriteResult` contains `writeErrors`, which include error codes. Currently, when a transaction is aborted, the message doesn’t clarify what went wrong.

Hibernate ORM throws `ConstraintViolationException` for unique constraint violations - we should align with that behavior.

This can be achieved by handling `writeErrors` and throw an exception with the appropriate code. This can then be mapped in `SQLExceptionConversionDelegate` to the relevant Hibernate exception for user visibility.

When an exception is thrown from `executeUpdate`, Hibernate calls rollback on `MongoConnection` as expected, rather than attempting to commit a transaction that no longer exists.

Here’s a draft commit showing the relevant test and changes for illustration: 057bc0edae6a1799b83e3af38d3087d4dd0f6ea1

For this PR, I propose:
1. Add a test to cover ConstraintViolationException (to be disabled until writeError handling is implemented).

2. Throw SQLException with an error message and MongoDB error code. Handling of additional write error codes can be addressed in a separate ticket.

separate ticket has been created for proper implementation: [HIBERNATE-95](https://jira.mongodb.org/browse/HIBERNATE-95) later.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Do we have a ticket for the latter? If not, should we create one in the backlog to track it?

NathanQingyangXu: We don't have a ticket. From my understanding, there is no MQL counterpart so we won't support it in the first place.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -666,17 +661,97 @@ public void visitTuple(SqlTuple sqlTuple) {
 
     @Override
     public void visitDeleteStatement(DeleteStatement deleteStatement) {
-        throw new FeatureNotSupportedException("TODO-HIBERNATE-46 https://jira.mongodb.org/browse/HIBERNATE-46");
+        checkMutationStatementSupportability(deleteStatement);
+        var collectionAndAstFilter = getCollectionAndFilter(deleteStatement);
```
vbabanin, line 665: I think using the `CollectionAndFilter` record here might be a bit unnecessary, since we immediately unpack it and pass its fields separately, so they are not being treated as a single pair that travels through the rest of the code.

Also it looks like `collection` is only logically groped with `filter`, and not with `fieldUpdates`, however, it applies to both. 
```
 new AstUpdateCommand(
       collectionAndAstFilter.collection, 
       collectionAndAstFilter.filter,
       fieldUpdates),
```

Given that, it seems simpler to keep `collection` and` filter` as separate local variables, without introducing a record, we could create a separate method for `addToAffectedCollections` to keep, as i understand, the original intention of reducing code duplication and remove grouping.

This method could be used in insert, update and delete:
```
private String addToAffectedCollections(NamedTableReference tableRef) {
    var collection = tableRef.getTableExpression();
    affectedTableNames.add(collection);
    return collection;
}
```

Usage:
```
var collection = addToAffectedCollections(updateStatement.getTargetTable());
var astFilter  = acceptAndYield(updateStatement.getRestriction(), FILTER);
```

What do you think?

NathanQingyangXu: I like it and adopted it. I dslike that record as well. thanks.
In other language like Python's unpack, it would be much more easy to pass multiple values from a method.

src/integrationTest/java/com/mongodb/hibernate/query/mutation/InsertionIntegrationTests.java
```diff
@@ -0,0 +1,160 @@
...
+class InsertionIntegrationTests extends AbstractQueryIntegrationTests {
+
+    @InjectMongoCollection(Book.COLLECTION_NAME)
+    private static MongoCollection<BsonDocument> mongoCollection;
+
+    @BeforeEach
+    void beforeEach() {
+        getTestCommandListener().clear();
+    }
+
```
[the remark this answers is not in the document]
NathanQingyangXu: thanks. Your testing case also showcases that such "partial insertion" is another source of mixture of null and missing fields (another one is `@DynamicInsert`). So the mixture seems inevitable.

vbabanin (dismissed):

NathanQingyangXu:
@stIncMale , would appreciate your taking a secondary review for @vbabanin had approved it for a while. Thanks.

src/integrationTest/java/com/mongodb/hibernate/query/mutation/UpdatingIntegrationTests.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private static final List<Book> testingBooks = List.of(
+            new Book(1, "War & Peace", 1869, true),
+            new Book(2, "Crime and Punishment", 1866, false),
+            new Book(3, "Anna Karenina", 1877, false),
+            new Book(4, "The Brothers Karamazov", 1880, false),
+            new Book(5, "War & Peace", 2025, false));
+
+    @BeforeEach
+    protected void beforeEach() {
```
stIncMale, line 44:
```suggestion
    void beforeEach() {
```

src/integrationTest/java/com/mongodb/hibernate/query/mutation/InsertionIntegrationTests.java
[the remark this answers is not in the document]
stIncMale: > the mixture seems inevitable

It's been said long ago, and no one ever claimed otherwise - applications may introduce a mixture of absent fields and null values. What was said also is that it's better for us to decide and store only one or the other always, provided we have control over what is being stored.

stIncMale (changes requested):

stIncMale:
The last reviewed commit is ea75d2f5080cfb14307c63925c90a88b1b118488.

src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionWhereClauseIntegrationTests.java
```diff
@@ -19,19 +19,22 @@
...
 @DomainModel(annotatedClasses = Book.class)
-class BooleanExpressionWhereClauseIntegrationTests extends AbstractSelectionQueryIntegrationTests {
+class BooleanExpressionWhereClauseIntegrationTests extends AbstractQueryIntegrationTests {
 
     private Book bookOutOfStock;
     private Book bookInStock;
 
     @BeforeEach
-    void beforeEach() {
+    protected void beforeEach() {
```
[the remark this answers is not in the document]
NathanQingyangXu: this is a legacy minor issue; tbh I don't recollect the reason. Removed.

src/integrationTest/java/com/mongodb/hibernate/query/AbstractQueryIntegrationTests.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @Override
+    public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
+        this.sessionFactoryScope = sessionFactoryScope;
+    }
+
+    @Override
+    public void injectServiceRegistryScope(ServiceRegistryScope serviceRegistryScope) {
+        this.testCommandListener = serviceRegistryScope.getRegistry().requireService(TestCommandListener.class);
+    }
```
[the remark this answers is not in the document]
NathanQingyangXu: ok, didn't notice we have such convention.

src/integrationTest/java/com/mongodb/hibernate/query/select/SortingSelectQueryIntegrationTests.java
```diff
@@ -60,6 +58,12 @@ private static List<Book> getBooksByIds(int... ids) {
                 .toList();
     }
 
+    @BeforeEach
+    protected void beforeEach() {
```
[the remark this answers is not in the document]
NathanQingyangXu: removed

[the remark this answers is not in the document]
NathanQingyangXu: I didn't notice such convention and I tried to be consistent with `LimitOffsetFetchClauseIntegrationTests`; relocate `@BeforeEach` in both classes after field declaration.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
stIncMale: > our user could easily achieve the same goal by breaking down the process to two separate steps programmatically (first grabbing all the values using some selection or native query and then using the normal insertion mutation or native query) and it is much more flexible

While I don't know if we can actually support `insert into ... select ...`, I know that the approach you described cannot generally replace `insert into ... select ...`. The important difference between the two approaches is that one requires the data to be transferred from a DBMS to an application (which, when Hibernate ORM is used, is further aggravated by transforming the data into entities/projections), while the other does not.

We don't have to create a ticket, I expect we will consider this functionality and start actually thinking about whether it is doable, only when there is demand for it.

stIncMale (approved):

PR #103, opened 2025-05-29
some tech debts

NathanQingyangXu:
Some minor issues accumulated so far, including:

- version bumping of both hibernate and mongo driver libraries
- an overlooked JSON formatting in `com.mongodb.hibernate.query.select.BooleanExpressionWhereClauseIntegrationTests`
- minor improvements to gradle usage
- misc

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -131,15 +131,14 @@ buildConfig {
 dependencies {
     testImplementation(libs.bundles.test.common)
     testImplementation(libs.mockito.junit.jupiter)
-    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testRuntimeOnly(libs.junit.platform.launcher)
     testCompileOnly(libs.checker.qual)
 
     integrationTestImplementation(libs.bundles.test.common)
-    @Suppress("UnstableApiUsage")
     integrationTestImplementation(libs.hibernate.testing) {
         exclude(group = "org.apache.logging.log4j", module = "log4j-core")
     }
-    integrationTestRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    integrationTestRuntimeOnly(libs.junit.platform.launcher)
 
     api(libs.jspecify)
 
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionWhereClauseIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionWhereClauseIntegrationTests.java
@@ -54,9 +54,32 @@ class BooleanExpressionWhereClauseIntegrationTests extends AbstractSelectionQuer
         assertSelectionQuery(
                 "from Book where" + (negated ? " not " : " ") + "outOfStock",
                 Book.class,
-                "{'aggregate': 'books', 'pipeline': [{'$match': {'outOfStock': {'$eq': "
-                        + (negated ? "false" : "true")
-                        + "}}}, {'$project': {'_id': true, 'discount': true, 'isbn13': true, 'outOfStock': true, 'price': true, 'publishYear': true, 'title': true}}]}",
+                """
+                {
+                  "aggregate": "books",
+                  "pipeline": [
+                    {
+                      "$match": {
+                        "outOfStock": {
+                          "$eq": %s
+                        }
+                      }
+                    },
+                    {
+                      "$project": {
+                        "_id": true,
+                        "discount": true,
+                        "isbn13": true,
+                        "outOfStock": true,
+                        "price": true,
+                        "publishYear": true,
+                        "title": true
+                      }
+                    }
+                  ]
+                }
+                """
+                        .formatted(negated ? "false" : "true"),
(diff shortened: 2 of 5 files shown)


build.gradle.kts
```diff
@@ -131,15 +131,14 @@ buildConfig {
...
+    testRuntimeOnly(libs.junit.platform.launcher)
     testCompileOnly(libs.checker.qual)
 
     integrationTestImplementation(libs.bundles.test.common)
-    @Suppress("UnstableApiUsage")
     integrationTestImplementation(libs.hibernate.testing) {
         exclude(group = "org.apache.logging.log4j", module = "log4j-core")
     }
-    integrationTestRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    integrationTestRuntimeOnly(libs.junit.platform.launcher)
```
NathanQingyangXu, line 141: previously I didn't create reference in toml for no version is required; however, it is still better to centralize its maven coordinates in it.

src/integrationTest/java/com/mongodb/hibernate/query/select/BooleanExpressionWhereClauseIntegrationTests.java
```diff
@@ -54,9 +54,32 @@ void testBooleanFieldPathExpression(boolean negated) {
...
+                        "outOfStock": true,
+                        "price": true,
+                        "publishYear": true,
+                        "title": true
+                      }
+                    }
+                  ]
+                }
+                """
+                        .formatted(negated ? "false" : "true"),
```
NathanQingyangXu, line 82: In my last Sorting PR, I reformatted the expected JSON in all the other integration testing, but overlooked this class.

src/main/java/com/mongodb/hibernate/internal/VisibleForTesting.java
```diff
@@ -36,6 +36,6 @@
     enum AccessModifier {
         PRIVATE,
         PACKAGE,
-        PROTECTED;
+        PROTECTED
```
NathanQingyangXu, line 39: IDE points the warning that `;` is unnecessary above

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -949,7 +950,10 @@ private static boolean isComparingFieldWithValue(ComparisonPredicate comparisonP
                 || (isFieldPathExpression(rhs) && isValueExpression(lhs));
     }
 
-    private static BsonValue toBsonValue(Object value) {
+    private static BsonValue toLiteralBsonValue(Object value) {
+        if (value instanceof ObjectId) {
+            throw new FeatureNotSupportedException();
+        }
```
NathanQingyangXu, line 956: Thanks for my colleague's refactoring to reuse his `ValueConversions`, but there is one caveat that not all domain values has Bson literal counterpart, noticably `ObjectId`. Also chnaged the method name to make it more explicit and less confusing.

jyemin: How so?  We already have in `ValueConversions` a call to this method:

```
    public static BsonObjectId toBsonValue(ObjectId value) {
        return new BsonObjectId(value);
    }
```

stIncMale: Even if `ValueConversions.toBsonValue` did not handle `ObjectId`, which id does, it would have been the responsibility of `ValueConversions.toBsonValue` to throw `SQLFeatureNotSupportedException`, and explicitly checking `instanceof ObjectId` to throw `FeatureNotSupportedException` prematurely would still have been surprising.

NathanQingyangXu: This method is used to transform query literal domain value to BSON literal values (as shown in `SimpleSelectQueryIntegrationTests#QueryLiteralTests`). For instance, below is an example HQL involving `BigDecimal` (HQL grammar choses `BD` suffix as the indicator):

```
"from Book where price = 123.50BD"
```

However, there is no `ObjectId` query literal recognized by Hibernate. For that reason, I added the exclusion logic to guard against it. It is hard to validate it by testing case for the following possible HQL won't work due to Hibernate's semantic checking:
```
var query = session.createSelectionQuery("from Book where id = :id").setParameter("id", objectId);
```
Currently only `ObjectId` should be put into blacklist, but it troubles me to code for future-proof.

stIncMale: > It is hard to validate it by testing case for the following possible HQL won't work due to Hibernate's semantic checking

This sounds like there is no issue at hand that needs to be solved, i.e., no change is needed.

NathanQingyangXu:
I close this PR for my original concern is the query literal bson value converion, but further discussion shows that concern is invalid. The other improvements are too minor to deserve a separate PR(I could include them in next normal PR, e.g.). For that reason I close this to avoid distraction.

PR #105, opened 2025-06-12
Hibernate-48 investigation

NathanQingyangXu:
We need to straighten out some doubts for Hibernate-48. This DRAFT PR will contain various commits to showcase the validation results.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/embeddable/EmptyStructAggregateRetrievalIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/embeddable/EmptyStructAggregateRetrievalIntegrationTests.java
@@ -1,0 +1,53 @@
+package com.mongodb.hibernate.embeddable;
+
+
+import com.mongodb.hibernate.junit.MongoExtension;
+import jakarta.persistence.Embeddable;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import org.hibernate.annotations.Struct;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.hibernate.testing.orm.junit.ServiceRegistry;
+import org.hibernate.testing.orm.junit.SessionFactory;
+import org.hibernate.testing.orm.junit.SessionFactoryScope;
+import org.hibernate.testing.orm.junit.Setting;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import static com.mongodb.hibernate.MongoTestAssertions.assertEq;
+
+@SessionFactory(exportSchema = false)
+@DomainModel(annotatedClasses = {
+        EmptyStructAggregateRetrievalIntegrationTests.Book.class,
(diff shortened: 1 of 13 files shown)


src/integrationTest/java/com/mongodb/hibernate/embeddable/EmptyStructIntegrationTests.java
```diff
@@ -0,0 +1,48 @@
...
+class EmptyStructIntegrationTests {
+
+
+    @Test
+    void test(SessionFactoryScope scope) {
+        scope.inTransaction(session -> {
+            var holder = new StructHolder();
+            holder.id = 1;
+            holder.emptyStruct = new EmptyStruct();
+            session.persist(holder);
```
vbabanin, line 31: What is the produced MQL result here?

src/main/java/com/mongodb/hibernate/internal/extension/MongoAdditionalMappingContributor.java
```diff
@@ -58,17 +58,15 @@ public void contribute(
...
             setIdentifierColumnName(persistentClass);
         });
-    }
-
-    private static void forbidDynamicInsert(PersistentClass persistentClass) {
-        if (persistentClass.useDynamicInsert()) {
-            throw new FeatureNotSupportedException(format("%s is not supported", DynamicInsert.class.getSimpleName()));
-        }
+        metadata.visitRegisteredComponents(component -> {
+            if (component.getStructName() != null && component.getProperties().isEmpty()) {
```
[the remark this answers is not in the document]
NathanQingyangXu: no, the above only applies to STRUCT (STRUCT must have name so telling whether the struct name is present suffices).
If we want to enlarge the forbidding to any `@Embeddable`, I think we simply drop the struct name checking above as below:
```
if (component.getProperties().isEmpty()) {
    ... ...
}
```

src/integrationTest/java/com/mongodb/hibernate/embeddable/EmptyStructAggregateRetrievalIntegrationTests.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    void testEmptyStructAggregateRetriedAsNonNull(SessionFactoryScope scope) {
+        var book = new Book();
+        book.id = 1;
+        book.author = new Author();
+
+        scope.inTransaction(session -> session.persist(book));
+
+        var retrievedBook = scope.fromTransaction(session -> session.get(Book.class, 1));
+        assertEq(book, retrievedBook);
```
stIncMale, line 37: My understanding is that this test requires `hibernate.create_empty_composites.enabled=true`. If that's true, then it does not ["showcase empty struct could be fetched intact (not null)"](https://github.com/mongodb/mongo-hibernate/pull/105/commits/7032c20b14d0dd12bd1ce24e71676ff676e078fc), as we won't use this configuration property knowing it does not exist in Hibernate ORM 7.0.

We should stop bringing up `hibernate.create_empty_composites.enabled` to avoid confusion.

PR #153, opened 2025-12-12
HIBERNATE-140: Run tests on JDK 17 and JDK 25

vbabanin:
HIBERNATE-140


.evergreen/java-config.sh
```diff
@@ -3,6 +3,7 @@
 # Java configurations for evergreen
 
 export JDK17="/opt/java/jdk17"
+export JDK25="/opt/java/jdk25"
```
stIncMale: Why do we need the `JDK25` variable? It is not used by anything in our codebase. There is still a chance that it is used by a subsequently executed command, because `export` marks a [_"`name` to be passed to subsequently executed commands in the environment"_](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-export), but I don't have reasons to believe that is what happens in our case.

.evergreen/config.yml
```diff
@@ -234,25 +246,26 @@ buildvariants:
...
-    display_name: "Unit Tests"
+    display_name: "Unit Tests ${jdk}"
     run_on: rhel80-medium
     tasks:
       - name: run-unit-tests
 
   - matrix_name: run-smoke-tests
-    matrix_spec: { mongo-version: "8.0", topology: replicaset, os: linux }
-    display_name: "Smoke Tests"
+    matrix_spec: { mongo-version: "8.0", topology: replicaset, os: linux, jdk: "jdk17" }
```
stIncMale: Let's also run smoke tests on JDK 25.


```suggestion
    matrix_spec: { mongo-version: "8.0", topology: replicaset, os: linux, jdk: "*" }
```

.evergreen/java-config.sh
```diff
@@ -3,6 +3,7 @@
...
 
 if [ -d "$JDK17" ]; then
   export JAVA_HOME=$JDK17
```
stIncMale, line 10: 1. I doubt we need the `JAVA_HOME` variable, as I don't see us using it anywhere. And Gradle is capable of [auto-detecting locations of installed JDKs](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection).
2. What's worse, is we always set `JAVA_HOME` to `/opt/java/jdk17`, if the directory exists, which is not what we want in this PR.

I suspect, we can and should just remove `java-config.sh‎`.

.evergreen/config.yml
```diff
@@ -234,25 +246,26 @@ buildvariants:
...
-  - name: unit-tests
+  - matrix_name: unit-tests
+    matrix_spec: { jdk: "*" }
     tags:
       - pr
-    display_name: "Unit Tests"
+    display_name: "Unit Tests ${jdk}"
     run_on: rhel80-medium
     tasks:
       - name: run-unit-tests
```
stIncMale, line 262: This does not seem to work. The [Unit Tests JDK25](https://spruce.corp.mongodb.com/variant-history/mongo-hibernate/unit-tests__jdk~jdk25) build variant [logs](https://parsley.corp.mongodb.com/evergreen/mongo_hibernate_unit_tests__jdk~jdk25_run_unit_tests_patch_199429997aa5397e1dd32c258f3624af8ac4754c_693b96ee7c37550007ec4f16_25_12_12_04_15_44/0/task?bookmarks=0%2C2351&selectedLineRange=L61-L67&shareLine=61)

```
2025/12/11 21:19:51.647] ++ export JAVA_HOME=/opt/java/jdk17
[2025/12/11 21:19:51.647] ++ JAVA_HOME=/opt/java/jdk17
[2025/12/11 21:19:51.647] ++ export JAVA_VERSION=17
[2025/12/11 21:19:51.647] ++ JAVA_VERSION=17
[2025/12/11 21:19:51.647] ++ echo 'Java Configs:'
[2025/12/11 21:19:51.647] ++ echo 'Java Home: /opt/java/jdk17'
[2025/12/11 21:19:51.647] ++ echo 'Java test version: 17'
```

[and](https://parsley.corp.mongodb.com/evergreen/mongo_hibernate_unit_tests__jdk~jdk25_run_unit_tests_patch_199429997aa5397e1dd32c258f3624af8ac4754c_693b96ee7c37550007ec4f16_25_12_12_04_15_44/0/task?bookmarks=0%2C2351&selectedLineRange=L2037&shareLine=2037)

```
[2025/12/11 21:21:18.654] Running tests using using JDK17
```

While `JAVA_HOME` is expected to be broken (I commented on it in [VAKOTODO "I doubt we need the"]()), having incorrect `JAVA_VERSION` is surprising, as `java-config.sh` sets it only if it is unset or null (see the [`:-` parameter expansion](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion-1)).

Also, you may see [here](https://parsley.corp.mongodb.com/evergreen/mongo_hibernate_unit_tests__jdk~jdk25_run_unit_tests_patch_199429997aa5397e1dd32c258f3624af8ac4754c_693b96ee7c37550007ec4f16_25_12_12_04_15_44/0/task?bookmarks=0%2C62%2C2351&selectedLineRange=L690-L716&shareLine=690) that Gradle emits all auto-detected JDKs/JREs, and JDK 25 does no appear to be installed. We may need to ask someone to do the changes necessary so that it becomes available (I don't know what the process is).

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package project
+
+plugins {
+    `java-library`
+}
+
+logger.info("Compiling ${project.name} using JDK${RELEASE_JAVA_VERSION}")
```
stIncMale: Let's say "Building" instead of "Compiling"? Because really, compiling is not the only thing that happens: there are static checkers that are run, there is generation of API documentation, etc.

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    toolchain { languageVersion = JavaLanguageVersion.of(RELEASE_JAVA_VERSION) }
+}
+
+tasks.withType<JavaCompile> {
+    options.release.set(RELEASE_JAVA_VERSION)
+}
+
+tasks.withType<Test>().configureEach {
+    val testJavaVersion: Int = (findProperty("javaVersion") as String?)?.toInt() ?: RELEASE_JAVA_VERSION
+    logger.info("Running tests using using JDK${testJavaVersion}")
```
stIncMale:
```suggestion
    logger.info("Running tests using JDK${testJavaVersion}")
```

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
```
stIncMale: The year here is incorrect for this codebase. Please change your IDE settings to avoid similar issues in the future.

```suggestion
 * Copyright 2026-present MongoDB, Inc.
```

buildSrc/src/main/kotlin/project/Companion.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
```
stIncMale:
```suggestion
 * Copyright 2026-present MongoDB, Inc.
```

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+
+tasks.withType<Test>().configureEach {
+    val testJavaVersion: Int = (findProperty("javaVersion") as String?)?.toInt() ?: RELEASE_JAVA_VERSION
+    logger.info("Running tests using using JDK${testJavaVersion}")
+    javaLauncher.set(
+        javaToolchains.launcherFor {
+            languageVersion.set(JavaLanguageVersion.of(testJavaVersion))
+        }
+    )
+}
```
stIncMale: Let's add the missing `LINE FEED (LF)` character at the end of this file.

stIncMale (changes requested):
The last reviewed commit is ad6f26ac40c6199cd916cc7c9cf2be2e1d48f5cd.

vbabanin:
//TODO add assertion in test execution environment that env variable for Java version from CI is equals to current Java runtime version.

.evergreen/java-config.sh
```diff
@@ -2,12 +2,8 @@
 
 # Java configurations for evergreen
 
-export JDK17="/opt/java/jdk17"
-
-if [ -d "$JDK17" ]; then
-  export JAVA_HOME=$JDK17
-fi
-
+# JAVA_HOME is required to launch Gradle itself (Gradle 9+ requires JDK 17+).
```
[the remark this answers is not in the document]
vbabanin: The [ -d ] check is restored so the depended script could be excecuted locally for debugging. Without it, `JAVA_HOME` might get set to a nonexistent `/opt/java/jdk17` and gradlew dies with `"JAVA_HOME is set to an invalid directory"`.  With the check, `JAVA_HOME` stays unset and gradlew falls back to whatever JDK is on their `PATH`. On CI, if `/opt/java/jdk17` is missing, Gradle already gives a clear error: ```"Gradle requires JVM 17 or later to run```" - so the check doesn't need to add a custom error message there.

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+tasks.withType<JavaCompile> {
+    options.release.set(RELEASE_JAVA_VERSION)
+}
+
+tasks.withType<Test>().configureEach {
+    val testJavaVersion: Int = (findProperty("javaVersion") as String?)?.toInt() ?: RELEASE_JAVA_VERSION
+    logger.info("Running tests using JDK${testJavaVersion}")
+    javaLauncher.set(
+        javaToolchains.launcherFor {
+            languageVersion.set(JavaLanguageVersion.of(testJavaVersion))
```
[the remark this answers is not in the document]
vbabanin: I don't think this is an issue:
- If not located build will fail and we will correct it (which is highly unlikely due to the next point below)
- The JDKs are located in default Linux path recognizable by Gradle.

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+java {
+    toolchain { languageVersion = JavaLanguageVersion.of(RELEASE_JAVA_VERSION) }
+}
+
+tasks.withType<JavaCompile> {
+    options.release.set(RELEASE_JAVA_VERSION)
+}
+
+tasks.withType<Test>().configureEach {
+    val testJavaVersion: Int = (findProperty("javaVersion") as String?)?.toInt() ?: RELEASE_JAVA_VERSION
```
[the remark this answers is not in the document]
vbabanin: It makes sense to provide better error message, however, it is not critical. I've added a descriptive error message using `Provider.map()` and `.orElse()` which is more concise than the suggested approach.

buildSrc/src/main/kotlin/project/Constants.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package project
+
+const val RELEASE_JAVA_VERSION = 17
```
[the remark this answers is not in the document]
vbabanin: I have renamed the file to Constants.kt. If later we want to add more logic to it, we can rename it.

.evergreen/java-config.sh
vbabanin: `JAVA_HOME` isn’t used by the build logic once Gradle is running - agreed. Gradle toolchains handle JDK selection for compilation/tests.

However, `JAVA_HOME` is used by gradlew to decide which JVM launches the Gradle daemon. On `rhel80` the jdk in the `PATH` is `JDK 8 (/usr/lib/jvm/java-1.8.0-openjdk)`, and Gradle 9+ won’t start on JDK 8 (“Gradle requires JVM 17 or later…”). So setting `JAVA_HOME` here is only to bootstrap Gradle itself on CI.

There are other ways to do this (`e.g., org.gradle.java.home=/opt/java/jdk17`), but that can affect local builds. Using `JAVA_HOME` in separate script looks cleaner.

Regarding removing `java-config.sh`: we could remove it, but I’d prefer to keep it to avoid duplicating the same bootstrap logic (setting `JAVA_HOME` when `JDK17` is available, `JAVA_VERSION` default, and diagnostic output) across the 5 CI scripts (static-checks, run-unit-tests, run-smoke-tests, run-integration-tests, publish). With a shared script, each one just needs `source java-config.sh`.

I currently added comments in `java-config.sh` to make the purpose of `JAVA_HOME` explicit.

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -0,0 +1,51 @@
...
+tasks.withType<Test>().configureEach {
+    logger.info("Running tests using JDK${javaVersion}")
+    javaLauncher.set(runtimeJavaLauncher)
+}
+
+// JavaExec tasks don't inherit the project toolchain automatically (unlike JavaCompile, Test, Javadoc).
+// Without this, they would silently use the Gradle daemon JDK instead of the intended version.
+tasks.withType<JavaExec>().configureEach {
+    javaLauncher.set(runtimeJavaLauncher)
+}
```
vbabanin, line 51: Unlike `Test`, `JavaCompile`, and `Javadoc`, `JavaExec` tasks don’t inherit the project toolchain - they silently fall back to the Gradle daemon JDK. That’s not obvious and can be easy to miss, so wiring the toolchain manually in each future JavaExec usage is error-prone. Adding this now is a low-cost guardrail that prevents that class of mistake upfront.

.evergreen/config.yml
vbabanin: Good catch. The Evergreen expansion was not being passed through as an environment variable, and I’ve corrected that. 

I opened [DEVPROD-31226](https://jira.mongodb.org/browse/DEVPROD-31226) to install JDK 25, and the deployment has since been completed. 

The tests are now passing and running on JDK 25.

jyemin (commented):
Looks like the unit tests are failing on Java 25 so this isn't quite ready to merge.

jyemin:
This is no longer mergeable, so closing until this is re-prioritized

PR #158, opened 2026-05-20
HIBERNATE-144: upgrade hibernate-orm from 6.6 to 7.3

jyemin:
Upgrades Hibernate ORM from 6.6.34.Final to 7.3.4.Final. The changes address eight categories of
breaking-API changes introduced across the Hibernate 7 release line.

### 1. SPI/implementation separation in `sql.exec`

Hibernate 7 completed a cleanup separating stable public SPI interfaces from internal
implementations in `org.hibernate.sql.exec`.

`JdbcOperationQuerySelect` had served as both the interface and the implementation. It is now split:
`JdbcSelect` is the stable SPI interface (stays in `.spi`), and `JdbcOperationQuerySelect` became
the concrete class (moved to `.internal`). `JdbcOperationQueryDelete`, `JdbcOperationQueryUpdate`,
and `AbstractJdbcOperationQuery` similarly moved from `.spi` to `.internal`.

`SelectMqlTranslator` and `MongoTranslatorFactory` now use `JdbcSelect` as the type parameter —
the correct SPI boundary.

### 2. Package reorganization for query types

- `FetchClauseType` moved from `org.hibernate.query.sqm` to `org.hibernate.query.common`,
  reflecting that it is not SQM-specific.
- `org.hibernate.query.NullPrecedence` was replaced by `jakarta.persistence.criteria.Nulls`,
  introduced in Jakarta Persistence 3.2, standardizing null-ordering across JPA providers.

### 3. Aggregate/struct type API reorganization

`MongoStructJdbcType` implements `@Struct` aggregate-embeddable support. Three related API
changes:

- `StructJdbcType` was renamed to `StructuredJdbcType`.
- `StructHelper` and `StructAttributeValues` moved from `org.hibernate.dialect` to
  `org.hibernate.type.descriptor.jdbc`, collocating them with the types they serve.
- `StructHelper.instantiate()` no longer takes a `SessionFactory` argument.

Hibernate 7's `AggregateComponentSecondPass` also now gates `@Struct` mapping acceptance on
`Dialect.supportsUserDefinedTypes()`. `MongoDialect` overrides this to return `true`.

### 4. New `SqlAstTranslator` interface methods

Hibernate 7 added four abstract methods to `SqlAstTranslator`:

- `getLiteralValue()` — extracts a literal value from an expression during translation.
- `renderNamedSetReturningFunction()` — renders table-valued / set-returning functions.
- `addAffectedTableName()` — tracks table names referenced by a mutation.
- `getSqlAst()` (`@Incubating`, 7.1) — exposes the input AST from the translator.

All four are stubbed to throw `FeatureNotSupportedException`. Our MQL translator does not support
the corresponding Hibernate features and should never reach them.

### 5. Dialect bootstrap path change

In Hibernate 6, `JdbcEnvironmentInitiator` called the no-arg `Dialect` constructor only as an
error fallback when the `DialectResolutionInfo`-taking constructor threw. We exploited this by
having `MongoDialect()` always throw a `RuntimeException` with an explanatory message.

Hibernate 7's `JdbcEnvironmentInitiator.getJdbcEnvironmentWithDefaults()` calls the no-arg
constructor as a **primary** code path — to build a default `JdbcEnvironment` when JDBC metadata
is unavailable (e.g., Hibernate Testing Framework bootstrap with `@ServiceRegistry` and no live
connection). The constructor must now succeed. `MongoDialect()` now delegates to
`super(MINIMUM_DBMS_VERSION)`.

`NativeBootstrappingIntegrationTests` is deleted; it tested that a `buildInfo` failure surfaced a
specific error via the throwing no-arg constructor — a scenario that no longer exists.

### 6. `ObjectIdGenerator` constructor unification and dead-code removal

Hibernate 6 used two distinct creation-context types: `CustomIdGeneratorCreationContext` for
`@IdGeneratorType` (identifier generators) and `GeneratorCreationContext` for `@ValueGenerationType`
(non-identifier generators). `ObjectIdGenerator` had one constructor for each, using the type
difference to set a `forIdentifier` flag that drove a workaround in `generate()`: when `currentValue`
was null, it called `getIdentifier()` explicitly on the assumption that Hibernate passed null for
pre-assigned identifiers.

Hibernate 7 removes `CustomIdGeneratorCreationContext` entirely; both generator types now receive
`GeneratorCreationContext`. With no way to set `forIdentifier = true`, the `forIdentifier` field
and its associated branch become dead code. Investigation confirmed the branch body was already
dead in Hibernate 6.6: `AbstractSaveEventListener` already passes `persister.getIdentifier(entity)`
as `currentValue` when `allowAssignedIdentifiers()` returns `true`, so the explicit `getIdentifier()`
call inside the branch always returned the same null as `currentValue`. The full test suite passes
with both the field and the branch removed.

### 7. JDBC boot-time metadata expansion

Hibernate 7 refactored `ExtractedDatabaseMetaDataImpl` from a lazy builder to a constructor that
eagerly queries all JDBC metadata at bootstrap. Several `DatabaseMetaData` methods that were never
reached in Hibernate 6 are now called:

- `supportsSchemasInDataManipulation()` / `supportsCatalogsInDataManipulation()` → `false`.
- `getURL()` → `""` (`MongoDatabaseMetaData` has no access to the connection URL string).
- `getDefaultTransactionIsolation()` → `TRANSACTION_NONE`.
- `storesMixedCaseIdentifiers()` / `storesMixedCaseQuotedIdentifiers()` → `true`; all
  `storesUpperCase*` and `storesLowerCase*` → `false`. Queried by `IdentifierHelperBuilder` to
  configure identifier-case policy; MongoDB preserves case verbatim.

`MongoConnection.getTransactionIsolation()` was added because `ExtractedDatabaseMetaDataImpl` reads
it directly from the connection. `MongoConnectionProvider.MongoDatabaseConnectionInfo` gained five
new methods required by the updated `DatabaseConnectionInfo` interface contract.

### 8. `DriverManagerConnectionProvider` connection eagerness

`NativeBootstrappingTests` had used `INITIAL_SIZE=0` to prevent `DriverManagerConnectionProvider`
from opening a real JDBC connection during test bootstrap. Hibernate 7 opens a connection eagerly
in `configure()`, making that setting ineffective. The test now registers a mock
`ConnectionProvider` instead.


src/main/java/com/mongodb/hibernate/internal/translate/SelectMqlTranslator.java
```diff
@@ -30,13 +30,14 @@
...
-import org.hibernate.sql.exec.spi.JdbcOperationQuerySelect;
+import org.hibernate.sql.exec.internal.JdbcOperationQuerySelect;
 import org.hibernate.sql.exec.spi.JdbcParameterBinder;
 import org.hibernate.sql.exec.spi.JdbcParameterBindings;
+import org.hibernate.sql.exec.spi.JdbcSelect;
 import org.hibernate.sql.results.jdbc.spi.JdbcValuesMappingProducerProvider;
 import org.jspecify.annotations.Nullable;
 
-final class SelectMqlTranslator extends AbstractMqlTranslator<JdbcOperationQuerySelect> {
+final class SelectMqlTranslator extends AbstractMqlTranslator<JdbcSelect> {
```
jyemin, line 44: Hibernate 7 split `JdbcOperationQuerySelect` into a public SPI interface (`JdbcSelect`, in `.spi`) and a concrete implementation (`JdbcOperationQuerySelect`, moved to `.internal`). We now bind to the interface, which is the stable API surface.

src/main/java/com/mongodb/hibernate/internal/translate/MutationMqlTranslator.java
```diff
@@ -31,10 +31,10 @@
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.query.spi.QueryOptions;
 import org.hibernate.sql.ast.tree.MutationStatement;
+import org.hibernate.sql.exec.internal.JdbcOperationQueryDelete;
```
jyemin, line 34: Unlike the select case, Hibernate 7 did not extract new SPI interfaces for delete and update operations — the concrete classes simply moved from `.spi` to `.internal`. We instantiate them directly here because `MutationMqlTranslator` builds the result, not just declares the type parameter.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -48,7 +48,7 @@
 import static com.mongodb.hibernate.internal.translate.mongoast.filter.AstLogicalFilterOperator.NOR;
 import static com.mongodb.hibernate.internal.translate.mongoast.filter.AstLogicalFilterOperator.OR;
 import static java.lang.String.format;
-import static org.hibernate.query.sqm.FetchClauseType.ROWS_ONLY;
+import static org.hibernate.query.common.FetchClauseType.ROWS_ONLY;
```
jyemin, line 51: `FetchClauseType` was moved from `query.sqm` to `query.common` in Hibernate 7; it applies to HQL and criteria queries alike, not just SQM-generated queries.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -238,7 +238,23 @@ public void render(SqlAstNode sqlAstNode, SqlAstNodeRenderingMode sqlAstNodeRend
     }
 
     @Override
-    public boolean supportsFilterClause() {
+    @SuppressWarnings("TypeParameterUnusedInFormals")
+    public <X> X getLiteralValue(Expression expression) {
```
jyemin, line 247: Hibernate 7 added these four abstract methods to `SqlAstTranslator`. All are safe to stub because none are called on our translator during normal MongoDB query translation:

- `getLiteralValue` — called only by dialect-specific XML-table and JSON-path function renderers (DB2, HANA, Oracle, PostgreSQL, SQL Server) and from within `AbstractSqlAstTranslator`'s private helpers. We don't extend `AbstractSqlAstTranslator` and register no such functions.
- `getSqlAst` — `@Incubating` (7.1); has zero callers anywhere in the Hibernate codebase outside of `AbstractSqlAstTranslator` itself.
- `renderNamedSetReturningFunction` — invoked only by `NamedSqmSetReturningFunctionDescriptor` when a set-returning (table-valued) function appears in a query. We register none.
- `addAffectedTableName` — called only from `AbstractSqlAstTranslator.visitSetAssignment()`, an internal method we never invoke. It uses `fail()` (AssertionError) rather than `FeatureNotSupportedException` because reaching it would indicate a bug in our translator, not an unsupported user feature.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -621,10 +642,10 @@ public void visitSqlSelectionExpression(SqlSelectionExpression sqlSelectionExpre
     @Override
     public void visitSortSpecification(SortSpecification sortSpecification) {
         var nullPrecedence = sortSpecification.getNullPrecedence();
-        if (nullPrecedence == null || nullPrecedence == NullPrecedence.NONE) {
+        if (nullPrecedence == null || nullPrecedence == Nulls.NONE) {
```
jyemin, line 652: Jakarta Persistence 3.2 standardized null-ordering as `jakarta.persistence.criteria.Nulls`, replacing Hibernate's proprietary `org.hibernate.query.NullPrecedence`. `Nulls.NONE` carries the same meaning as the old `NullPrecedence.NONE`: no explicit null-ordering preference was given, defer to the database default.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -52,15 +51,16 @@
...
+import org.hibernate.type.descriptor.jdbc.StructuredJdbcType;
 import org.jspecify.annotations.Nullable;
 
 /**
  * Thread-safe.
  *
  * @hidden
  */
-public final class MongoStructJdbcType implements StructJdbcType {
+public final class MongoStructJdbcType implements StructuredJdbcType {
```
jyemin, line 63: `StructJdbcType` was renamed to `StructuredJdbcType` in Hibernate 7. `StructHelper` and `StructAttributeValues` moved from `org.hibernate.dialect` to `org.hibernate.type.descriptor.jdbc` at the same time, collocating the helper utilities with the type hierarchy they serve.

src/main/java/com/mongodb/hibernate/internal/type/MongoStructJdbcType.java
```diff
@@ -276,10 +276,7 @@ public MongoStructJdbcType getJdbcType() {
             } else {
                 var embeddableMappingType = getEmbeddableMappingType();
                 assertTrue(classX.equals(embeddableMappingType.getJavaType().getJavaTypeClass()));
-                result = instantiate(
-                        embeddableMappingType,
-                        new StructAttributeValues(jdbcValues.length, jdbcValues),
-                        options.getSessionFactory());
+                result = instantiate(embeddableMappingType, new StructAttributeValues(jdbcValues.length, jdbcValues));
```
jyemin, line 279: `StructHelper.instantiate()` no longer takes a `SessionFactory` in Hibernate 7; the factory was removed from the signature after the method was refactored to obtain its context through the mapping type directly.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -160,18 +160,12 @@ public MongoDialect(DialectResolutionInfo info) {
...
-     *
-     * @deprecated Exists only to avoid the confusing {@link NoSuchMethodException} thrown by Hibernate ORM when
-     *     {@link MongoDialect#MongoDialect(DialectResolutionInfo)} fails.
-     * @throws RuntimeException Always.
+     * Hibernate 7's {@link org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator JdbcEnvironmentInitiator}
+     * invokes this no-arg form when it builds a default {@code JdbcEnvironment} (test bootstrap without a real JDBC
+     * connection). The {@link DialectResolutionInfo} variant is still preferred when JDBC metadata is available.
      */
-    @Deprecated
     public MongoDialect() {
```
jyemin, line 174: In Hibernate 6, this constructor was called only when the `DialectResolutionInfo`-taking constructor threw, so making it throw an explanatory message was safe. Hibernate 7's `JdbcEnvironmentInitiator.getJdbcEnvironmentWithDefaults()` calls the no-arg constructor as a primary path (no JDBC connection available, e.g., `@ServiceRegistry` tests in the Hibernate Testing Framework), so it must succeed. We pass `MINIMUM_DBMS_VERSION` as the version hint so the dialect is initialized consistently regardless of which constructor path is taken.

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
```diff
@@ -248,6 +242,13 @@ public boolean supportsStandardArrays() {
         return true;
     }
 
+    // Hibernate 7's `AggregateComponentSecondPass` rejects `@Struct` mappings unless this returns `true`.
+    // MQL supports embedded documents as user-defined struct types via `MongoStructJdbcType`.
+    @Override
+    public boolean supportsUserDefinedTypes() {
```
jyemin, line 257: Without this override, `AggregateComponentSecondPass` silently rejects all `@Struct` (aggregate embeddable) mappings during bootstrap. The check was added in Hibernate 7 as a gate to ensure dialects explicitly opt in to user-defined type support. Our existing `MongoStructJdbcType` implementation already handles the full struct lifecycle, so `true` is correct.

src/main/java/com/mongodb/hibernate/internal/id/objectid/ObjectIdGenerator.java
```diff
@@ -40,26 +37,6 @@ public final class ObjectIdGenerator implements BeforeExecutionGenerator {
...
-            GeneratorCreationContext context) {
-        this(false);
-    }
-
-    private ObjectIdGenerator(boolean forIdentifier) {
-        this.forIdentifier = forIdentifier;
-    }
-
     @Override
     public Object generate(
```
jyemin, line 41: `CustomIdGeneratorCreationContext` is removed in Hibernate 7, so removed that constructor, but since both constructor have empty bodies, and Hibernate also looks for a no-args constructor, I just removed them both.

src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
```diff
@@ -106,6 +106,67 @@ public boolean supportsCatalogsInTableDefinitions() {
         return false;
     }
 
+    // Hibernate 7's `ExtractedDatabaseMetaDataImpl` queries these (and a handful of others below) when it builds
+    // the boot-time metadata snapshot. The 6.x bootstrap path did not, so they had remained unimplemented.
+    @Override
+    public boolean supportsSchemasInDataManipulation() {
```
jyemin, line 113: `ExtractedDatabaseMetaDataImpl` was refactored in Hibernate 7 from a lazy builder (which populated fields on demand) to an eager constructor that queries all relevant `DatabaseMetaData` methods up front during bootstrap. These methods were previously dead code in our adapter; they are now called on every boot. The values are straightforward: MongoDB has no schemas or catalogs, no JDBC-style isolation levels, and no access to the original connection URL from within `MongoDatabaseMetaData`.

src/main/java/com/mongodb/hibernate/jdbc/MongoDatabaseMetaData.java
```diff
@@ -106,6 +106,67 @@ public boolean supportsCatalogsInTableDefinitions() {
...
+
+    @Override
+    public int getDefaultTransactionIsolation() {
+        return Connection.TRANSACTION_NONE;
+    }
+
+    // Hibernate 7's `IdentifierHelperBuilder` queries the four `stores*CaseIdentifiers` variants when initializing
+    // its identifier-case policy. MongoDB preserves identifier case verbatim, so we report "mixed case" only.
+    @Override
+    public boolean storesUpperCaseIdentifiers() {
```
jyemin, line 140: `IdentifierHelperBuilder` uses these flags to decide how to normalize quoted and unquoted identifiers. Returning `storesMixedCaseIdentifiers() == true` (and all others `false`) tells Hibernate that MongoDB is case-sensitive and preserves identifiers exactly as written — which is correct for both collection names and field names.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnection.java
```diff
@@ -194,6 +194,14 @@ public DatabaseMetaData getMetaData() throws SQLException {
         return null;
     }
 
+    // Hibernate 7's `ExtractedDatabaseMetaDataImpl` queries this when building boot-time JDBC metadata.
+    // MongoDB has no JDBC-style transaction isolation level, so report TRANSACTION_NONE.
+    @Override
+    public int getTransactionIsolation() throws SQLException {
```
jyemin, line 200: `ExtractedDatabaseMetaDataImpl` now reads `getTransactionIsolation()` directly from the `Connection` object, in addition to `DatabaseMetaData.getDefaultTransactionIsolation()`. Both must return consistently. MongoDB client sessions use snapshot isolation internally, but there is no corresponding JDBC isolation constant, so `TRANSACTION_NONE` is the least-misleading value.

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
```diff
@@ -187,6 +187,31 @@ public Integer getPoolMaxSize() {
                     .getMaxSize();
         }
 
+        @Override
+        public @Nullable String getSchema() {
```
jyemin, line 196: Hibernate 7 added these abstract methods to the `DatabaseConnectionInfo` interface. MongoDB has no schema or catalog concepts and no notion of a JDBC fetch size at the connection-info level, so all return null/false.

src/test/java/com/mongodb/hibernate/boot/NativeBootstrappingTests.java
```diff
@@ -88,7 +87,12 @@ void testMongoAdditionalMappingContributorIsSkipped() {
                         .applySetting(DIALECT, "org.hibernate.dialect.PostgreSQLDialect")
                         .applySetting(JAKARTA_JDBC_URL, "jdbc:postgresql://host/")
                         .applySetting(ALLOW_METADATA_ON_BOOT, false)
-                        .applySetting(DriverManagerConnectionProviderImpl.INITIAL_SIZE, 0)
+                        // Hibernate 7's `DriverManagerConnectionProvider` eagerly opens a connection during
```
jyemin, line 90: In Hibernate 6, `INITIAL_SIZE=0` deferred connection creation long enough that our test could bootstrap Hibernate with a PostgreSQL URL but no actual PostgreSQL server. Hibernate 7's `DriverManagerConnectionProvider.configure()` eagerly opens a connection regardless of `INITIAL_SIZE`. Plugging in a mock `ConnectionProvider` achieves the same effect: Hibernate accepts it and never attempts a real JDBC connection.

src/integrationTest/java/com/mongodb/hibernate/type/temporal/CalendarIntegrationTests.java
```diff
@@ -72,7 +71,6 @@ void unsupported() {
     static void assertNotSupported(Class<?> entityClass) {
         assertThatThrownBy(() ->
```
jyemin, line 64: Hibernate 7 rejects all generic-typed entities with `AnnotationException` at annotation-processing time ("has an unbound type and no explicit target entity") before our `TypeContributor` runs. The assertion is updated to reflect the exception that is actually thrown on 7.x.

gradle/libs.versions.toml
```diff
@@ -26,7 +26,7 @@ assertj = "3.27.6"
 google-errorprone-core = "2.42.0"
 nullaway = "0.12.14"
 jspecify = "1.0.0" # Remember to update javadoc links
-hibernate-orm = "6.6.38.Final" # Remember to update javadoc links
+hibernate-orm = "7.3.4.Final" # Remember to update javadoc links
```
vbabanin:
```suggestion
hibernate-orm = "7.3.5.Final" # Remember to update javadoc links
```

src/main/java/com/mongodb/hibernate/dialect/MongoDialect.java
vbabanin: I dug into the Hibernate ORM bootstrap paths. In 6.6, `getJdbcEnvironmentWithDefaults()` also ends up calling the no-arg dialect constructor, but only as a fallback. It looks like this area hasn’t materially changed in 7.3, and the behavior is mostly the same.

### When the no-arg constructor gets used (7.3)

1) **Fallback after failure in `getJdbcEnvironmentUsingJdbcMetadata()`**, e.g.
- no `ConnectionProvider` is configured (or it throws), or
- `MongoDialect(DialectResolutionInfo)` throws during dialect resolution.

In these cases, Hibernate catches and falls back to `getJdbcEnvironmentWithDefaults()`, which instantiates the dialect via the no-arg constructor.

2) **No `ConnectionProvider` + `dialect` not specified**: Hibernate calls `dialectFactory.buildDialect(config, null)` first; that fails with `HibernateException("Unable to determine Dialect")` before the no-arg fallback path can resolve anything.

### What Hibernate documents as the intended contract (ORM 7.3)

Hibernate’s javadoc states:
> A subclass must provide a public constructor with a single parameter of type DialectResolutionInfo. Alternatively, for purposes of backward compatibility with older versions of Hibernate, a constructor with no parameters is also allowed.

So  `Dialect(DialectResolutionInfo)` appears to be the primary constructor; the no-arg constructor exists for backward compatibility.

### Why failing fast matters here

If the version mismatch isn’t surfaced as a hard error, Hibernate can recover via the fallback path and instantiate the dialect using the no-arg constructor, which suppresses the version failure. Net effect: an application against a <7.0 server can keep working “by accident” even though the version gate was hit internally.

### What happens when we do throw

<details>
<summary> 1. WARN log with version mismatch + stack trace </summary>

```
2026-05-25T22:28:58,227 [Test worker] WARN org.hibernate.orm.jdbc - HHH100046: Could not obtain connection to query JDBC database metadata
  org.hibernate.boot.registry.selector.spi.StrategySelectionException: Could not instantiate named strategy class [com.mongodb.hibernate.dialect.MongoDialect]
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:182)
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:118)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.constructDialect(DialectFactoryImpl.java:122)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.buildDialect(DialectFactoryImpl.java:87)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator$1.execute(JdbcEnvironmentInitiator.java:398)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator$1.execute(JdbcEnvironmentInitiator.java:367)
        at org.hibernate.jdbc.WorkExecutor.executeReturningWork(WorkExecutor.java:56)
        at org.hibernate.jdbc.AbstractReturningWork.accept(AbstractReturningWork.java:32)
        at org.hibernate.resource.transaction.backend.jdbc.internal.JdbcIsolationDelegate.doWorkAndCommit(JdbcIsolationDelegate.java:89)
        at org.hibernate.resource.transaction.backend.jdbc.internal.JdbcIsolationDelegate.delegateWork(JdbcIsolationDelegate.java:64)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentUsingJdbcMetadata(JdbcEnvironmentInitiator.java:366)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironment(JdbcEnvironmentInitiator.java:143)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:120)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:80)
        at org.hibernate.boot.registry.internal.StandardServiceRegistryImpl.initiateService(StandardServiceRegistryImpl.java:133)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.createService(AbstractServiceRegistryImpl.java:260)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.initializeService(AbstractServiceRegistryImpl.java:235)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.getService(AbstractServiceRegistryImpl.java:212)
        at org.hibernate.boot.model.relational.Database.<init>(Database.java:44)
        at org.hibernate.boot.internal.InFlightMetadataCollectorImpl.getDatabase(InFlightMetadataCollectorImpl.java:251)
        at org.hibernate.boot.internal.InFlightMetadataCollectorImpl.<init>(InFlightMetadataCollectorImpl.java:203)
        at org.hibernate.boot.model.process.spi.MetadataBuildingProcess.complete(MetadataBuildingProcess.java:174)
        at org.hibernate.boot.model.process.spi.MetadataBuildingProcess.build(MetadataBuildingProcess.java:129)
        at org.hibernate.boot.internal.MetadataBuilderImpl.build(MetadataBuilderImpl.java:440)
        at org.hibernate.testing.orm.junit.DomainModelExtension.lambda$createDomainModelScope$0(DomainModelExtension.java:247)
        ... [JUnit/Gradle frames]
  Caused by: org.hibernate.boot.registry.selector.spi.StrategySelectionException: Could not instantiate named dialect class [com.mongodb.hibernate.dialect.MongoDialect]
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.lambda$constructDialect$0(DialectFactoryImpl.java:142)
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:178)
        ... 113 common frames omitted
  Caused by: java.lang.reflect.InvocationTargetException: null
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.lambda$constructDialect$0(DialectFactoryImpl.java:130)
        ... 114 common frames omitted
  Caused by: java.lang.RuntimeException: The minimum supported version of MongoDB is 7.0, but you are using 6.0.5
        at com.mongodb.hibernate.dialect.MongoDialect.checkVersion(MongoDialect.java:190)
        at org.hibernate.dialect.Dialect.<init>(Dialect.java:365)
        at com.mongodb.hibernate.dialect.MongoDialect.<init>(MongoDialect.java:162)
        ... 120 common frames omitted
```

</details>

<details>
<summary> 2. Bootstrap fails with exception from no-arg </summary>

```
  org.hibernate.service.spi.ServiceException: Unable to create requested service [org.hibernate.engine.jdbc.env.spi.JdbcEnvironment] due to: Could not instantiate named strategy class
  [com.mongodb.hibernate.dialect.MongoDialect]
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.createService(AbstractServiceRegistryImpl.java:273)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.initializeService(AbstractServiceRegistryImpl.java:235)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.getService(AbstractServiceRegistryImpl.java:212)
        at org.hibernate.boot.model.relational.Database.<init>(Database.java:44)
        at org.hibernate.boot.internal.InFlightMetadataCollectorImpl.getDatabase(InFlightMetadataCollectorImpl.java:251)
        at org.hibernate.boot.internal.InFlightMetadataCollectorImpl.<init>(InFlightMetadataCollectorImpl.java:203)
        at org.hibernate.boot.model.process.spi.MetadataBuildingProcess.complete(MetadataBuildingProcess.java:174)
        at org.hibernate.boot.model.process.spi.MetadataBuildingProcess.build(MetadataBuildingProcess.java:129)
        at org.hibernate.boot.internal.MetadataBuilderImpl.build(MetadataBuilderImpl.java:440)
        at org.hibernate.testing.orm.junit.DomainModelExtension.lambda$createDomainModelScope$0(DomainModelExtension.java:247)    
        ... [JUnit/stream frames]
  Caused by: org.hibernate.boot.registry.selector.spi.StrategySelectionException: Could not instantiate named strategy class [com.mongodb.hibernate.dialect.MongoDialect]
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:182)
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:118)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.constructDialect(DialectFactoryImpl.java:122)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.buildDialect(DialectFactoryImpl.java:87)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentWithDefaults(JdbcEnvironmentInitiator.java:200)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironmentUsingJdbcMetadata(JdbcEnvironmentInitiator.java:452)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.getJdbcEnvironment(JdbcEnvironmentInitiator.java:143)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:120)
        at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:80)
        at org.hibernate.boot.registry.internal.StandardServiceRegistryImpl.initiateService(StandardServiceRegistryImpl.java:133)
        at org.hibernate.service.internal.AbstractServiceRegistryImpl.createService(AbstractServiceRegistryImpl.java:260)
        ... 26 more
  Caused by: java.lang.RuntimeException: Could not instantiate [com.mongodb.hibernate.dialect.MongoDialect], see the earlier exceptions to find out why
        at com.mongodb.hibernate.dialect.MongoDialect.<init>(MongoDialect.java:172)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.ReflectAccess.newInstance(ReflectAccess.java:128)
        at java.base/java.lang.Class.newInstance(Class.java:645)
        at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.lambda$constructDialect$0(DialectFactoryImpl.java:138)
        at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:178)
        ... 36 more
```

</details>


This is noisy (multiple layers of wrapping), but it is fail-fast and prevents silently booting on an unsupported server version. This was discussed here:  https://github.com/mongodb/mongo-hibernate/pull/79#discussion_r2024782618

So I think the resolution-info constructor could remain the enforcement point, to ensure the fallback/no-arg path doesn’t bypass the version gate. 

What do you think?

**UPD: discussed**

src/main/java/com/mongodb/hibernate/jdbc/MongoConnectionProvider.java
vbabanin: It looks like we could use MongoConfiguration#databaseName in getCatalog() (and return no catalog), similar to the MySQL dialect. However, it is not critical.

MySql dialect:

connection.getCatalog()-> returns the current database name (e.g., "mydb")
connection.getSchema() -> returns null

****UPD:** After discussion, we decided to skip this for now since returning a non-null value could have side effects we haven’t validated yet.**

vbabanin (approved):
LGTM!

PR #160, opened 2026-05-22
HIBERNATE-145: Support inner and left outer joins

jyemin:
## Summary

Implements initial HQL inner and left outer join support for the MQL translator, translating HQL path-navigation joins (`JOIN c.orders o`) into MongoDB `$lookup` + `$unwind` aggregation pipeline stages.

- **`MongoResultSet`**: Support dotted-path column keys (e.g. `o1_0.total`) for reading nested BSON documents returned by joined queries
- **`AstLookupStage` / `AstUnwindStage`**: New MQL AST nodes for `$lookup` and `$unwind` pipeline stages, with unit tests
- **`AbstractMqlTranslator`**: Translate `TableGroupJoin` nodes into `$lookup`+`$unwind` stages; gate unsupported shapes with `FeatureNotSupportedException` (HIBERNATE-161–170)
- **Design spec**: `docs/superpowers/specs/2026-05-21-mql-join-support-design.md` documents the MongoDB mechanism, supported shapes, and implementation approach

## Supported shapes

- INNER JOIN and LEFT OUTER JOIN on a single equijoin predicate (`ON a.x = b.y`)
- Path-navigation joins (`JOIN c.orders o`) — entity joins (`JOIN Order o ON ...`) are not yet supported (HIBERNATE-168)

## Unsupported shapes (throw `FeatureNotSupportedException`)

| Shape | Ticket |
|---|---|
| RIGHT / FULL / CROSS join | HIBERNATE-161/162/163 |
| Compound ON condition (Junction) | HIBERNATE-164 |
| Non-equijoin ON condition | HIBERNATE-165 |
| Non-column ON expression | HIBERNATE-166 |
| Subquery join | HIBERNATE-167 |
| Entity-syntax join / association-navigation nested join | HIBERNATE-168 |
| Non-`@Struct` embeddable element collection root | HIBERNATE-169 |
| Same-table ON condition | HIBERNATE-170 |
| Function table reference join | HIBERNATE-111 |

## Test plan

- [ ] `AstLookupStageTests` / `AstUnwindStageTests` — unit tests for new AST nodes
- [ ] `MongoResultSetTests` — unit tests for dotted-path nested key resolution
- [ ] `JoinSelectQueryIntegrationTests` — integration tests covering INNER JOIN, LEFT OUTER JOIN, JOIN FETCH (both directions), and all negative `FeatureNotSupportedException` cases
- [ ] Existing integration test suite passes (no regressions in `ArrayAndCollectionIntegrationTests`, etc.)


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -572,7 +581,24 @@ public void visitColumnReference(ColumnReference columnReference) {
         if (columnReference.isColumnExpressionFormula()) {
             throw new FeatureNotSupportedException("Formula is not supported");
         }
-        astVisitorValueHolder.yield(FIELD_PATH, columnReference.getColumnExpression());
+        astVisitorValueHolder.yield(FIELD_PATH, resolveFieldPath(columnReference));
+    }
+
+    private String resolveFieldPath(ColumnReference columnReference) {
```
jyemin, line 627: Columns from joined collections are projected as sub-documents keyed by the join alias (e.g. `o1_0.total` rather than `total`), matching the nested BSON structure produced by `$unwind`. Columns from the driving (root) table use bare field names, so the qualifier is only prepended when it belongs to a joined table.

src/main/java/com/mongodb/hibernate/jdbc/MongoResultSet.java
```diff
@@ -245,6 +245,22 @@ private <T> T getValue(int columnIndex, SqlFunction<BsonValue, T> toJavaConverte
         }
     }
 
+    private static @Nullable BsonValue getNestedBsonValue(BsonDocument document, String key) {
```
jyemin: Column keys for joined tables use dotted paths (e.g. `o1_0.total`) to navigate into the sub-documents produced by `$lookup`/`$unwind`. Bare keys still work because the loop finds no dot and falls through to a direct `get`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1144,13 +1174,108 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
             throw new FeatureNotSupportedException("Only single root from clause is supported");
         }
         var root = fromClause.getRoots().get(0);
-        if (root.hasRealJoins()) {
-            throw new FeatureNotSupportedException("TODO-HIBERNATE-65 https://jira.mongodb.org/browse/HIBERNATE-65");
+        if (root instanceof PluralTableGroup pluralRoot) {
```
jyemin, line 1222: `PluralTableGroup` appears as a FROM-clause root when Hibernate generates a collection-loader query — for example, for `@OneToMany` or `@ElementCollection` mappings — during `SessionFactory` initialization. It is never the result of a user-written HQL SELECT.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1144,13 +1174,108 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+            var elementDescriptor = pluralRoot.getModelPart().getElementDescriptor();
+            if (elementDescriptor instanceof EmbeddableValuedModelPart embeddablePart
```
jyemin, line 1224: `@Struct` embeddable element collections (`aggregateMapping != null`) store as nested BSON documents, which we already support. Non-`@Struct` embeddables require a different physical representation that is not supported yet (HIBERNATE-169).

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstUnwindStage.java
```diff
@@ -0,0 +1,41 @@
...
+ * See <a href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/unwind/">{@code $unwind}</a>.
+ *
+ * @hidden
+ */
+public record AstUnwindStage(String path, boolean preserveNullAndEmptyArrays) implements AstStage {
+    @Override
+    public void render(BsonWriter writer) {
+        writer.writeStartDocument();
+        writer.writeName("$unwind");
+        if (preserveNullAndEmptyArrays) {
```
jyemin, line 31: `$lookup` produces an empty array for documents with no matching foreign rows. Without `preserveNullAndEmptyArrays`, `$unwind` would silently drop those documents, losing LEFT JOIN (outer) semantics. INNER JOIN can omit the flag because dropping non-matching documents is the correct behaviour.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1144,13 +1174,108 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+                                    "TODO-HIBERNATE-161 https://jira.mongodb.org/browse/HIBERNATE-161");
+                        case FULL ->
+                            throw new FeatureNotSupportedException(
+                                    "TODO-HIBERNATE-162 https://jira.mongodb.org/browse/HIBERNATE-162");
+                        case CROSS ->
+                            throw new FeatureNotSupportedException(
+                                    "TODO-HIBERNATE-163 https://jira.mongodb.org/browse/HIBERNATE-163");
+                    };
+
+            if (!joinedGroup.getNestedTableGroupJoins().isEmpty()) {
```
jyemin, line 1282: Non-empty nested table group joins arise from entity-syntax joins (`JOIN Order o ON ...`) and from association-navigation joins that produce a sub-group structure, neither of which can be flattened to a simple `$lookup` (HIBERNATE-168).

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1144,13 +1174,108 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+            if (elementDescriptor instanceof EmbeddableValuedModelPart embeddablePart
+                    && embeddablePart.getEmbeddableTypeDescriptor().getAggregateMapping() == null) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-169 https://jira.mongodb.org/browse/HIBERNATE-169");
+            }
+            if (!(root.getPrimaryTableReference() instanceof NamedTableReference)) {
+                throw new FeatureNotSupportedException("Only named table references are supported");
+            }
+        } else {
+            if (!(root.getModelPart() instanceof EntityPersister entityPersister)
```
jyemin: `getQuerySpaces()` returns one entry per physical table. Multiple entries indicate joined-table or table-per-class inheritance, where the entity spans several tables. Translating that would require cross-table joins we cannot yet express in MQL.

docs/superpowers/specs/2026-05-21-mql-join-support-design.md
jyemin, line 1: This file is background info and will not be committed.

docs/superpowers/specs/2026-05-21-mql-join-support-design.md
```diff
@@ -0,0 +1,350 @@
...
+]}
+```
+
+### LEFT OUTER JOIN
+
+HQL: `SELECT c.id, o.total FROM Customer c LEFT JOIN Order o ON c.id = o.customerId`
+
+```json
+{ "aggregate": "customers", "pipeline": [
+  { "$lookup": { "from": "orders", "localField": "_id", "foreignField": "customerId", "as": "o1_0" } },
```
vbabanin: Should we consider using `#` instead of `_` in the generated intermediate names as well? It would reduce the chance of colliding with user-provided aliases like `o1_0`. Probably low risk either way, so we can also leave it as-is and revisit later if needed.

UPD: discussed, decided to use `#` in front the alias.

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,629 @@
...
+import org.junit.jupiter.api.Test;
+
+@DomainModel(
+        annotatedClasses = {
+            JoinSelectQueryIntegrationTests.Customer.class,
+            JoinSelectQueryIntegrationTests.Order.class,
+            JoinSelectQueryIntegrationTests.LineItem.class,
+            JoinSelectQueryIntegrationTests.OrderWithArray.class
+        })
+class JoinSelectQueryIntegrationTests extends AbstractQueryIntegrationTests {
```
vbabanin, line 58: Should we add a few targeted test cases here to document what we support and increase use-case coverage?

For example, `@ManyToMany`, `@OneToOne` and `@ElementColllection` associations also work via the translator today. We could cover each association type in isolation using a `@Nested` class with its own minimal `@DomainModel`, so the mapping doesn’t bleed into other tests/entities.

Example for `@ManyToMany`:

```java
@Nested
@DomainModel(annotatedClasses = { M2mStudent.class, M2mCourse.class })
class ManyToManyJoin extends AbstractQueryIntegrationTests {


   // setUp() ...

    @Test
    void testManyToManyJoinWorks() {
       assertSelectionQuery("FROM M2mStudent s JOIN s.courses c"
       ....
       );
      
    }
}

@Entity(name = "M2mStudent") @Table(name = "m2m_student")
static class M2mStudent {
    @Id int id;

    @ManyToMany
    @JoinTable(
        name = "m2m_enrollment",
        joinColumns = @JoinColumn(name = "student_id"),
        inverseJoinColumns = @JoinColumn(name = "course_id")
    )
    List<M2mCourse> courses = new ArrayList<>();
}

@Entity(name = "M2mCourse") @Table(name = "m2m_course")
static class M2mCourse {
    @Id int id;
    String title;
}
```

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,629 @@
...
+    }
+
+    @Test
+    void testJoinFetchManyToOne() {
+        var orders = new ArrayList<Order>();
+        getSessionFactoryScope().inTransaction(session -> {
+            orders.addAll(session.createSelectionQuery("FROM Order o JOIN FETCH o.customer ORDER BY o.id", Order.class)
+                    .getResultList());
+            assertActualCommandsInOrder(
+                    BsonDocument.parse(
```
vbabanin: We have `assertSelectionQuery` helpers, but they don’t currently return a `List<...>` for JOIN FETCH-style assertions. We could add an overload that returns the hydrated results and reuse it across both tests.

Concrete example of what I mean: https://github.com/vbabanin/mongo-hibernate/commit/7c657e1a9f5a4dc92a490f1e812a67ae99b42261

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,629 @@
...
+        }
+
+        @Test
+        void testOnConditionWithAssociationNavigationThrows() {
+            assertSelectQueryFailure(
+                    "SELECT c.id FROM Customer c JOIN c.orders o ON o.customer.name = 'Alice'",
+                    Object[].class,
+                    FeatureNotSupportedException.class,
+                    "TODO-HIBERNATE-168 https://jira.mongodb.org/browse/HIBERNATE-168");
+        }
```
vbabanin, line 836:
```suggestion
        }

        @Test
        void testOnConditionWithBothSidesFromOuterTableThrows() {
            assertSelectQueryFailure(
                    "SELECT c.id FROM Customer c JOIN Order o ON c.id = c.id",
                    Object[].class,
                    FeatureNotSupportedException.class,
                    "TODO-HIBERNATE-170 https://jira.mongodb.org/browse/HIBERNATE-170");
        }

        @Test
        void testOnConditionWithBothSidesFromJoinedTableThrows() {
            assertSelectQueryFailure(
                    "SELECT c.id FROM Customer c JOIN Order o ON o.id = o.total",
                    Object[].class,
                    FeatureNotSupportedException.class,
                    "TODO-HIBERNATE-170 https://jira.mongodb.org/browse/HIBERNATE-170");
        }
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1172,13 +1206,114 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+            if (primaryRef instanceof QueryPartTableReference) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-167 https://jira.mongodb.org/browse/HIBERNATE-167");
+            }
+            if (!(primaryRef instanceof NamedTableReference joinedNtr)) {
+                throw new FeatureNotSupportedException("Unsupported table reference type: "
+                        + primaryRef.getClass().getSimpleName());
+            }
+
+            // TODO-HIBERNATE-171: when JOINED inheritance and @SecondaryTable are supported, reject (or handle)
```
vbabanin: The TODO reads a bit self-contradictory: `@SecondaryTable` implies multiple query spaces, so if we “support @SecondaryTable” we shouldn’t also “reject entities spanning multiple query spaces.”

Also, the HIBERNATE-171 reference seems unrelated (it’s about auto-inferring dialect/connection provider from the MongoDB JDBC URL).

Should we create a ticket to track support for `@Inheritance(JOINED)` / `@SecondaryTable`? It looks like both go through the same underlying mechanism (`getTableReferenceJoins`).

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -0,0 +1,629 @@
...
+            // HQL parser rejects subquery join syntax before translation with SemanticException
+            assertSelectQueryFailure(
+                    "SELECT c.id FROM Customer c JOIN (SELECT o.id FROM Order o) ord ON c.id = ord.id",
+                    Object[].class,
+                    org.hibernate.query.SemanticException.class,
+                    "Select item at position 1 in select list has no alias");
+        }
+
+        @Test
+        void testEntitySyntaxJoinThrows() {
```
vbabanin: [nit] `testEntitySyntaxJoinThrows` sounds like it’s testing any join that mentions an entity name (which is effectively every join). Maybe we could rename to reflect the specific unsupported form being exercised (e.g., `testRootEntityJoinSyntaxThrows`).

Ref: https://docs.hibernate.org/orm/6.6/userguide/html_single/#hql-root-join

```suggestion
        void testRootEntityJoinSyntaxThrows() {
```

docs/superpowers/specs/2026-05-21-mql-join-support-design.md
```diff
@@ -0,0 +1,368 @@
...
+}
+```
+
+`getAggregateMapping()` returns non-null only for `@Struct`-annotated embeddables, which are stored as an aggregate BSON document in the parent collection. Non-`@Struct` embeddable element collections map to a secondary table and are rejected.
+
+## Supported and Unsupported Shapes
+
+All unsupported shapes throw `FeatureNotSupportedException`. Shapes that are reasonably implementable in MongoDB MQL reference a Jira ticket; shapes that are outside the MQL join surface get a descriptive message.
+
+| Shape | Disposition | Ticket |
```
vbabanin: There are also JOIN-related features we could consider tracking via tickets:
- `@FilterJoinTable` - triggers `visitFilterPredicate` - @vbabanin todo
- `@JoinFormula` - @vbabanin todo
- `@Inheritance(JOINED)` - discussed there is a ticket HIBERNATE-69 for that work.
- `@SecondaryTable` - located in `getTableReferenceJoins()` - @vbabanin todo

We could also add them to the `unsupported` tests suite. Right now these fail with `FeatureUnsupportedException` with a `null` message.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1172,13 +1206,114 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+                        "TODO-HIBERNATE-111 https://jira.mongodb.org/browse/HIBERNATE-111");
+            }
+            if (primaryRef instanceof QueryPartTableReference) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-167 https://jira.mongodb.org/browse/HIBERNATE-167");
+            }
+            if (!(primaryRef instanceof NamedTableReference joinedNtr)) {
+                throw new FeatureNotSupportedException("Unsupported table reference type: "
+                        + primaryRef.getClass().getSimpleName());
+            }
```
vbabanin, line 1303: Nice catch on guarding `FunctionTableReference` and `QueryPartTableReference`. Two more to watch out for: `UnionTableReference`, and also `ColumnReference` when it represents a `@JoinFormula`.

Both can leak into MQL translation. For example, with `@JoinFormula` the generated `$lookup.localField` becomes a SQL expression:

<details> 
<summary> MQL </summary>

```json
{
  "aggregate": "ItemWithJoinFormula",
  "pipeline": [
    {
      "$lookup": {
        "from": "ItemWithJoinFormulaTarget",
        "localField": "COALESCE(iwjf1_0.targetRef, 0)",
        "foreignField": "_id",
        "as": "#t1_0"
      }
    },
    { "$unwind": "$#t1_0" },
    { "$project": { "_id": true, "t1_0#_id": "$#t1_0._id", "t1_0#string": "$#t1_0.string", "targetRef": true } }
  ]
}
```

</details>

**UPD:** Added a check.

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
vbabanin: Added suggested commit to this PR: https://github.com/mongodb/mongo-hibernate/pull/160/changes/be8d0f78c4ce617adc29f1b0ffafdcf4edeb8a9e

docs/superpowers/specs/2026-05-21-mql-join-support-design.md
vbabanin: Created tickets and  added `EntityPersister` checks to throw `FeatureUnsupportedException` per each:
 https://jira.mongodb.org/browse/HIBERNATE-182
 https://jira.mongodb.org/browse/HIBERNATE-181

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Changed reference to HIBERNATE-69 and HIBERNATE-181 (new ticket).

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
vbabanin: Added suggested commit to this PR: https://github.com/mongodb/mongo-hibernate/pull/160/commits/4a8a69aa33a0b9696c9c741cfa41bdc122398b4a.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1172,13 +1219,133 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+            if (!(root.getPrimaryTableReference() instanceof NamedTableReference)) {
+                throw new FeatureNotSupportedException("Only named table references are supported");
+            }
+        } else {
+            if (!(root.getModelPart() instanceof EntityPersister entityPersister)) {
+                throw new FeatureNotSupportedException("Only single table from clause is supported");
+            }
+            if (entityPersister.getQuerySpaces().length != 1) {
+                if (entityPersister instanceof JoinedSubclassEntityPersister) {
+                    throw new FeatureNotSupportedException(
```
jyemin, line 1238: This path is untested.  Let me know if that's intentional.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1172,13 +1219,133 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+                    && embeddablePart.getEmbeddableTypeDescriptor().getAggregateMapping() == null) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-169 https://jira.mongodb.org/browse/HIBERNATE-169");
+            }
+            if (!(root.getPrimaryTableReference() instanceof NamedTableReference)) {
+                throw new FeatureNotSupportedException("Only named table references are supported");
+            }
+        } else {
+            if (!(root.getModelPart() instanceof EntityPersister entityPersister)) {
+                throw new FeatureNotSupportedException("Only single table from clause is supported");
```
jyemin, line 1234: This path is untested.  Let me know if that's intentional.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1172,13 +1219,133 @@ private static void checkFromClauseSupportability(FromClause fromClause) {
...
+            if (primaryRef instanceof FunctionTableReference) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-111 https://jira.mongodb.org/browse/HIBERNATE-111");
+            }
+            if (primaryRef instanceof QueryPartTableReference) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-167 https://jira.mongodb.org/browse/HIBERNATE-167");
+            }
+            if (primaryRef instanceof UnionTableReference) {
+                throw new FeatureNotSupportedException("TABLE_PER_CLASS inheritance joins are not supported");
```
jyemin, line 1298: This path is untested.  Let me know if that's intentional.

jyemin (commented):
Please confirm that the three untested code paths for unsupported functionality are intentional. 

Other than that, LGTM.  

FYI: I  pushed three commits

1. spotlessApply
2. delete design doc
3. merge commit with main

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Yes, this is intentional. We already fail fast earlier:

```java
if (entityPersister instanceof UnionSubclassEntityPersister) {
    throw new FeatureNotSupportedException("TABLE_PER_CLASS inheritance is not supported");
}
```

This additional check is for defensiveness and to stay consistent with the other guards in this code path.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: I initially left this untested intentionally because Hibernate was hitting `writeColumnFragment` and failing first. 
However, after removing the `@Discriminator`, it started reaching this path, so I updated the test in this commit and it’s now covered. Thanks! 

Commit: https://github.com/mongodb/mongo-hibernate/pull/160/commits/ff48f63fbef99d8ad9abc7e085e9976325b9136a

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Yes, intentional. Now that each specific case has dedicated handling (with a corresponding ticket in the exception message), I wasn’t able to find an additional scenario that hits this generic fallback path. It serves as a defensive guard.

vbabanin (dismissed):
LGTM!

vbabanin (approved):

strogiyotec (approved):
Approved without review !

PR #161, opened 2026-05-29
HIBERNATE-67: translate EXISTS-over-UNNEST subquery to $elemMatch

jyemin:
Translates HQL `WHERE EXISTS (FROM c.lineItems li WHERE li.sku = :v)` to a MongoDB `$elemMatch` filter within the existing `$match` stage, preserving parent cardinality.

- Add MongoUnnestFunction and register it in MongoDialect so Hibernate's HQL parser resolves `FROM c.collection alias` inside EXISTS subqueries
- Add AstElemMatchFilterOperation AST record that renders `$elemMatch`
- Implement visitExistsPredicate: recognizes the EXISTS-over-UNNEST shape, sets elemMatchInnerAlias context flag around body translation, builds the $elemMatch filter
- Augment visitColumnReference to throw HIBERNATE-177 when an outer entity field reference is encountered inside an $elemMatch body, covering all predicate types automatically

Unsupported shapes throw FeatureNotSupportedException:
- Correlated outer references in the body (TODO-HIBERNATE-177)
- All other EXISTS forms — over entities, with GROUP BY, UNION, etc. (TODO-HIBERNATE-178)


src/main/java/com/mongodb/hibernate/internal/dialect/function/array/MongoUnnestFunction.java
```diff
@@ -0,0 +1,57 @@
...
+ * alias} inside EXISTS subqueries. Rendering is never reached — the MQL translator intercepts the
+ * {@code FunctionTableReference} directly in {@code recognizeExistsOverUnnest}.
+ *
+ * @hidden
+ * @mongoCme Must be thread-safe.
+ */
+public final class MongoUnnestFunction extends AbstractSqmSelfRenderingSetReturningFunctionDescriptor {
+
+    public MongoUnnestFunction() {
+        super(
```
jyemin, line 43: This constructor is copied from Hibernate's own `UnnestFunction` class.  Rather than extend that class, we just extend the same class that it does, since we don't want any of its implementation.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -214,6 +215,8 @@ public abstract class AbstractMqlTranslator<T extends JdbcOperation> implements
 
     private final AstVisitorValueHolder astVisitorValueHolder = new AstVisitorValueHolder();
 
+    private @Nullable String elemMatchInnerAlias;
```
jyemin, line 233: lmk if you think this deserves a comment, since its global state.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -997,9 +1007,72 @@ public void visitInArrayPredicate(InArrayPredicate inArrayPredicate) {
...
     public void visitExistsPredicate(ExistsPredicate existsPredicate) {
-        throw new FeatureNotSupportedException();
+        var shape = recognizeExistsOverUnnest(existsPredicate);
+        if (shape == null) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-178 https://jira.mongodb.org/browse/HIBERNATE-178");
+        }
+        AstFilter bodyFilter;
+        if (shape.body() != null) {
+            var previousInnerAlias = elemMatchInnerAlias;
+            elemMatchInnerAlias = shape.innerAlias();
```
jyemin, line 1065: I don't see how you'd ever get recursive exists predicate over unnest, so this may be too defensive, but it isn't wrong.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -997,9 +1007,72 @@ public void visitInArrayPredicate(InArrayPredicate inArrayPredicate) {
...
+        var arg = args.get(0);
+        if (!(arg instanceof BasicValuedPathInterpretation<?> bvpi)) {
+            return null;
+        }
+        var columnReference = bvpi.getColumnReference();
+        if (columnReference == null) {
+            return null;
+        }
+        var arrayFieldName = columnReference.getColumnExpression();
+        return new ExistsOverUnnestShape(
```
jyemin: The most dangerous part, since any improper fall through to here will cause incorrect translation rather an exception.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -997,9 +1007,72 @@ public void visitInArrayPredicate(InArrayPredicate inArrayPredicate) {
...
+                || qs.hasOffsetOrFetchClause()) {
+            return null;
+        }
+        var root = qs.getFromClause().getRoots().get(0);
+        if (!root.getTableGroupJoins().isEmpty()
+                || !root.getNestedTableGroupJoins().isEmpty()) {
+            return null;
+        }
+        if (!(root.getPrimaryTableReference() instanceof FunctionTableReference ftr)
+                || !"unnest".equals(ftr.getFunctionExpression().getFunctionName())) {
```
vbabanin: Let’s move this constant to `MongoUnnestFunction`?

That way, readers can click through from MongoUnnestFunction.FUNCTION_NAME to the function implementation or any related usages in the future.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -997,9 +1007,72 @@ public void visitInArrayPredicate(InArrayPredicate inArrayPredicate) {
...
+        if (existsPredicate.isNegated()) {
+            filter = new AstLogicalFilter(NOR, List.of(filter));
+        }
+        astVisitorValueHolder.yield(FILTER, filter);
     }
 
+    private static @Nullable ExistsOverUnnestShape recognizeExistsOverUnnest(ExistsPredicate existsPredicate) {
+        var select = existsPredicate.getExpression();
+        if (!(select.getQueryPart() instanceof QuerySpec qs)) {
+            return null;
```
vbabanin: [nit] The pattern we’ve been using in the translator is to return `Optional` instead of null for absent optional stages, mainly for readability.

See: 
- [creteaMatchStage](https://github.com/mongodb/mongo-hibernate/blob/ba8816e1f4086ef1ed4cf5f3227de41538fda962/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java#L437)
- [createSortStage](https://github.com/mongodb/mongo-hibernate/blob/ba8816e1f4086ef1ed4cf5f3227de41538fda962/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java#L450)

Let’s use the same pattern here as well?

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -997,9 +1007,72 @@ public void visitInArrayPredicate(InArrayPredicate inArrayPredicate) {
 
     @Override
     public void visitExistsPredicate(ExistsPredicate existsPredicate) {
```
vbabanin, line 1054: [nit] What do you think about moving this logic into a separate method so `visitExistsPredicate` reads as:

```java
@Override
public void visitExistsPredicate(final ExistsPredicate existsPredicate) {
    astVisitorValueHolder.yield(FILTER, translateExistsOverUnnest(existsPredicate));
}
```

I think naming the helper makes the intent easier to understand at a glance, given that that this path only supports `EXISTS` over `unnest`.

vbabanin (commented):
Looks good! Just a couple of nits.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -600,6 +603,13 @@ public void visitColumnReference(ColumnReference columnReference) {
         if (columnReference.isColumnExpressionFormula()) {
             throw new FeatureNotSupportedException("Formula is not supported");
         }
+        if (elemMatchInnerAlias != null) {
+            var qualifier = columnReference.getQualifier();
+            if (qualifier != null && !qualifier.equals(elemMatchInnerAlias)) {
```
vbabanin: Currently, as far as I can tell, all select queries should produce a qualifier, even though `getQualifier`() is marked `@Nullable` - so we should never observe `null`.

However, should we use `assertNotNull` here so we don’t silently produce incorrect MQL if a future Hibernate version returns `null` in some edge-cases? This is mostly just defensive.

```suggestion
            var qualifier = assertNotNull(columnReference.getQualifier());
            if (!qualifier.equals(elemMatchInnerAlias)) {
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Agreed.  Done.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: I like it. Done.

jyemin:
Comments addressed, design doc deleted.

jyemin:
There were merge conflicts with the join PR, and new failing tests.  I've pushed the merge commit and the test fixes.

vbabanin (approved):
LGTM!

PR #162, opened 2026-06-09
Auto-infer dialect from url and connection provider from dialect

jyemin:
Simplifies MongoDB extension configuration in two steps:

**HIBERNATE-124** registers `"MongoDB"` as the dialect short name and auto-configures `MongoConnectionProvider` when the MongoDB dialect is in use, so users no longer need to set `hibernate.connection.provider_class`. Moves `com.mongodb.hibernate.dialect` and `com.mongodb.hibernate.jdbc` to `internal`, removing them from the public API.

**HIBERNATE-171** auto-infers the dialect from a `mongodb://` or `mongodb+srv://` JDBC URL, so users no longer need to set `hibernate.dialect` either.

The net result: a MongoDB URL alone is sufficient to configure the extension:

```properties
jakarta.persistence.jdbc.url=mongodb://localhost/mydb
```

**Key functional changes:**

 - `MongoNamedStrategyContributor` — new: registers "MongoDB" as the short name for the MongoDB dialect
 - `StandardServiceRegistryScopedState.contribute():` auto-infers dialect when a `mongodb://` or `mongodb+srv://` URL is present (HIBERNATE-171); auto-configures `MongoConnectionProvider` when MongoDB
  dialect is configured (HIBERNATE-124)
 - `StandardServiceRegistryScopedState`: removes `checkMongoConnectionProviderIsPluggedIn()` — the connection provider is now auto-configured instead of required to be set explicitly
 - `META-INF/services/...NamedStrategyContributor` — new SPI registration

**Key test changes:**

`StandardServiceRegistryScopedStateTests` — new tests:
  - `testDialectInferredFromMongoUrl` / `testDialectInferredFromMongoSrvUrl` — both dialect and connection provider are auto-set from a `mongodb://` / `mongodb+srv://` URL
  - `testExplicitDialectNotOverriddenByMongoUrl` — explicit dialect is preserved when URL is also present
  - `testNonMongoDialectWithMongoUrlIsSilent` — non-Mongo dialect alongside a MongoDB URL is silently ignored
  - `testIncompatibleConnectionProvider` — throws when a non-Mongo connection provider is explicitly set
  - `testMongoDialectNotPluggedIn` — updated error message to [hibernate.dialect] must be set to [MongoDB]
  - `testMongoConnectionProviderNotPluggedIn` → replaced by `testIncompatibleConnectionProvider` (connection provider is now auto-configured; error only fires on explicit incompatible value)

  Deleted:
  - `NativeBootstrappingTests.testMongoDialectNotPluggedIn` / `testMongoConnectionProviderNotPluggedIn` — redundant with the unit tests above; removed
  - `JakartaPersistenceBootstrappingIntegrationTests` — was testing explicit CONNECTION_PROVIDER config; no longer needed since it's auto-configured


docs/superpowers/specs/2026-05-30-hibernate-124-design.md
strogiyotec, line 1: I am not sure how I feel about these spec files, I didn't write any of them for my previous PRs
It's good to know what LLM plan was but do we really have to commit these files? 
If we do I think we should enforce it under every PR
I personally don't think there is any value in maintaining these spec files as long as DOCSP team will document every change

docs/superpowers/specs/2026-05-30-hibernate-124-design.md
```diff
@@ -0,0 +1,97 @@
...
+## Dialect Short Name
+
+Register `"MongoDB"` as a short name for `MongoDialect` via a new
+`MongoNamedStrategyContributor` implementing Hibernate's `NamedStrategyContributor`
+SPI. This follows Hibernate ORM naming conventions (PascalCase for dialects,
+e.g. `PostgreSQL`, `MySQL`).
+
+## Connection Provider Auto-Configuration
+
+`MongoConnectionProvider` cannot be registered as a named strategy because
```
strogiyotec: Note for myself: Postgres also doesn't register it's own provider 
so if client has two providers in the classpath Postgres and Mongo none of them will conflict with each other

src/integrationTest/java/com/mongodb/hibernate/embeddable/StructAggregateEmbeddableIntegrationTests.java
```diff
@@ -64,7 +65,8 @@
             StructAggregateEmbeddableIntegrationTests.Unsupported.ItemWithNestedValueHavingEmbeddable.class
         })
 @ExtendWith(MongoExtension.class)
-public class StructAggregateEmbeddableIntegrationTests implements SessionFactoryScopeAware {
+public class StructAggregateEmbeddableIntegrationTests
```
strogiyotec, line 68: what do you think about adding this line to AGENTS.md 
Code style section ? 
> - **Integration test registry setup:** every integration test that builds a Hibernate `SessionFactory`/`ServiceRegistry` through the testing framework must implement `MongoServiceRegistryProducer` (in `com.mongodb.hibernate.junit`). It strips the `SharedDriverManagerConnectionProvider` that Hibernate's test framework injects, which would otherwise trip the auto-configured connection provider

strogiyotec (changes requested):
Left a few comments
Also raised a PR against your fork to fix a build 
https://github.com/jyemin/mongo-hibernate/pull/3

docs/superpowers/specs/2026-05-30-hibernate-124-design.md
jyemin: @vbabanin said he found them helpful during review, but the plan is to strip then from the PR before merging.

strogiyotec (dismissed):

strogiyotec (approved):

PR #164, opened 2026-06-11
HIBERNATE-176: upgrade hibernate-orm from 7.3 to 7.4

jyemin:


src/test/java/com/mongodb/hibernate/boot/NativeBootstrappingTests.java
```diff
@@ -79,28 +79,20 @@ void testMongoConnectionProviderNotPluggedIn() {
      */
     @Test
     void testMongoAdditionalMappingContributorIsSkipped() {
-        var standardServiceRegistryBuilder = new StandardServiceRegistryBuilder();
```
jyemin, line 82: The upgrade caused this test to fail, so it had to be changed. In 7.3 it used a mock `ConnectionProvider` + PostgreSQL dialect and asserted `InstantiationException` when persisting an entity with a null composite ID. In 7.4 a different exception is thrown for the same operation, so rather than track down and pin the new type, the test was rewritten around what it actually needs to prove: `MongoAdditionalMappingContributor` throws during `buildMetadata()`, not at query time, so we now simply assert that that `buildSessionFactory()` succeeds using the H2 in-memory database.

src/test/java/com/mongodb/hibernate/boot/NativeBootstrappingTests.java
```diff
@@ -79,28 +79,20 @@ void testMongoConnectionProviderNotPluggedIn() {
...
-            assertThatThrownBy(
-                            () -> sessionFactory.inSession(session -> session.persist(new ItemWithUnsupportedId(null))))
-                    .hasMessageNotContaining("does not support primary key spanning multiple columns")
-                    .isInstanceOf(InstantiationException.class)
-                    .hasMessageMatching("Could not instantiate entity .* due to: null");
-        }
+        // clearSettings() prevents hibernate.properties from injecting MongoConnectionProvider
+        var builder = new StandardServiceRegistryBuilder();
+        builder.clearSettings();
+        assertThatNoException().isThrownBy(() -> {
```
jyemin, line 85: I tested that a similar test with  the mongodb dialect, throws from `MongoAdditionalMappingContributor#forbidStructIdentifier` when using the MongoDB dialect, so this is proving that we're _not_ using the MongoDB dialect (which means that  the early return in `MongoAdditionalMappingContributor#contribute` is happening) when the dialect is not MongoDB.

gradle/libs.versions.toml
```diff
@@ -26,12 +26,13 @@ assertj = "3.27.6"
...
 nullaway = "0.12.14"
 jspecify = "1.0.0" # Remember to update javadoc links
-hibernate-orm = "7.3.6.Final" # Remember to update javadoc links
+hibernate-orm = "7.4.1.Final" # Remember to update javadoc links
 mongo-java-driver-sync = "5.6.2" # Remember to update javadoc links
 findbugs-jsr = "3.0.2"
 slf4j-api = "2.0.17"
 logback-classic = "1.5.21"
 mockito = "5.21.0"
+h2 = "2.3.232"
```
strogiyotec, line 35: let's upgrade it to `2.4.240` I tested it out locally and it works 
2.3 was released in 2024
2.4 was released in 2025
https://github.com/h2database/h2database/releases/tag/version-2.4.240

gradle/libs.versions.toml
```diff
@@ -26,12 +26,13 @@ assertj = "3.27.6"
 google-errorprone-core = "2.42.0"
 nullaway = "0.12.14"
 jspecify = "1.0.0" # Remember to update javadoc links
-hibernate-orm = "7.3.6.Final" # Remember to update javadoc links
+hibernate-orm = "7.4.1.Final" # Remember to update javadoc links
```
strogiyotec, line 29: as this [test ](https://github.com/mongodb/mongo-hibernate/blob/9b0e49584461aa88896c740ffff8d5677ea285f1/src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java#L907)still passes should we update the comment to "Hibernate 7.4 fails" ? 
Feel free to skip this comment

strogiyotec (changes requested):

gradle/libs.versions.toml
jyemin: Not worth it because unnest support is in review, and the whole test will be removed.

gradle/libs.versions.toml
jyemin: Done

strogiyotec (approved):

PR #166, opened 2026-06-16
HIBERNATE-93: support path expressions for @Struct aggregate embeddables

jyemin:
Translate HQL path expressions into @Struct aggregate embeddable fields (nested.a, outer.inner.a) to MongoDB dot notation in $match, $sort, $set, and to the #-aliased form ({"nested#a": "$nested.a"}) in $project.

- MongoAggregateSupport: switch to the Hibernate 7 int-typed signatures and return the dot-notation path; remove the obsolete checkSupported guard, UNSUPPORTED_MESSAGE_PREFIX, and the four call sites in MongoStatement / MongoPreparedStatement
- AbstractMqlTranslator: add a $project branch for non-join dotted struct paths, preserving the existing join-projection branch
- integration tests: convert the former Unsupported cases to positive and add coverage for select, filter, sort, update, delete, multi-field, and two-level struct paths


vbabanin (dismissed):
LGTM!

vbabanin (approved):
LGTM!

PR #168, opened 2026-06-18
HIBERNATE-172: Spring Boot auto-configuration and starter for MongoDB-backed JPA

jyemin:
## Summary

Adds Spring Boot support for MongoDB-backed JPA via this extension: a starter, an
auto-configuration module, and a small core-module change letting Hibernate borrow
a Spring-managed `MongoClient` instead of creating its own.

In a Spring Boot app, setting `spring.jpa.database-platform=MongoDB` (plus
`spring-boot-mongodb` config) creates a JPA `EntityManagerFactory`,
`JpaTransactionManager`, and Spring Data JPA repository with no `DataSource`.

Design spec: `docs/superpowers/specs/2026-06-07-hibernate-172-design.md`

Spring app that uses this starter module: https://github.com/jyemin/spring-data-with-mongo-hibernate-test-app/tree/with-auto-configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,40 +71,40 @@ spring.mongodb.uri=mongodb://localhost/mydb
 When `spring.jpa.database-platform` is `MongoDB`, the starter auto-configures a JPA
 `EntityManagerFactory`, a `JpaTransactionManager`, and Spring Data JPA repositories backed by MongoDB.
 The connection is configured through Spring Boot's `spring-boot-mongodb` support (the `spring.mongodb.*`
-namespace), and the integration **borrows that `MongoClient`** rather than creating its own — a single
+namespace), and the integration **borrows that `MongoClient`** rather than creating its own: a single
 connection pool shared with health checks, metrics, and any other use of the Spring-managed client. This
 reuses `spring-boot-mongodb`'s client infrastructure and is **not** Spring Data MongoDB.
 
 The starter brings no SQL connection pool, so a MongoDB-only application needs no `spring.datasource.url`
 and Spring Boot's `DataSourceAutoConfiguration` stays inert. Without `spring.jpa.database-platform=MongoDB`
-the starter contributes nothing — it is safe to have on the classpath of a non-MongoDB application.
+the starter contributes nothing. It is safe to have on the classpath of a non-MongoDB application.
 
 #### Requirements
 
-The starter requires **Spring Boot 4.x** and **Hibernate ORM 7.3 or later** (the version the MongoDB
+The starter requires **Spring Boot 4.x** and **Hibernate ORM 7.4 or later** (the version the MongoDB
 extension is built against). Spring Boot manages the Hibernate version through its BOM, and some
-Spring Boot 4.x releases still ship an older Hibernate — for example, Spring Boot 4.0.6 manages
-Hibernate 7.2.12. When the managed version is below 7.3, the application fails to start with a
-`NoSuchMethodError` from the MongoDB dialect. Until your Spring Boot version's BOM ships Hibernate 7.3
+Spring Boot 4.x releases still ship an older Hibernate, for example, Spring Boot 4.0.6 manages
+Hibernate 7.2.12. When the managed version is below 7.4, the application fails to start with a
+`NoSuchMethodError` from the MongoDB dialect. Until your Spring Boot version's BOM ships Hibernate 7.4
 or later, override the managed version:
 
 Gradle (with the Spring Boot or `io.spring.dependency-management` plugin):
 
 ```groovy
-ext['hibernate.version'] = '7.3.6.Final'
+ext['hibernate.version'] = '7.4.1.Final'
 ```
 
 Maven:
 
 ```xml
 <properties>
-    <hibernate.version>7.3.6.Final</hibernate.version>
+    <hibernate.version>7.4.1.Final</hibernate.version>
 </properties>
 ```
 
 #### Configuration
 
-The standard Spring Boot JPA properties under `spring.jpa.*` are honored — for example
+The standard Spring Boot JPA properties under `spring.jpa.*` are honored. For example
 `spring.jpa.show-sql`, `spring.jpa.hibernate.ddl-auto`, and `spring.jpa.properties.*`, along with any
 `HibernatePropertiesCustomizer` / `EntityManagerFactoryBuilderCustomizer` beans and
 `spring.jpa.open-in-view` (Open Session in View, on by default in servlet web applications).
--- a/README.md
+++ b/README.md
@@ -53,6 +53,93 @@ The Java module with example applications is located in
 
 The examples may be run by running the smoke tests as specified in [Run Smoke Tests](#run-smoke-tests).
 
+### Spring Boot
+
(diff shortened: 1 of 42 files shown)


mongodb-hibernate-spring-boot-autoconfigure/src/integrationTest/java/com/mongodb/hibernate/springboot/MongoHibernateSpringBootIntegrationTests.java
```diff
@@ -0,0 +1,123 @@
...
+        txTemplate.executeWithoutResult(status -> bookRepository.save(book));
+
+        var found = bookRepository.findById(id);
+        assertThat(found).isPresent();
+        assertThat(found.get().title).isEqualTo("The Hobbit");
+    }
+
+    @Test
+    void jpaRepositoryBeanIsCreated() {
+        assertThat(bookRepository).isNotNull();
```
strogiyotec: this test doesn't check anything
Spring will fail on startup if there is no instance of `bookRepository` if we want it not to fail then 
```
Autowired(required=false)
BookRepository repository;
```
should be used

README.md
```diff
@@ -53,6 +53,93 @@ The Java module with example applications is located in
...
+
+```properties
+spring.jpa.database-platform=MongoDB
+spring.mongodb.uri=mongodb://localhost/mydb
+```
+
+When `spring.jpa.database-platform` is `MongoDB`, the starter auto-configures a JPA
+`EntityManagerFactory`, a `JpaTransactionManager`, and Spring Data JPA repositories backed by MongoDB.
+The connection is configured through Spring Boot's `spring-boot-mongodb` support (the `spring.mongodb.*`
+namespace), and the integration **borrows that `MongoClient`** rather than creating its own — a single
```
strogiyotec: this is an implementation detail do you think we should document it to the end users ?

README.md
```diff
@@ -53,6 +53,93 @@ The Java module with example applications is located in
 
 The examples may be run by running the smoke tests as specified in [Run Smoke Tests](#run-smoke-tests).
 
+### Spring Boot
```
strogiyotec: nit: maybe call this section `Spring Data` ?

README.md
```diff
@@ -53,6 +53,93 @@ The Java module with example applications is located in
...
+
+#### Limitations
+
+- **Testing:** the `@DataJpaTest` slice is not supported (it wires Spring Boot's SQL JPA
+  auto-configuration); use `@SpringBootTest` to test against MongoDB. Slice support is tracked by
+  [HIBERNATE-179](https://jira.mongodb.org/browse/HIBERNATE-179).
+- **Single persistence unit:** a single `@Primary` `EntityManagerFactory` is auto-configured.
+- **Native image:** GraalVM native-image execution is not yet validated.
+- A [replica set](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-replica-set) is
+  required, as transactions are (see [User Documentation](#user-documentation) above).
```
strogiyotec: I think you meant
`as transactions are not supported in standalone instance ` ?

mongodb-hibernate-spring-boot-autoconfigure/src/main/java/com/mongodb/hibernate/autoconfigure/MongoHibernateAutoConfiguration.java
```diff
@@ -0,0 +1,364 @@
...
+
+/**
+ * Spring Boot {@link AutoConfiguration auto-configuration} for MongoDB-backed JPA via the MongoDB Extension for
+ * Hibernate ORM.
+ *
+ * <p>Activated when {@link MongoConfigurationContributor} is on the classpath and {@code spring.jpa.database-platform}
+ * is {@code MongoDB}. It reuses Spring Boot's own JPA property machinery ({@link JpaProperties},
+ * {@link HibernateProperties}, {@link EntityManagerFactoryBuilder}) so the full {@code spring.jpa.*} surface is honored
+ * — {@code ddl-auto}, {@code show-sql}, naming strategies, {@link HibernatePropertiesCustomizer} and
+ * {@link EntityManagerFactoryBuilderCustomizer} beans — rather than only the raw {@code spring.jpa.properties.*}
```
strogiyotec: let's remove em dash as it makes the build fail 
```suggestion
 * {@link EntityManagerFactoryBuilderCustomizer} beans - rather than only the raw {@code spring.jpa.properties.*}
```
https://spruce.corp.mongodb.com/version/6a3338d945761000073dd91e

strogiyotec (changes requested):

README.md
jyemin: It's user-visible behavior, and it's also one of the main reasons for integrating in the way that we did.  I think it's worth keeping.

README.md
jyemin: It's about both, so I changed to reflect that.

mongodb-hibernate-spring-boot-autoconfigure/src/integrationTest/java/com/mongodb/hibernate/springboot/MongoHibernateSpringBootIntegrationTests.java
jyemin: I made all the beans `required=false`, and check for non-null of all three in a single renamed test.

README.md
jyemin: I removed the bullet since it's already mentioned in `User Documentation` anyway.

mongodb-hibernate-spring-boot-autoconfigure/src/main/java/com/mongodb/hibernate/autoconfigure/MongoHibernateAutoConfiguration.java
jyemin: Done, and also replaced em dashes with other alternatives elsewhere.

strogiyotec (approved):

src/main/java/com/mongodb/hibernate/cfg/MongoConfigurator.java
```diff
@@ -100,4 +114,15 @@ public sealed interface MongoConfigurator permits MongoConfigurationBuilder {
...
+    /**
+     * Supplies an externally managed {@link MongoClient} to use instead of creating one from the configured
+     * {@link MongoClientSettings}. When supplied, the client's lifecycle is owned by the caller: the extension uses it
+     * as-is and does not close it, and {@link #applyToMongoClientSettings(Consumer)} and the
+     * {@value AvailableSettings#JAKARTA_JDBC_URL} connection string are not used to build a connection.
+     *
+     * @param mongoClient the externally managed client to use.
+     * @return {@code this}.
+     */
+    MongoConfigurator mongoClient(MongoClient mongoClient);
```
mp911de, line 127: This is likely the most crucial aspect to provide a driver object as the JPA configuration approach either requires a datasource URL or a DataSource itself.

It is great to have `MongoConfigurationContributor` as config hook for applications to provide their own customizations, also considering ordering is great.

PR #169, opened 2026-06-22
HIBERNATE-61 native query with parameter marker 

strogiyotec:
[HIBERNATE-61](https://jira.mongodb.org/browse/HIBERNATE-61) allow question marks in native query 

~~The only thing that is not supported is `in :ids` operator~~
~~For example~~
```
{"aggregate": "items", "pipeline": [{"$match": {"_id": {"$in": [:ids] }}},...
```
~Will be converted to~ 
```
[{"aggregate": "items", "pipeline": [{"$match": {"_id": {"$in": [({"$undefined": true},{"$undefined": true})] }}},
```
~~the list of ids is wrapped with `()` which is invalid json~~
~~it happens [here ](https://github.com/hibernate/hibernate-orm/blob/c0af5d372fab35caedd6561fd4a7e71cd066d42e/hibernate-core/src/main/java/org/hibernate/query/sql/internal/NativeQueryImpl.java#L1006) within hibernate~~
# 2026-06-30 - new implementation 
The new implementation rewrites the whole query to replace `?` with undefined keyword 
## Advantages:
1. Now query logs are clear and do not carry `undefined` for example `2026-06-30T21:59:49,121 [main] DEBUG org.hibernate.SQL - {"aggregate": "items", "pipeline": [{"$match": {"_id": {"$in": [(?,?)] }}}` instead of list of `undefined`
2. We can now support `in` keyword that didn't work because hibernate wraps parameters into `()`

## Disadvantages
In `MongoConnection#translateParameterMarkers` we have to rebuild the query everytime which adds O(n) complexity 
I tried to run this code sample
```
 @Test
        void testMultiValuedParameter() {
            sessionFactoryScope.inSession(session -> {
                var mql =
                        """
                        {"aggregate": "%s", "pipeline": [{"$match": {"%s": {"$in": [:ids] }}}, %s]}"""
                                .formatted(
                                        COLLECTION_NAME,
                                        ID_FIELD_NAME,
                                        Item.projectAll().toBsonDocument().toJson(EXTENDED_JSON_WRITER_SETTINGS));
                assertEq(
                        item,
                        session.createNativeQuery(mql, Item.class)
                                .setParameter("ids", List.of(item.id, 999))
                                .getSingleResult()

               );
 session.createNativeQuery(mql, Item.class)
                                .setParameter("ids", List.of(item.id, 999))
                                .getSingleResult()
            });
        }
```
it basically runs the same native query twice and both times the query had to be rebuild from scratch , no caching

## Side note
I also found that queries with regex  that contain `?` without `"` do not work 
For example 
```
  sessionFactoryScope.inSession(session -> {
                var mql =
                        """
                        {"aggregate": "%s", "pipeline": [{"$match": {"%s": :id, "string": /st?r/ }}, %s]}"""
                                .formatted(
                                        COLLECTION_NAME,
                                        ID_FIELD_NAME,
                                        Item.projectAll().toBsonDocument().toJson(EXTENDED_JSON_WRITER_SETTINGS));
                assertThatThrownBy(() -> session.createNativeQuery(mql, Item.class))
                        .isInstanceOf(ParameterRecognitionException.class)
                        .hasMessageContaining("Cannot mix parameter styles");
            });
```
This doesn't work no matter the changes in this PR б ,Hibernate expects ? to be wrapped by `'` or `"` and fails if it is not


src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
```diff
@@ -106,6 +107,16 @@ public void contribute(StandardServiceRegistryBuilder serviceRegistryBuilder) {
                     serviceRegistryBuilder.applySetting(
                             AvailableSettings.CONNECTION_PROVIDER, MongoConnectionProvider.class.getName());
                 }
+                // Native queries must use the marker supplied by `MongoDialect.getNativeParameterMarkerStrategy`
```
strogiyotec: the parameter is enforced , can't be null or set to false

strogiyotec (commented):
a few comments

jyemin:
> The only thing that is not supported is in :ids operator

I didn't know about this feature, but glad you found it.  

The only way I can see to support it is by looking for the open/close paren pattern in the "JSON" before parsing it, and removing the offending characters.  But then once we do that we might as well start using `?` for the parameters themselves, and replacing with `{$undefined: true}` in post-processing, along with paren-removal.

If we do this in `MongoConnection#prepareStatement(java.lang.String)`, the nice side effect is that we will no longer see `{$undefined: true}` in hibernate query logs, when enabled.

What do you think?

CC @vbabanin.

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoConnection.java
```diff
@@ -131,6 +132,60 @@ public PreparedStatement prepareStatement(String mql) throws SQLException {
...
+     *
+     * <ul>
+     *   <li>a double-quoted string, e.g. {@code "a?b"}, delimited by {@code "} with {@code \} escapes;
+     *   <li>a shell-style regular-expression literal that MongoDB extended JSON accepts in a value position, e.g.
+     *       {@code /a(bc)?d/i}, delimited by {@code /} with {@code \} escapes (a regex commonly contains {@code ?}
+     *       quantifiers and {@code (} {@code )} groups). Outside a string, a {@code /} can only begin such a literal
+     *       because MQL has no division operator.
+     * </ul>
+     */
+    static String translateParameterMarkers(String mql) {
```
strogiyotec, line 158: this query has to be rebuilt every time to support regex and `?` along with `in` keyword

src/integrationTest/java/com/mongodb/hibernate/query/select/NativeQueryIntegrationTests.java
```diff
@@ -627,6 +628,140 @@ void testArrayAndCollectionValuesOfStructAggregateEmbeddablesHavingArraysAndColl
...
+    /**
+     * See <a href="https://docs.hibernate.org/orm/7.4/userguide/html_single/#sql-query-parameters">Using parameters in
+     * native queries</a>.
+     *
+     * <p>Hibernate hard-codes the JDBC {@code ?} marker for native queries, and expands a multi-valued parameter into a
+     * SQL-style parenthesized list. Neither is valid MQL, so the JDBC adapter rewrites the markers in
+     * {@link com.mongodb.hibernate.internal.jdbc.MongoConnection#prepareStatement(String)} before parsing.
+     */
+    @Nested
+    class WithParameters implements MongoServiceRegistryProducer {
```
jyemin, line 638: Let's also assert the actual commands sent to the server in these tests.  In order to do that it:

1. Make `NativeQueryIntegrationTests` extend `AbstractQueryIntegrationTests` and make any follow-up changes required from that change.  That gives access to the CommandListener needed for the actual assertions.
2. Add call to `AbstractQueryIntegrationTests#assertActualCommandsInOrder` to each test.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstParameterMarker.java
```diff
@@ -27,6 +27,9 @@ public final class AstParameterMarker implements AstValue {
 
     public static final AstParameterMarker INSTANCE = new AstParameterMarker();
 
+    /** The MQL ({@linkplain org.bson.json.JsonMode#EXTENDED extended JSON}) rendering of {@link #render}. */
```
jyemin: Now that we support `?` for parameters, let's use '? here too instead of `{"$undefined": true}`.  That way, all queries look the same -- native or HQL -- in Hibernate's SQL logs, and `{"$undefined": true}` becomes an implementation detail of the JDBC driver.

Fortunately, that's easy to do using `BsonWriter`, though it's a bit indirect:  just add a custom converter for `UNDEFINED` to the `JsonWriterSettings` in `MongoConstants`:

```
JsonWriterSettings.builder()
     ...
     .undefinedConverter((value, writer) -> writer.writeRaw("?"))
     ...
```

That will render UNDEFINED as `?` (`Raw` means the string is not wrapped in double-quotes, which is what we want).

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoConnection.java
```diff
@@ -131,6 +132,59 @@ public PreparedStatement prepareStatement(String mql) throws SQLException {
...
+     * are valid MQL, so before parsing we replace each {@code ?} with the BSON {@code undefined} marker (which
+     * {@link MongoPreparedStatement} binds) and drop the list-wrapping parentheses, turning {@code $in: [(?,?,?)]} into
+     * {@code $in: [{"$undefined": true},{"$undefined": true},{"$undefined": true}]}.
+     *
+     * <p>A {@code ?}, {@code (} or {@code )} is rewritten only when it is structural MQL syntax, i.e. outside any
+     * literal. Two literal forms can legitimately contain those characters and must be copied verbatim:
+     *
+     * <ul>
+     *   <li>a double-quoted string, e.g. {@code "a?b"}, delimited by {@code "} with {@code \} escapes;
+     *   <li>a shell-style regular-expression literal that MongoDB extended JSON accepts in a value position, e.g.
```
jyemin: Nice find!  I forgot JsonReader support shell syntax.

But this is actually a bigger issue than shell-style regex syntax because JsonReader also supports 

1. single-quoted strings, e.g. `{"foo": 'ba?r'}`.  Easily handled by this method by extending the quote-handling.
2. unquoted keys, e.g. `{foo: "bar"}`.  Maybe nothing to do here because I think you're only allowed to use the unquoted key syntax with alphanumeric characters.
3. constructor-syntax e.g. `{"foo": ISODate("2024-01-01T10:00:00Z")`.  This one is problematic, and I'm not sure we can work around it in this method without building a more fully-blown parser.  However, this syntax is not necessary, as you can always use the Extended JSON syntax (as you can with regex)

Perhaps we should just document that we only support the Extended JSON syntax, and not shell syntax (which most drivers don't support, this is just something Java driver did, following C#, because it seemed like a useful thing to provide to users).  Or maybe we support the first two scenarios but not the third.

WDYT?

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoConnection.java
strogiyotec: @jyemin  I think we are free to choose whatever is more convenient, as Hibernate won't interact with `MongoClient` directly we can add to the release notes that we do not support shell syntax 
My main concern is complexity, I don't like the idea that we are rewriting every query 
Meanwhile I can add support for a single quoted strings

src/integrationTest/java/com/mongodb/hibernate/query/select/NativeQueryIntegrationTests.java
strogiyotec: addressed it [here ](https://github.com/mongodb/mongo-hibernate/pull/169/changes/10bce670146c91a317ac37b0432f2e32f1fb832a#diff-8518c1d0d2076e27c84c352dc711b98365c424a54c767c4f05ac6e9e56d3ff0dR80)

the test now extends AbstractQueryIntegrationTests  and asserts that the actual command was ran

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstParameterMarker.java
strogiyotec: addressed it [here ](https://github.com/mongodb/mongo-hibernate/pull/169/changes/10bce670146c91a317ac37b0432f2e32f1fb832a#diff-6f6ebb38651a8babb88e2b60c228ef84c2bc300b72994078ed91be86f9446305R35)
also adjusted some unit tests that now assers `?` in the MQL instead of `undefined`

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoConnection.java
strogiyotec: I rewrote the `translateParameterMarkers` it now only supports extended json syntax
Added a comment in JIRA ticket to let docsp team know and asked Ajay to start a thread in dbx channel about the process to document our changes

src/integrationTest/java/com/mongodb/hibernate/query/select/NativeQueryIntegrationTests.java
```diff
@@ -333,69 +323,78 @@ void testEntity() {
...
-                            List.of(
-                                    match(eq(itemWithNestedValueHavingArraysAndCollections.id)),
-                                    project(include("nested"))));
-                    assertEq(
-                            itemWithNestedValueHavingArraysAndCollections.nested,
-                            session.createNativeQuery(
-                                            mql, StructAggregateEmbeddableIntegrationTests.ArraysAndCollections.class)
-                                    .getSingleResult());
-                }));
+        getSessionFactoryScope()
```
strogiyotec, line 326: had to adjust this test case because the session factory is now coming from the parent class

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstParameterMarker.java
```diff
@@ -27,7 +27,11 @@ public final class AstParameterMarker implements AstValue {
 
     public static final AstParameterMarker INSTANCE = new AstParameterMarker();
 
-    /** The MQL ({@linkplain org.bson.json.JsonMode#EXTENDED extended JSON}) rendering of {@link #render}. */
+    /**
+     * The BSON {@code undefined} value in ({@linkplain org.bson.json.JsonMode#EXTENDED extended}) JSON. This is the
+     * marker the JDBC adapter parses and binds; a query parameter renders as the JDBC standard {@code ?} in MQL (see
+     * {@code MongoConstants#EXTENDED_JSON_WRITER_SETTINGS}), which the adapter rewrites to this before parsing.
+     */
     public static final String MARKER = "{\"$undefined\": true}";
```
jyemin: Let's move this constant into MongoConnection, the only place it is still used.

jyemin (dismissed):
One small thing to change if you agree, but LGTM either way.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstParameterMarker.java
strogiyotec: Addressed it [here](https://github.com/mongodb/mongo-hibernate/pull/169/changes/f03daf53559ea51bfad10a8ecbe0d96bab3ef191)

jyemin (approved):
LGTM!

PR #170, opened 2026-06-23
HIBERNATE-51 Add optimistic locking support.

vbabanin:
## Scope

**In scope:**
 Both UPDATE and DELETE on single-table entities
- `@Version` numeric (`int`/`Integer`, `long`/`Long`) and `Instant`
- `@Version @CurrentTimestamp(source=VM)` - Hibernate generates value in JVM, passes as normal parameter
- `@OptimisticLocking(ALL)` with `@DynamicUpdate`
- `@OptimisticLocking(DIRTY)` with `@DynamicUpdate`
- `@OptimisticLocking(NONE)` - already works
- `@OptimisticLock(excluded=true)` — handled by Hibernate at AST level
- `OPTIMISTIC_FORCE_INCREMENT` — works via `visitStandardTableUpdate` (standard coordinator path)

**Out of scope (separate tickets):**
- `StatelessSession.upsert()` — stays blocked via `createOptionalTableUpdateOperation`.
- `@SecondaryTable` updates — `isOptional()==true` - throw HIBERNATE-181
- JOINED inheritance updates — `isOptional()==true` - throw HIBERNATE-69
- `@Version @CurrentTimestamp(source=DB)` - DB-generated timestamp, needs `$currentDate` + type check [HIBERNATE-195](https://jira.mongodb.org/browse/HIBERNATE-195)
- `@UserVersionType` -   HIBERNATE-XXXX

**Not applicable:**
- HQL DML (`UPDATE`/`DELETE`) - bypasses persistence context, no optimistic-lock bindings, no path to `createKeyFilter` or `visitOptionalTableUpdate`

HIBERNATE-51


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1178,7 +1188,15 @@ public void visitCustomTableDelete(TableDeleteCustomSql tableDeleteCustomSql) {
 
     @Override
     public void visitOptionalTableUpdate(OptionalTableUpdate optionalTableUpdate) {
```
vbabanin, line 1205: `visitOptionalTableUpdate` exists because Hibernate ORM eagerly builds a merge coordinator for every `@Version` entity during bootstrap. That coordinator produces an `OptionalTableUpdate` and invokes this visitor. `MergeCoordinator` is responsible for handling `upsert`.

We implement this path so bootstrap succeeds. For primary-table version-increment updates, it translates the mutation the same way as `visitStandardTableUpdate`.

The visitor also defensively handles two unsupported optional-table cases:

1. JOINED inheritance
2. @SecondaryTable
3. Upsert

Both are identified by `optionalTableUpdate.getMutatingTable().getTableMapping().isOptional()` and throw `TODO-HIBERNATE-69` and `TODO-HIBERNATE-181`, respectively.

`Upsert` does not reach this visitor. `OptionalTableUpdate.createMutationOperation` routes directly through the dialect factory to MongoDB’s dialect override, which throws` TODO-HIBERNATE-94`.

We have unsupported tests for `JOINED inheritance` and `@SecondaryTable`, but they fail earlier during bootstrap, before `visitOptionalTableUpdate` is reached because they are unsupported in associations as well. As a result, the `isOptional() == true` branch in this visitor is currently not exercised by integration tests.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -371,34 +372,32 @@ public void visitStandardTableUpdate(TableUpdateStandard tableUpdate) {
...
-        return new AstFieldOperationFilter(astFilterFieldPath, new AstComparisonFilterOperation(EQ, fieldValue));
+        var predicates = new ArrayList<AstFilter>(1 + tableMutation.getNumberOfOptimisticLockBindings());
+        predicates.add(createEqualityFiler(tableMutation.getKeyBindings().get(0)));
+        for (var lockBinding : tableMutation.getOptimisticLockBindings()) {
+            predicates.add(createEqualityFiler(lockBinding));
+        }
+        return predicates.size() == 1 ? predicates.get(0) : new AstLogicalFilter(AND, predicates);
+    }
+
+    private AstFieldOperationFilter createEqualityFiler(ColumnValueBinding binding) {
```
jyemin: createEqualityFiler => createEqualityFilter

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1178,7 +1188,15 @@ public void visitCustomTableDelete(TableDeleteCustomSql tableDeleteCustomSql) {
 
     @Override
     public void visitOptionalTableUpdate(OptionalTableUpdate optionalTableUpdate) {
-        throw new FeatureNotSupportedException();
+        if (optionalTableUpdate.getMutatingTable().getTableMapping().isOptional()) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-69 https://jira.mongodb.org/browse/HIBERNATE-69");
+        }
+        var keyFilter = createKeyFilter(optionalTableUpdate);
```
jyemin: Consider pulling the code that this duplicates from `visitStandardTableUpdate` into a common method. The respective params both implement `TableUpdate` interface so it should be possible.

jyemin (changes requested):

src/integrationTest/java/com/mongodb/hibernate/locking/OptimisticLockingIntegrationTests.java
```diff
@@ -0,0 +1,729 @@
...
+        @Table(name = "ItemWithDbTimestamp")
+        static class ItemWithDbTimestamp {
+            @Id
+            int id;
+
+            String string;
+
+            @Version
+            @CurrentTimestamp(source = SourceType.DB)
+            Instant version;
```
strogiyotec, line 572: looks like the not supported exception comes from [here ](https://github.com/mongodb/mongo-hibernate/blob/main/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java#L347)
```
    @Override
    public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
        if (columnWriteFragment.getParameters().size() != 1) {
        }
        columnWriteFragment.getParameters().iterator().next().accept(this);
    }
```
I am not sure if it's the same for other annotations but for CurrentTimestamp specifically do you think we should do 
```
if columnWriteFragment.getFragment().equals("current_timestamp")
   then throw NotSupportedException("CurrentTimestamp is not supported")
```
just for better error message ?

strogiyotec (dismissed):
1 small comment

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Thanks, I looked into this. `OptionalTableUpdate` does not appear to implement `TableUpdate`, so a shared method would need to take the individual inputs rather than a common update type.

I can simplify the standard path without moving result emission into the helper:

```
@Override
public void visitStandardTableUpdate(TableUpdateStandard tableUpdate) {
    if (tableUpdate.getNumberOfReturningColumns() > 0) {
    }
    if (tableUpdate.getWhereFragment() != null) {
    }

    var result = createMutationResult(
            tableUpdate.getValueBindings(),
            tableUpdate.getMutatingTable().getTableName(),
            createKeyFilter(tableUpdate));
    astVisitorValueHolder.yield(MODEL_MUTATION_RESULT, result);
}
```

I would keep` yield(...)` in the top-level visitor methods. Those methods are the Hibernate ORM AST entry points, and the same structure is used for the paths producing `MODEL_MUTATION_RESULT`, `SELECT_RESULT`, and `MUTATION_RESULT`.

Moving yield(...) into a shared helper would hide which visitor callback produces which result. A reader following `MODEL_MUTATION_RESULT` would need to find the helper first, then trace back through its callers to determine whether it came from  top-level`visitStandardTableUpdate()` or `visitOptionalTableUpdate()`. Keeping the yield at the visitor level makes that control flow a bit more explicit.

vbabanin: Commit: [9adb21d](https://github.com/mongodb/mongo-hibernate/pull/170/commits/9adb21da732969693d019f4e259c3a942e933e9f)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Fixed. Commit: [ad67df5](https://github.com/mongodb/mongo-hibernate/pull/170/commits/ad67df5ad02a07611d6fdc4d1c22bfcd397de532)

src/integrationTest/java/com/mongodb/hibernate/locking/OptimisticLockingIntegrationTests.java
vbabanin: Thank you - that makes sense. I added a specific error message for the current_timestamp case, so @CurrentTimestamp now reports that it is not supported.

vbabanin: Commit: [ae26d31](https://github.com/mongodb/mongo-hibernate/pull/170/commits/ae26d312e0614066268f7f16d2ae6d04c1b7f07f)

vbabanin: I eventually reverted the specialized `@CurrentTimestamp` error message after checking the other paths that use `ColumnWriteFragment`.

This visitor is also used for:

1. dialect-supplied expressions, such as current_timestamp from `@CurrentTimestamp(source = DB)`;
2. user-provided expressions, such as` @ColumnTransformer(write = "encrypt(?)")`;
3. generated SQL, such as `@Generated(sql = "...")`.

Therefore, checking whether the fragment equals "current_timestamp" could misclassify user-provided/generated SQL - for example, `@Generated(sql = "current_timestamp")` - and report an inaccurate `@CurrentTimestamp`-specific error.

While tracing this, I also found that non-simple write fragments such as `@ColumnTransformer(write = "test(?)")` were being ignored on this path rather than rejected, that was an existing issue we missed. I added validation so that a fragment that is not a simple `?` parameter now fails explicitly instead of silently passing.

Commit: [ca2b30f](https://github.com/mongodb/mongo-hibernate/pull/170/commits/ca2b30fdae8da9076f3fb586f612550e9c782737)

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoPreparedStatement.java
```diff
@@ -57,6 +56,7 @@ final class MongoPreparedStatement extends MongoStatement implements PreparedSta
             throws SQLSyntaxErrorException {
         super(mongoDatabase, clientSession, mongoConnection);
         MongoAggregateSupport.checkSupported(mql);
+        System.err.println(mql);
```
jyemin: Remove println

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -371,12 +372,17 @@ public void visitStandardTableUpdate(TableUpdateStandard tableUpdate) {
         if (tableUpdate.getWhereFragment() != null) {
             throw new FeatureNotSupportedException();
         }
-        var keyFilter = createKeyFilter(tableUpdate);
-        List<ColumnValueBinding> valueBindings = tableUpdate.getValueBindings();
-        String tableName = tableUpdate.getMutatingTable().getTableName();
+        var murationResult = createMutationResult(
```
jyemin: murationResult -> mutationResult

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoPreparedStatement.java
```diff
@@ -55,6 +55,8 @@ final class MongoPreparedStatement extends MongoStatement implements PreparedSta
             MongoDatabase mongoDatabase, ClientSession clientSession, MongoConnection mongoConnection, String mql)
             throws SQLSyntaxErrorException {
         super(mongoDatabase, clientSession, mongoConnection);
+        MongoAggregateSupport.checkSupported(mql);
```
jyemin: I suspect this is an incorrect merge conflict fix since this method no longer exists.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: I don't think this is worth it anymore for a two-line helper so feel free to revert.

jyemin (changes requested):
Looks like the patch build is failing in CI.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1187,7 +1203,14 @@ public void visitCustomTableDelete(TableDeleteCustomSql tableDeleteCustomSql) {
...
     public void visitOptionalTableUpdate(OptionalTableUpdate optionalTableUpdate) {
-        throw new FeatureNotSupportedException();
+        if (optionalTableUpdate.getMutatingTable().getTableMapping().isOptional()) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-69 https://jira.mongodb.org/browse/HIBERNATE-69");
+        }
+        var murationResult = createMutationResult(
+                optionalTableUpdate.getValueBindings(),
+                optionalTableUpdate.getMutatingTable().getTableName(),
+                createKeyFilter(optionalTableUpdate));
+        astVisitorValueHolder.yield(MODEL_MUTATION_RESULT, murationResult);
```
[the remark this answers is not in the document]
vbabanin: Fixed

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -343,8 +344,9 @@ public void visitStandardTableInsert(TableInsertStandard tableInsert) {
 
     @Override
     public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
-        if (columnWriteFragment.getParameters().size() != 1) {
-            throw new FeatureNotSupportedException();
+        if (!columnWriteFragment.getFragment().equals("?")) {
+            throw new FeatureNotSupportedException(
+                    "@CurrentTimestamp(source=DB), @Generated, and @ColumnTransformer write expressions are not supported");
         }
```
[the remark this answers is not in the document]
vbabanin: `ColumnValueParameters` are constructed one-per-`?` in the fragment, and Hibernate validates at bootstrap that `@ColumnTransformer.write` must contain exactly one `?`

jyemin (commented):
LGTM!

jyemin (approved):

strogiyotec (approved):

PR #171, opened 2026-06-23
HIBERNATE-76 Add support for HQL's `like`

apmasell:
This adds a filter for regular expressions to allow the `like` operator in HSQL to be bound to an MQL regular expression.

HIBERNATE-76: Support HQL pattern matching

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/RegularExpressionIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/RegularExpressionIntegrationTests.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2026-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.query.select;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
+import com.mongodb.hibernate.query.AbstractQueryIntegrationTests;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Stream;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@DomainModel(annotatedClasses = {RegularExpressionIntegrationTests.ItemWithString.class})
+class RegularExpressionIntegrationTests extends AbstractQueryIntegrationTests {
+
+    private static final ItemWithString ITEM1 = new ItemWithString(1, "ab .");
+    private static final ItemWithString ITEM2 = new ItemWithString(2, "AB .");
+    private static final ItemWithString ITEM3 = new ItemWithString(3, "zz%");
+
+    @BeforeEach
+    void beforeEach() {
+        getSessionFactoryScope()
+                .inTransaction(session -> Stream.of(ITEM1, ITEM2, ITEM3).forEach(session::persist));
+        getTestCommandListener().clear();
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"", "i"})
+    void testLikeMatchesNegated(String options) {
+        runBasicNegatedTest(
+                "from Item where str not %slike '%%ab%% .'".formatted(options),
+                "/^.*ab.* \\.$/" + options,
+                options.equals("i") ? List.of(ITEM3) : List.of(ITEM2, ITEM3));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"", "i"})
+    void testLikeMatches(String options) {
+        runBasicTest(
+                "from Item where str %slike '%%ab%% .'".formatted(options),
+                "/^.*ab.* \\.$/" + options,
+                options.equals("i") ? List.of(ITEM1, ITEM2) : List.of(ITEM1));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"", "i"})
+    void testLikeMatchSingle(String options) {
+        runBasicTest(
+                "from Item where str %slike 'a_ .'".formatted(options),
+                "/^a. \\.$/" + options,
+                options.equals("i") ? List.of(ITEM1, ITEM2) : List.of(ITEM1));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"", "i"})
+    void testLikeMatchSingleNegated(String options) {
+        runBasicNegatedTest(
+                "from Item where str not %slike 'a_ .'".formatted(options),
+                "/^a. \\.$/" + options,
+                options.equals("i") ? List.of(ITEM3) : List.of(ITEM2, ITEM3));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"", "i"})
+    void testLikeMatchesWithEscape(String options) {
+        runBasicTest(
+                "from Item where str %slike 'z%%!%%' escape '!'".formatted(options),
+                "/^z.*%$/" + options,
+                List.of(ITEM3));
+    }
+
+    private void runBasicTest(String query, String regex, List<ItemWithString> results) {
+        assertSelectionQuery(
+                query,
+                ItemWithString.class,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$match": {
+                        "str": {
+                          "$regex": %ss
+                        }
+                      }
+                    },
+                    {
+                      "$project": {
+                        "_id": true,
+                        "str": true
+                      }
+                    }
+                  ]
+                }"""
+                        .formatted(regex),
+                results,
+                Set.of(ItemWithString.COLLECTION_NAME));
+    }
+
+    private void runBasicNegatedTest(String query, String regex, List<ItemWithString> results) {
+        assertSelectionQuery(
+                query,
+                ItemWithString.class,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$match": {
+                        "$nor": [
+                          { "str": { "$regex": %ss } }
+                        ]
+                      }
+                    },
+                    {
+                      "$project": {
+                        "_id": true,
+                        "str": true
+                      }
+                    }
+                  ]
+                }"""
+                        .formatted(regex),
+                results,
+                Set.of(ItemWithString.COLLECTION_NAME));
+    }
+
+    @Test
+    void testLikeDoesNotTreatPlusAsRegexQuantifier() {
+        var exactMatch = new ItemWithString(4, "a+b");
+        var shouldNotMatch = new ItemWithString(5, "ab");
+        getSessionFactoryScope().inTransaction(session -> {
+            session.persist(exactMatch);
+            session.persist(shouldNotMatch);
+        });
+
+        var results = getSessionFactoryScope().fromTransaction(session -> session.createSelectionQuery(
+                        "from Item where str like 'a+b'", ItemWithString.class)
+                .getResultList());
+
+        // '+' in a LIKE pattern is a literal character, not a regex quantifier: only the item whose
+        // value is exactly "a+b" should match "a+b", not "ab".
+        assertThat(results).containsExactly(exactMatch);
+    }
+
+    @Test
+    void testLikeDoesNotMatchAcrossEmbeddedNewlines() {
+        var itemWithEmbeddedNewline = new ItemWithString(4, "abc\ndef");
+        getSessionFactoryScope().inTransaction(session -> session.persist(itemWithEmbeddedNewline));
+
+        var results = getSessionFactoryScope().fromTransaction(session -> session.createSelectionQuery(
+                        "from Item where str like 'def%'", ItemWithString.class)
+                .getResultList());
+
+        // The whole value "abc\ndef" does not start with "def", so LIKE 'def%' must not match it, even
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
```diff
@@ -0,0 +1,64 @@
...
+     * Escapes a string in a way that is compatible with the server PCRE-implementation.
+     *
+     * <p>This is preferable to {@link java.util.regex.Pattern#quote(String)} as it matches the server implementation in
+     * <code>src/mongo/util/pcre_util.cpp</code>
+     *
+     * @param text the text to convert to an exact-match equivalent regular expression
+     * @return the regular expression that matches the supplied text
+     */
+    public static String quoteMeta(String text) {
+        final var result = new StringBuilder();
```
strogiyotec: can we preallocate the builder `new StringBuilder(text.length)`
as I understand by the end of the loop it will have the same size as given text ?

src/integrationTest/java/com/mongodb/hibernate/query/select/RegularExpressionIntegrationTests.java
```diff
@@ -0,0 +1,123 @@
...
+    @BeforeEach
+    void beforeEach() {
+        getSessionFactoryScope()
+                .inTransaction(session -> Stream.of(ITEM1, ITEM2, ITEM3).forEach(session::persist));
+        getTestCommandListener().clear();
+    }
+
+    @Test
+    void testLikeMatches() {
+        runBasicTest("from Item where str like '%ab% .'");
```
strogiyotec: will it also work with `not` and `ilike` ? 
https://docs.hibernate.org/orm/7.1/querylanguage/html_single/#like-predicate
Also let's add the test case for `_`(matches any single character)
so basically I believe we should cover
1. like
2. ilike
3. not like
4. not ilike
5. `%`
6. `_`

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionOperationTests.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast.filter;
+
+import static com.mongodb.hibernate.internal.translate.mongoast.AstNodeAssertions.assertRendering;
+
+class AstRegularExpressionOperationTests {
+    void testRendering() {
```
strogiyotec, line 26: this test never runs
I added `@Test` and ran it locally
```
Expected :{"$regex": /pa[t]t?ern*/i}
Actual   :{"$regex": {"$regularExpression": {"pattern": "pa[t]t?ern*", "options": "i"}}}
```

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
```diff
@@ -0,0 +1,64 @@
...
+ * Documents</a>.
+ *
+ * @hidden
+ */
+public record AstRegularExpressionFilterOperation(String pattern, String options) implements AstFilterOperation {
+    /**
+     * Escapes a string in a way that is compatible with the server PCRE-implementation.
+     *
+     * <p>This is preferable to {@link java.util.regex.Pattern#quote(String)} as it matches the server implementation in
+     * <code>src/mongo/util/pcre_util.cpp</code>
```
strogiyotec: sorry I am not too familiar with `cpp`
https://github.com/mongodb/mongo/blob/master/src/mongo/util/pcre_util.cpp
What exactly are we doing here to match whatever mongo server does ?

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1124,7 +1127,20 @@ private record ExistsOverUnnestShape(
 
     @Override
     public void visitLikePredicate(LikePredicate likePredicate) {
```
strogiyotec, line 1152: this logic completely ignores
```
        likePredicate.isCaseSensitive();
        likePredicate.isNegated();
```

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1124,7 +1127,20 @@ private record ExistsOverUnnestShape(
...
-        throw new FeatureNotSupportedException();
+        var escape = '%';
+        if (likePredicate.getEscapeCharacter() != null) {
+            escape = extractLiteral(likePredicate.getEscapeCharacter(), Character.class, "escape character in LIKE");
+        }
+        var pattern = Pattern.compile(Pattern.quote(String.valueOf(escape)))
```
strogiyotec: nit: the type is `String`, should we call it `compiledPattern` ? Otherwise I though it's an instance of `Pattern`

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
apmasell: Updated to use a different string builder.

src/integrationTest/java/com/mongodb/hibernate/query/select/RegularExpressionIntegrationTests.java
apmasell: Added.

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionOperationTests.java
apmasell: Fixed and fixed.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
apmasell: We are converting strings that contain text we want to match exactly into a regular expression. This needs to escape special regex characters, including `.`, `[` and `*`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: This has been refactored. Let me know if it's still a problem.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1133,7 +1136,32 @@ private record ExistsOverUnnestShape(
 
     @Override
     public void visitLikePredicate(LikePredicate likePredicate) {
-        throw new FeatureNotSupportedException();
+        var escape = "[%_]";
+        var allowSingleCharacterSubstiution = true;
```
jyemin: Substiution -> Substitution

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1133,7 +1136,32 @@ private record ExistsOverUnnestShape(
...
+        if (likePredicate.getEscapeCharacter() != null) {
```
strogiyotec, line 1154: I believe this escape character is for HQL `like ... escape '!'` case
https://docs.hibernate.org/orm/4.1/devguide/en-US/html/ch11.html
11.6.3. Like predicate
> like_expression ::=
        string_expression
        [NOT] LIKE pattern_value
        [ESCAPE escape_character]

I wrote a simple integration test and looks like this logic doesn't work
```
    @Test
    void escapedPercentIsLiteral() {
        var literal = new Item(1, "%foo");
        var wildcardLeak = new Item(2, "X%foo");
        seed(literal, wildcardLeak);

        assertThat(query("from Item where str like '!%foo' escape '!'")).containsExactly(literal);
    }

    private void seed(Item... items) {
        getSessionFactoryScope().inTransaction(session -> Stream.of(items).forEach(session::persist));
        getTestCommandListener().clear();
    }

```
Result:
```
Expected :[Item{id=1, str='%foo'}]
Actual   :[Item{id=1, str='%foo'}, Item{id=2, str='X%foo'}]
```

I believe backslash is the right way to do escaping
https://www.mongodb.com/community/forums/t/regex-not-working/125224
I haven't verified it though

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
```diff
@@ -0,0 +1,84 @@
...
+public record AstRegularExpressionFilterOperation(String pattern, String options) implements AstFilterOperation {
+    /**
+     * Escapes a string in a way that is compatible with the server PCRE-implementation.
+     *
+     * <p>This is a reimplementation of {@link java.util.regex.Pattern#quote(String)} that append to a string builder
+     *
+     * @param text the text to convert to an exact-match equivalent regular expression
+     * @param result a string builder that will hold the regular expression that matches the supplied text
+     */
+    public static void quoteMeta(CharSequence text, StringBuilder result) {
```
strogiyotec: can we add unit tests to this method please

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
```diff
@@ -0,0 +1,84 @@
...
+
+        if (text.isEmpty()) {
+            return;
+        }
+
+        result.append("\\Q");
+        final var consumer = new IntConsumer() {
```
strogiyotec: the escaping logic here is hard to follow, I had to google what \Q and \E does
Correct me if I am wrong , the intention here is to translate something like
`like 'a.b'` into `regex: a\.b` because `.` in HQL means dot but in MQL it means any character

If my understanding is correct do you think it's easier to just wrap all escape characters with `\` ? 

I made a small PR to your branch to show what I mean https://github.com/apmasell/mongo-hibernate/pull/1

The same method will be simplified toward
```
  public static void quoteMeta(CharSequence text, StringBuilder result) {
        text.codePoints().forEach(c -> {
            if (isMetacharacter(c)) {
                result.append('\\');
```
I ran all integration tests(considering I had to change the expected string regex) and they passed

What do you think @apmasell  ?

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: Okay. I think i implemented the wrong semantics for this operation.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
apmasell: This is the escaping semantics that Java uses for generating regular expressions. What would be the reference for `isMetacharacter` because the PCRE documentation doesn't provide a canonical list.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public static String quoteMeta(String pattern, @Nullable Character escape) {
+        var result = new StringBuilder(pattern.length());
+        result.append("^");
+        for (var index = 0; index < pattern.length(); index++) {
+            var current = pattern.charAt(index);
+            if (escape != null && current == escape) {
+                if (index == pattern.length() - 1) {
+                    throw new IllegalArgumentException(
+                            "Escape character %s found at end of LIKE pattern: %s".formatted(escape, pattern));
```
strogiyotec, line 47: let's remove this check 
I ran the same case against H2 database that we use for testing
```
        var literal = new Item(1, "foo");
        var wildcardLeak = new Item(2, "X%foo");
        seed(literal, wildcardLeak);

        assertThat(query("from Item where str like 'foo!' escape '!'")).isEmpty();
```
H2 didn't throw an exception

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
strogiyotec: The meta character logic was copied from cpp server implementation [here ](https://github.com/mongodb/mongo/blob/ac4b289b1f9c2cf972078fae41f619a10f8d4e94/src/mongo/util/pcre_util.cpp#L86)
```
        if (!ctype::isAlnum(c) && c != '_' && !(c & 0x80))
```

so my java implementation just copies the same semantic , I didn't find if it's documented somewhere @jyemin  if you know

@apmasell  I saw you changed the implementation and introduced the list , I suggest we keep the same logic as the server does by comparing the ascii range instead of checking if character is in this list, what do you think ?

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
apmasell: If you want. I think it's an invalid input.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
apmasell: This code has changed enough that I don't know what to think. This was the preferred solution discussed in the meeting. It's worth pointing out that Java is using UTF-16 encoded strings while C++ will be using UTF-8, so we can't copy this directly anyway.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/filter/AstRegularExpressionFilterOperation.java
strogiyotec: confirmed postgres does the same, thanks

strogiyotec (approved):

jyemin (approved):
LGTM as well

PR #173, opened 2026-06-29
HIBERNATE-56: throw when @Id column is explicitly mapped to a name other than _id

jyemin:
MongoDB requires the primary key field to be named `_id`. Previously, any explicitly-set `@Id` column name was silently overridden. This change detects the misconfiguration at bootstrap and throws `FeatureNotSupportedException` with an actionable message.

## Scope

| Mapping style | Covered |
|---|---|
| `@Column(name = "<anything but _id>")` on `@Id` field | ✅ |
| JPA orm.xml `<column name="..."/>` override | ✅ (Hibernate merges orm.xml into the same annotation model before bootstrap validation runs) |
| Legacy HBM XML | ❌ — deprecated, removed in Hibernate 8.0; `MemberDetails` is null for HBM-sourced mappings so there is no clean detection mechanism |

## Implementation

Guard added in `MongoAdditionalMappingContributor.setIdentifierColumnName`. Before the unconditional rename to `_id`, if the column name differs and the identifier is a `SimpleValue` with non-null `MemberDetails` carrying a `@Column` annotation with a non-blank name, it throws:

```
EntityWithBadId: the @Id column name cannot be overridden to [customId];
MongoDB requires the primary key field to be named [_id]
— remove @Column(name = "customId") or change it to @Column(name = "_id")
```

`hibernate-models` is added as an explicit `implementation` dependency (it was already a transitive dep via `hibernate-core`; no version pin needed) and `requires org.hibernate.models` added to `module-info.java`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -146,8 +146,9 @@ dependencies {
     errorprone(libs.nullaway)
     errorprone(libs.google.errorprone.core)
 
+    implementation(platform(libs.hibernate.platform))
     api(libs.hibernate.core)
-    implementation("org.hibernate.models:hibernate-models")
+    implementation(libs.hibernate.models)
     api(libs.mongo.java.driver.sync)
     // We need the `libs.findbugs.jsr` dependency to stop `javadoc` from emitting
     // `warning: unknown enum constant When.MAYBE`
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -147,6 +147,7 @@ dependencies {
     errorprone(libs.google.errorprone.core)
 
     api(libs.hibernate.core)
+    implementation("org.hibernate.models:hibernate-models")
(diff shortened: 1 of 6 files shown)


apmasell (dismissed):
The code looks good. Not sure what's going on with the failing JavaDoc build or why this would have triggered it.

jyemin:
> The code looks good. Not sure what's going on with the failing JavaDoc build or why this would have triggered it.

OK, got to the bottom of this.  The Javadoc build failure is unrelated, and flaky, and there's a PR up to fix it.

But it was masking a real problem.  On a run where the Javadoc build failure _didn't_ happen, the smoke test exposed a real issue with our generated pom.xml.  That issue is now fixed.  PTAL.

apmasell (approved):

PR #174, opened 2026-07-01
HIBERNATE-183: require com.mongodb.hibernate.semantics.nulls=MQL at bootstrap

jyemin:
Adds a required Hibernate bootstrap configuration property, `com.mongodb.hibernate.semantics.nulls`, with `MQL` as the only supported value for now. This defers the decision on SQL vs. MQL null-comparison semantics without picking a silent default — every user must declare their expectation explicitly.

Wires the new requirement through every existing bootstrap site so nothing else breaks: unit test fixtures, integration test fixtures, the Spring Boot autoconfigure module's integration test fixture, the example-module apps/smoke test, and the README's Spring Boot config snippet.

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
+++ b/src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
@@ -176,6 +176,7 @@ public final class StandardServiceRegistryScopedState implements Service {
                         JAKARTA_JDBC_URL, MongoConfigurationContributor.class.getName()));
             }
             forbidTemporalConfiguration(configurationValues);
+            checkNullSemantics(configurationValues);
             var mongoConfigurationBuilder = new MongoConfigurationBuilder(configurationValues);
             if (mongoConfigurationContributor != null) {
                 mongoConfigurationContributor.configure(mongoConfigurationBuilder);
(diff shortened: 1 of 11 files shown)


src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
```diff
@@ -212,5 +213,18 @@ private static void forbidTemporalConfiguration(Map<String, Object> configuratio
...
             }
         }
+
+        private static void checkNullSemantics(Map<String, Object> configurationValues) {
+            var nullSemantics = configurationValues.get("com.mongodb.hibernate.semantics.nulls");
+            if (nullSemantics == null) {
+                throw new HibernateException(
+                        "Configuration property [com.mongodb.hibernate.semantics.nulls] is required");
+            }
+            if (!"MQL".equals(nullSemantics)) {
```
vbabanin, line 223: [nit] Should we make this value an enum?

It would only have one value for now, so it’s not required, but I think it would make the intended extension point clearer in the code.

apmasell (approved):

vbabanin (approved):
LGTM!

src/main/java/com/mongodb/hibernate/internal/service/StandardServiceRegistryScopedState.java
jyemin: I'm going to leave it and we can reconsider when we add SQL mode.

PR #177, opened 2026-07-07
HIBERNATE-69: Support mapped superclasses and single-table inheritance

apmasell:
- `MappedSuperclass`: This adds tests to ensure that the `@MappedSuperclass` inheritance type works. It required no additional modifications to work as it does not require any database behaviour to handle inheritance.
- Single-table:  Add support for inheritance hierarchies that use a single table for all subclasses.

Support for table-per-class and joined inheritance requires other features and will be delivered separately.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/boot/NativeBootstrappingIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/boot/NativeBootstrappingIntegrationTests.java
@@ -19,15 +19,31 @@ package com.mongodb.hibernate.boot;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.mongodb.client.MongoClient;
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
 import com.mongodb.hibernate.junit.InjectMongoClient;
 import com.mongodb.hibernate.junit.MongoExtension;
+import com.mongodb.hibernate.junit.MongoServiceRegistryProducer;
+import com.mongodb.hibernate.query.AbstractQueryIntegrationTests;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import jakarta.persistence.Version;
+import java.time.Instant;
 import org.bson.BsonDocument;
+import org.hibernate.annotations.CurrentTimestamp;
+import org.hibernate.annotations.Generated;
+import org.hibernate.annotations.GeneratedColumn;
+import org.hibernate.annotations.SourceType;
 import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(MongoExtension.class)
-class NativeBootstrappingIntegrationTests {
+class NativeBootstrappingIntegrationTests extends AbstractQueryIntegrationTests
+        implements MongoServiceRegistryProducer {
     @InjectMongoClient
     private static MongoClient mongoClient;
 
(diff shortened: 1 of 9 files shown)


src/integrationTest/java/com/mongodb/hibernate/query/select/InheritanceIntegrationTests.java
```diff
@@ -0,0 +1,488 @@
...
+            MAPPED_A.forEach(session::persist);
+            session.persist(SINGLE_INT_A);
+            session.persist(SINGLE_INT_B);
+            session.persist(SINGLE_STR_A);
+            session.persist(SINGLE_STR_B);
+        });
+        getTestCommandListener().clear();
+    }
+
+    @Test
```
jyemin, line 72: Our typical practice is to nest positive tests into Nested classes, e.g.

```
@Nested
class SingleInheritanceTests implements MongoServiceRegistryProducer
```

and negative tests 

```
@Nested
class Unsupported implements MongoServiceRegistryProducer
```

src/integrationTest/java/com/mongodb/hibernate/query/select/InheritanceIntegrationTests.java
```diff
@@ -0,0 +1,488 @@
...
+                    {
+                      "$match": {
+                        "uniqueOnA": {"$gt": 64}
+                      }
+                    },
+                    {
+                      "$project": {
+                        "_id": true,
+                        "commonField": true,
+                        "uniqueOnA": true,
```
jyemin: The trailing comma works but it's dependent on what's arguably a bug in the Java driver's `JsonReader` that we shouldn't rely on.  I suggest removing all the trailing comma, which are not legal JSON.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -348,10 +352,23 @@ public void visitStandardTableInsert(TableInsertStandard tableInsert) {
 
     @Override
     public void visitColumnWriteFragment(ColumnWriteFragment columnWriteFragment) {
-        if (columnWriteFragment.getParameters().size() != 1) {
-            throw new FeatureNotSupportedException();
+
```
jyemin: Nit: our convention is to have no blank lines at the beginning of methods.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -348,10 +352,23 @@ public void visitStandardTableInsert(TableInsertStandard tableInsert) {
...
+        switch (columnWriteFragment.getParameters().size()) {
+            case 0 -> {
+                // Hibernate inheritance discriminators should be string/char or integer. It is possible that other
+                // fragments are generated with other types through other code paths.
+                var matcher = SQL_STRING.matcher(columnWriteFragment.getFragment());
+                BsonValue result;
+                if (matcher.matches()) {
+                    result = new BsonString(matcher.group(1).replace("''", "'"));
+                } else {
+                    result = new BsonInt64(Long.parseLong(columnWriteFragment.getFragment()));
```
jyemin: Let's do

```
new BsonInt32(Integer.parseInt(columnWriteFragment.getFragment()))
```

since Hibernate rejects anything that's not an integer anyway, and using `int32` saves a bit on storage size and also matches the BSON type that will be used to query the discriminator (though MQL semantics for integral comparisons ensures it works as is).

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -348,10 +352,23 @@ public void visitStandardTableInsert(TableInsertStandard tableInsert) {
...
+                }
+                astVisitorValueHolder.yield(VALUE, new AstLiteral(result));
+            }
+            case 1 -> columnWriteFragment.getParameters().iterator().next().accept(this);
+            default -> throw new FeatureNotSupportedException();
```
jyemin: Afaik there is no longer an unsupported feature here, so let's change this to

```
throw fail("unexpected multi-parameter discriminator write fragment: " + columnWriteFragment)
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -230,6 +233,7 @@ public abstract class AbstractMqlTranslator<T extends JdbcOperation> implements
     // '#' is blocked in mapped field names, so prefixing join aliases with it prevents $lookup from shadowing
     // a local field that happens to share the Hibernate-generated alias name (e.g. "o1_0").
     private static final String JOIN_ALIAS_PREFIX = "#";
+    private static final Pattern SQL_STRING = Pattern.compile("^ *'((?:''|[^'])*)' *$");
```
jyemin: Hibernate will not produce a string with leading/trailing spaces, so we can simplify the regex to 

```
^'((?:''|[^'])*)'$
```

This matches the assumption for integers, where we also don't trim whitespace before parsing.

jyemin: Regular expressions are not so easy for (most) humans to parse, so worth a comment saying what this matches: a single-quoted string with interior quotes doubled and no surrounding whitespace.

jyemin: I think the regex is correct, but let's add at least one regression test.  Something like:

```
@DiscriminatorValue("a'b")
```

?

src/integrationTest/java/com/mongodb/hibernate/query/select/InheritanceIntegrationTests.java
```diff
@@ -0,0 +1,488 @@
...
+
+@DomainModel(
+        annotatedClasses = {
+            InheritanceIntegrationTests.MappedA.class,
+            InheritanceIntegrationTests.SingleIntA.class,
+            InheritanceIntegrationTests.SingleIntB.class,
+            InheritanceIntegrationTests.SingleStrA.class,
+            InheritanceIntegrationTests.SingleStrB.class
+        })
+public class InheritanceIntegrationTests extends AbstractQueryIntegrationTests {
```
jyemin, line 48: Let's add a test for `DiscriminatorType#CHAR`.  I'm pretty sure it maps to the same thing as DiscriminatorType#STRING`, but let's prove it with a test.

jyemin (changes requested):

jyemin:
Please update the HIBERNATE-69 description to match what we're doing in this PR, and file/link follow-up tickets for the remaining work (joined-table and table-per-class) so we don't lose track of it.

apmasell:
Will do on the tickets. I have partial tests for those I intend to include in the tickets, but I will need to rebase on the finished version of this before I can do that.

jyemin (dismissed):
LGTM!

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -224,6 +228,19 @@ private static void forbidEmbeddablesWithoutPersistentAttributes(InFlightMetadat
...
     }
 
+    /**
+     * Forbid usage of {@link org.hibernate.annotations.Generated} and
+     * {@link org.hibernate.annotations.CurrentTimestamp} annotations.
+     */
+    private static void forbidColumnFragmentAnnotations(PersistentClass persistentClass) {
+        ClassElementChecker.check(
+                persistentClass,
+                ClassElementChecker.forbid(Generated.class),
```
jyemin, line 208: The three `ClassElementChecker.forbid` checks are not tested (confirmed by commenting them out and seeing that no tests fail).  If they're important, let's add regression tests.  Otherwise, let's remove them.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Consider refactoring this code to use the new `ClassElementChecker` in order to DRY this up.

apmasell: Refactored.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
apmasell: Added. Interestingly, in the refactored code, the equivalent code path was also not tested.

jyemin: Looks like you added `ClassElementChecker` unit tests, which is great, but to test the actual forbidding of  these `Generated*` annotations you'll need to create integration tests that attempt to load entities that use these annotations.  They would be similar to the `JdbcTypeCodeIntegrationTests` that you had to modify as part of the refactoring.

jyemin (changes requested):

apmasell:
Those tests are pre-existing, as far as I understand. They were covered by the column fragment writer tests.

src/integrationTest/java/com/mongodb/hibernate/locking/OptimisticLockingIntegrationTests.java
```diff
@@ -533,7 +536,40 @@ void testVersionWithCurrentTimestampSourceDbThrows() {
                             .buildSessionFactory())
                     .isInstanceOf(FeatureNotSupportedException.class)
                     .hasMessage(
-                            "@CurrentTimestamp(source=DB), @Generated, and @ColumnTransformer write expressions are not supported");
+                            "Field version of class com.mongodb.hibernate.locking.OptimisticLockingIntegrationTests.Unsupported.ItemWithDbTimestamp is not supported: Annotation @CurrentTimestamp(source=DB) is forbidden");
+        }
+
+        @Test
```
jyemin: These tests are exactly what I was looking for, but they seem mis-named and also mis-placed.  

* They are unrelated to versioning so `testVersionWith...` is not accurate.  Rename to something like `testGeneratedAnnotationForbidden`
* Since they are unrelated to optimistic locking, let's move them all to a place related to bootstrapping.  `NativeBootstrappingIntegrationTests` is closest, or you could add a new file.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Looks like this was done.

src/integrationTest/java/com/mongodb/hibernate/boot/NativeBootstrappingIntegrationTests.java
```diff
@@ -53,4 +69,104 @@ void testCouldNotInstantiateDialectExceptionMessage() {
...
+                            .addAnnotatedClass(NativeBootstrappingIntegrationTests.Unsupported.ItemWithGenerated.class)
+                            .buildMetadata(new StandardServiceRegistryBuilder().build())
+                            .buildSessionFactory())
+                    .isInstanceOf(FeatureNotSupportedException.class)
+                    .hasMessage(
+                            "Field version of class com.mongodb.hibernate.boot.NativeBootstrappingIntegrationTests.Unsupported.ItemWithGenerated is not supported: Annotation org.hibernate.annotations.Generated is forbidden");
+        }
+
+        @Test
+        void testForbidGenerateValueAnnotation() {
```
jyemin: Typo: testForbidGenerateValueAnnotation -> testForbidGeneratedValueAnnotation

jyemin (changes requested):
Just one typo.  Sorry missed it last round.

jyemin (approved):

PR #178, opened 2026-07-07
HIBERNATE-73: forbid unsupported BSON and UUID domain types at boot

jyemin:
## Summary

Forbid unsupported `Serializable` BSON/UUID Java types from being mapped as Hibernate domain (entity/embeddable/collection-element) attribute types. Such types (e.g. `org.bson.Document`, `BsonDocument`, `org.bson.types.*`, `java.util.UUID`) are `Serializable`, so today Hibernate silently binary-serializes them to a BSON binary (subtype 0). That "works" but stores a representation that we're not likely to choose if we ever do support any of these types— and fixing it after release would be a breaking change.

This makes those mappings fail fast at metadata-build time with a `FeatureNotSupportedException`, so introducing support later is a feature addition rather than a breaking change.

## Approach

Extends the existing boot-time checker `MongoAdditionalMappingContributor`, which already walks every persistent attribute (recursing into embeddables and collection/array element types) and rejects unsupported types. The forbidden classes are added to its `UNSUPPORTED_TYPES` set and the now-misnamed `forbidTemporalTypes` check is renamed to `forbidUnsupportedTypes`.

**Forbidden** (all `Serializable`, would otherwise be silently persisted):
- `org.bson.types`: `BSONTimestamp`, `Binary`, `Code`, `CodeWithScope`, `CodeWScope`, `MinKey`, `MaxKey`, `Symbol`, `Decimal128`
- `org.bson`: `Document`, `BsonDocument`, `RawBsonDocument`, `BsonDocumentWrapper`
- `java.util.UUID`

`ObjectId` and `BigDecimal` (stored as `Decimal128`) are still supported.

**Notes**
- The forbidding is by domain type: an `AttributeConverter` does not bypass it (matching the existing temporal-type behavior).
- Non-`Serializable` BSON types (`BinaryVector`, other `BsonValue` subtypes, `JsonObject`) already fail at boot via Hibernate's own `JdbcTypeRecommendationException`, so they are out of scope here.

## Out of scope
- Read-side assertion rejecting unsupported BSON binary subtypes (anything besides 0).
- Support for `UUID`, `Decimal128`, `Binary`, `BinaryVector`, etc.


src/integrationTest/java/com/mongodb/hibernate/type/temporal/CalendarIntegrationTests.java
```diff
@@ -58,12 +56,4 @@ void unsupported() {
                         CalendarItems.WithNestedAggregateEmbeddableWithCollectionPersistentAttribute.class),
                 () -> assertNotSupported(CalendarItems.WithNestedCollectionOfAggregateEmbeddable.class));
     }
-
-    // package-private because it's used by all the tests in this package
-    static void assertNotSupported(Class<?> entityClass) {
```
jyemin, line 63: Moved this to its own class as it's awkward placed here

apmasell (dismissed):

apmasell (approved):

PR #179, opened 2026-07-08
HIBERNATE-86 support expressions in SELECT and WHERE

jyemin:
## Summary

Adds support for expressions in HQL `SELECT` and `WHERE` clauses — arithmetic and comparison operators, and the scalar predicate family — translated to MongoDB aggregation expressions.

**SELECT** — computed expressions are projected via aggregation expression operators in `$project`. The AS alias is used as the projection key when present; otherwise a generated `#c_N` key is used (`#` is blocked in mapped field names, preventing collisions).

**WHERE** — when either side of a comparison is not a simple field-or-value, the predicate is wrapped in `$expr`. Plain field comparisons continue to use the existing `{field: {$op: value}}` form. Each predicate independently decides whether it needs `$expr`; AND/OR junctions compose naturally.

### Supported

**Arithmetic** — `+`, `-`, `*`, `/`, `div`, unary negation/plus. Integer division truncates via `$toInt($divide)` (`$toLong` for a `BIGINT` result) since MongoDB's `$divide` always returns a double — one rule for `DIVIDE`/`QUOT`/`DIVIDE_PORTABLE`, independent of `PORTABLE_INTEGER_DIVISION`.

**Criteria API** — `CriteriaBuilder.quot` → `QUOT` (behaviorally identical to `DIVIDE`) and `CriteriaBuilder.mod` → the `MODULO` operator → `$mod`. HQL's `%` instead becomes a `mod()` function call (not the `MODULO` operator) and remains unsupported (HIBERNATE-196).

**Comparison** — `>`, `>=`, `<`, `<=`, `=`, `<>` — as a value in SELECT (`select x > 1`) and with computed operands in WHERE via `$expr` (`where x + 1 > 5`, `where x > y`).

**Predicate family** (as value-producing aggregation expressions — in a SELECT item or as an operand of another predicate):

- AND / OR junctions → `$and` / `$or`
- NOT → `$not` (a one-element array)
- IN list → `$in`; negated wrapped in `$not`
- LIKE → `$regexMatch`; negated wrapped in `$not`
- IS NULL / IS NOT NULL → `$eq` / `$ne` against `null`, including on a computed operand (`select (x + y) is null`)
- BETWEEN / NOT BETWEEN → `$and` of `$gte`/`$lte` bounds (negated: `$or` of `$lt`/`$gt`). Extends the compact `$match` filter form (HIBERNATE-89): a computed operand or bound falls back to `$expr`, with the AND/OR hoisted to the filter level like any other computed comparison.
- boolean fields → `$eq` against `true`/`false`

### Unsupported (tracked separately)

- Function calls as operands, including HQL `%` / `mod()`: HIBERNATE-196
- CASE expressions: HIBERNATE-83
- `EXISTS` in SELECT (match-only `$elemMatch`, no scalar form) and IN-subquery — throw `FeatureNotSupportedException`

### Design

- **A value is not an expression; it is converted to one.** `AstValue` does **not** extend `AstExpression` — a value renders differently in aggregation-expression position (a `$`-prefixed string is a field path there, a document/array is an operator invocation). A value used as an expression goes through **`AstValueExpression`**, which wraps it in `{$literal: …}` when it would otherwise be misread. Treating every value as an expression silently mistranslated, e.g. `select name = '$foo'` → `{$eq: ["$name", "$foo"]}` where `"$foo"` is read as a field reference.
- Expression translation goes through the same `SqlAstWalker` visitor dispatch as the rest of the translator, rather than a separate recursive method. A caller requests an expression with `acceptAndYield(node, EXPRESSION)`; each node's visit method yields an `AstExpression` and recurses into its operands the same way. Nodes that render differently by context (a column reference, a literal/parameter, a predicate) branch on `AstVisitorValueHolder.expects(EXPRESSION)` — yielding the `$expr` expression form in expression context and the `$match` filter form otherwise.
- **New predicate-expression nodes:** an `AstLogicalOperator` enum + `AstLogicalOperatorExpression` (`$and`/`$or`/`$not`), mirroring `AstLogicalFilter`/`AstLogicalFilterOperator` on the filter side; `AstInExpression` (`$in`); `AstRegexMatchExpression` (`$regexMatch`). IS NULL and boolean-field predicates are genuine equalities, so they reuse `AstBinaryOperatorExpression` with `$eq`/`$ne` — not false sharing, but what those predicates mean. BETWEEN adds no new node: it desugars to `AstLogicalOperatorExpression` (`$and`/`$or`) over two `AstBinaryOperatorExpression` bounds.
- `MongoStatement.isExcludeProjectSpecification` updated to treat document-valued `$project` specs as inclusions (aggregation expressions render as documents).

There's both a spec and a design note (a shorter version of the spec that is more like a requirements spec). Looking for feedback on which is more useful.

## Test plan

- Unit tests for all new AST node types: `AstFieldPathExpression`, `AstBinaryOperatorExpression`, `AstUnaryOperatorExpression`, `AstValueExpression`, `AstProjectStageExpressionSpecification`, `AstExprFilter`, `AstLogicalOperatorExpression`, `AstInExpression`, `AstRegexMatchExpression`
- `ExpressionIntegrationTests` covering all supported arithmetic/comparison operators, nested arithmetic, AS alias, named parameter operands, unary negation/plus, field-vs-field comparison, mixed AND predicates, both-sides-computed predicates, alias correctness, `$literal` wrapping (a `$`-prefixed string literal and a string parameter), integer division (`$toInt`), long division (`$toLong`, separate nested class), portable integer division (separate nested class with its own `@ServiceRegistry`), and Criteria `quot`/`mod` (separate nested class)
- `PredicateExpressionIntegrationTests` covering AND/OR junctions, NOT (as an operand), boolean field and negated boolean field, IN list (numbers, strings, negated, and a `$`-prefixed option that gets `$literal`-wrapped), LIKE / NOT LIKE, IS NULL / IS NOT NULL, computed-operand IS NULL, BETWEEN / NOT BETWEEN (incl. computed- and parameter-operand cases), and composed predicates. `ExpressionIntegrationTests` also covers a computed-operand BETWEEN in WHERE (`$expr` fallback)
- Negative tests for CASE expressions (HIBERNATE-83), function call operands (HIBERNATE-196), HQL `%` (HIBERNATE-196), IN-subquery, tuple comparison, and EXISTS in SELECT


src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstExpression.java
```diff
@@ -0,0 +1,26 @@
...
+package com.mongodb.hibernate.internal.translate.mongoast;
+
+/**
+ * An aggregation-pipeline expression — a field path, literal, parameter, or operator application — rendered in value
+ * position (a {@code $project} field value or an operand of another expression).
+ *
+ * @hidden
+ */
+@SuppressWarnings("MissingSummary")
+public interface AstExpression extends AstNode {}
```
apmasell, line 26: Given that the `AstFilter` and `AstExpression` are not interconvertible, does it make sense to keep `AstNode` as a super type? I only see it used for testing.

jyemin: `AstNode` is the interface that declares the `render(BsonWriter)` method, so without this supertype expressions can't be rendered.

apmasell: Right. But pulling that down into `AstFilter` and `AstExpression` with different names, would allow the implementations to be shared where possible. In some sense, `BsonWriter` for `AstFilter` and `AstExpression` is two different types of `BsonWriter` because they are at two different insertion contexts in the document. I guess I'm asking if it's worth trying to differentiate those two effective types of `BsonWriter` even though we can't change the type of `BsonWriter` itself.

jyemin: Ah, I see what you mean now.  I would prefer to wait until we have a scenario that makes it useful, unless you can think of one for this PR.

apmasell: Well, I'm wondering if we can share the match predicate implementations to avoid duplication, but if that's better to be done later, that's good too.

jyemin: I'm open to DRYing it up, but It's not clear to me what exactly can be shared.

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
```diff
@@ -0,0 +1,868 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@DomainModel(annotatedClasses = {ExpressionIntegrationTests.Item.class, ExpressionIntegrationTests.LongItem.class})
+class ExpressionIntegrationTests extends AbstractQueryIntegrationTests {
+
+    @Nested
+    class Positive implements MongoServiceRegistryProducer {
+
+        @BeforeEach
```
strogiyotec, line 48: let's add an example with static variables
```

package com.mongodb.hibernate.query.select;


public final class HqlConstants {

    public static final int BONUS = 100;

    private HqlConstants() {}
}
```
and test
```
        @Test
        void testStaticConstantOperand() {
            assertSelectionQuery(
                    "select x + com.mongodb.hibernate.query.select.HqlConstants.BONUS from Item",
                    Integer.class,
                    """
                    {
                      "aggregate": "items",
                      "pipeline": [
                        {
                          "$project": {
                            "#c_1": {"$add": ["$x", {"$numberInt": "100"}]}
                          }
                        }
                      ]
                    }""",
                    List.of(110, 104, 105),
                    Set.of(Item.COLLECTION_NAME));
        }

```

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstValueExpression.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A {@link AstValue} (a literal or parameter) used in aggregation-expression position. A value and an expression do not
+ * always render the same: in expression position a string beginning with {@code $} is a field path, and a
+ * document/array is an operator invocation, so such a value must be wrapped in {@code $literal} to be taken verbatim.
+ * {@code literalWrapped} says whether that wrapping is needed.
+ *
+ * @hidden
+ */
+@SuppressWarnings("MissingSummary")
+public record AstValueExpression(AstValue value, boolean literalWrapped) implements AstExpression {
```
strogiyotec: I am thinking about making a two helper methods here 
```
  public record AstValueExpression(AstValue value, boolean literalWrapped) implements AstExpression {
      static AstValueExpression bare(AstValue value)          { return new AstValueExpression(value, false); }
      static AstValueExpression literalWrapped(AstValue value){ return new AstValueExpression(value, true); }
      // render unchanged
  }
```
and the caller will be `AstValueExpression.bare(...)`
what do you think ? For me it would have been easier instead of checking what this boolean flag does every time 
(just a suggestion feel free to ignore it )

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1193,27 +1462,49 @@ public void visitLikePredicate(LikePredicate likePredicate) {
...
 
     @Override
     public void visitNullnessPredicate(NullnessPredicate nullnessPredicate) {
         var expression = nullnessPredicate.getExpression();
-        if (!isFieldPathExpression(expression)) {
-            throw new FeatureNotSupportedException(
-                    "Only the following nullness predicates are supported: field is [not] null");
+        if (astVisitorValueHolder.expects(EXPRESSION)) {
+            var operand = acceptAndYieldExpression(expression);
+            var operator = nullnessPredicate.isNegated() ? "$ne" : "$eq";
```
strogiyotec: can we use enum here ? 
```
            var operator = nullnessPredicate.isNegated() ? NE : EQ;
```
but then `getOperatorName` should be public

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
```diff
@@ -0,0 +1,868 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@DomainModel(annotatedClasses = {ExpressionIntegrationTests.Item.class, ExpressionIntegrationTests.LongItem.class})
+class ExpressionIntegrationTests extends AbstractQueryIntegrationTests {
+
+    @Nested
+    class Positive implements MongoServiceRegistryProducer {
+
```
strogiyotec, line 47: let's also add this test case for selection in `where not`
```
        @Test
        void testNotComputedComparisonInWhere() {
            assertSelectionQuery(
                    "from Item where not (x + 1 > 5)",
                    Item.class,
                    """
                    {
                      "aggregate": "items",
                      "pipeline": [
                        {
                          "$match": {
                            "$nor": [{"$expr": {"$gt": [{"$add": ["$x", {"$numberInt": "1"}]}, {"$numberInt": "5"}]}}]
                          }
                        },
                        {
                          "$project": {
                            "_id": true,
                            "x": true,
                            "y": true
                          }
                        }
                      ]
                    }""",
                    List.of(new Item(2, 4, 7)),
                    Set.of(Item.COLLECTION_NAME));
        }

```

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
```diff
@@ -0,0 +1,868 @@
...
+import org.hibernate.testing.orm.junit.Setting;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@DomainModel(annotatedClasses = {ExpressionIntegrationTests.Item.class, ExpressionIntegrationTests.LongItem.class})
+class ExpressionIntegrationTests extends AbstractQueryIntegrationTests {
+
+    @Nested
+    class Positive implements MongoServiceRegistryProducer {
```
strogiyotec, line 46: let's also add this test case
```
        @Test
        void testDivideByZeroFailsAtExecution() {
            assertThatThrownBy(() -> getSessionFactoryScope().inTransaction(session -> session.createSelectionQuery(
                                    "select x / 0 from Item", Integer.class)
                            .getResultList()))
                    .isInstanceOf(JDBCException.class)
                    .hasRootCauseInstanceOf(MongoCommandException.class)
                    .rootCause()
                    .hasMessageContaining("can't $divide by zero");
        }

```

strogiyotec (changes requested):

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
jyemin: Done.

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/query/select/ExpressionIntegrationTests.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstValueExpression.java
jyemin: I also detected a smell here, but I decided to address it slightly differently. Instead of introducing the static factory pattern, I split it into two records

```
record AstValueExpression(AstValue value)
record AstLiteralExpressinon((AstValue value)
```

and removed the conditional logic from `AstValueExpression`.  I think this is truer to the overall AST design, where each expression gets its own node.

(d3bd1fe)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Agreed on killing the magic strings — done, though a bit differently than suggested. 

NE/EQ are the filter-side enum (`AstComparisonFilterOperator`, for the `{field:{$op:value}}` query form); reusing them in expression position would mix the two contexts and force `getOperatorName()` to become public, as you pointed out.

Instead I added expression-side peers — `AstComparisonExpressionOperator` (plus `AstArithmeticExpressionOperator` and `AstConversionExpressionOperator`) — and gave `AstBinaryOperatorExpression`/`AstUnaryOperatorExpression` delegating constructors that take them.

Every operator string in the translator — comparison, arithmetic, and the `$toInt`/`$toLong` casts — is now a typed constant. (ac07885)

strogiyotec (dismissed):
LGTM

strogiyotec (approved):

PR #180, opened 2026-07-08
HIBERNATE-184 add copilot instructions

strogiyotec:
[HIBERNATE-184](https://jira.mongodb.org/browse/HIBERNATE-184)


jyemin (approved):
LGTM.   Let's see if it does the trick in taming Copilot.

PR #181, opened 2026-07-08
HIBERNATE-89: Support HQL BETWEEN predicate

apmasell:
Adds support to generate MQL for HQL's `BETWEEN` operation

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
+++ b/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
@@ -1047,7 +1047,29 @@ public abstract class AbstractMqlTranslator<T extends JdbcOperation> implements 
 
     @Override
     public void visitBetweenPredicate(BetweenPredicate betweenPredicate) {
-        throw new FeatureNotSupportedException();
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1031,7 +1031,29 @@ public void visitModifiedSubQueryExpression(ModifiedSubQueryExpression modifiedS
 
     @Override
     public void visitBetweenPredicate(BetweenPredicate betweenPredicate) {
-        throw new FeatureNotSupportedException();
+        if (!isFieldPathExpression(betweenPredicate.getExpression())
```
strogiyotec, line 1050: could you add negative(assertThrows) test cases in SimpleSelectQueryIntegrationTests  for these unsupported cases

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: Added.

strogiyotec (approved):
LGTM

PR #182, opened 2026-07-09
HIBERNATE-128: Enable automatic release publishing and fix release workflow

jyemin:
Rounds out the release automation so the next alpha can be cut end-to-end.

- **Auto-publishing** — `publish.sh` now runs `closeAndReleaseSonatypeStagingRepository` instead of `closeSonatypeStagingRepository`, so a release publishes to Maven Central automatically rather than needing a manual release step.
- **Version convention** — `release.yml` assumes the dev snapshot in `gradle.properties` is `X.Y.Z-SNAPSHOT`, with the pre-release counter living only in the release input and git tag (same model as mongo-java-driver). It was `1.0.0-alpha1-SNAPSHOT`, which fails the snapshot-version check when releasing a pre-release. Changed to `1.0.0-SNAPSHOT`.
- **Release title** — the drafted GitHub release title was `Java Driver <version>` (copied from mongo-java-driver); corrected to `MongoDB Extension for Hibernate ORM <version>`.

### DIFF
--- a/.evergreen/publish.sh
+++ b/.evergreen/publish.sh
@@ -17,7 +17,7 @@ export ORG_GRADLE_PROJECT_signingKey="${SIGNING_KEY}"
 export ORG_GRADLE_PROJECT_signingPassword=${SIGNING_PASSWORD}
 
 if [ "$RELEASE" == "true" ]; then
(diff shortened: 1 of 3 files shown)


vbabanin (dismissed):

vbabanin (approved):
LGTM!

PR #185, opened 2026-07-13
HIBERNATE-165 Support non-equijoin ON conditions

strogiyotec:
[HIBERNATE-165](https://jira.mongodb.org/browse/HIBERNATE-165) Support non-equijoin ON conditions

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
+++ b/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
@@ -57,6 +57,7 @@ import com.mongodb.hibernate.internal.service.StandardServiceRegistryScopedState
 import com.mongodb.hibernate.internal.translate.mongoast.AstArithmeticExpressionOperator;
 import com.mongodb.hibernate.internal.translate.mongoast.AstBinaryOperatorExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstComparisonExpressionOperator;
+import com.mongodb.hibernate.internal.translate.mongoast.AstComputedFieldUpdate;
 import com.mongodb.hibernate.internal.translate.mongoast.AstConversionExpressionOperator;
 import com.mongodb.hibernate.internal.translate.mongoast.AstDocument;
 import com.mongodb.hibernate.internal.translate.mongoast.AstElement;
@@ -76,7 +77,10 @@ import com.mongodb.hibernate.internal.translate.mongoast.AstValue;
 import com.mongodb.hibernate.internal.translate.mongoast.AstValueExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstVariableExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.command.AstDeleteCommand;
+import com.mongodb.hibernate.internal.translate.mongoast.command.AstDocumentUpdate;
 import com.mongodb.hibernate.internal.translate.mongoast.command.AstInsertCommand;
+import com.mongodb.hibernate.internal.translate.mongoast.command.AstPipelineUpdate;
+import com.mongodb.hibernate.internal.translate.mongoast.command.AstUpdate;
 import com.mongodb.hibernate.internal.translate.mongoast.command.AstUpdateCommand;
 import com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstAggregateCommand;
 import com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstLetVariable;
@@ -141,7 +145,6 @@ import org.hibernate.query.sqm.ComparisonOperator;
 import org.hibernate.query.sqm.function.SelfRenderingFunctionSqlAstExpression;
 import org.hibernate.query.sqm.sql.internal.BasicValuedPathInterpretation;
 import org.hibernate.query.sqm.sql.internal.SqmParameterInterpretation;
-import org.hibernate.query.sqm.sql.internal.SqmPathInterpretation;
 import org.hibernate.query.sqm.tree.expression.Conversion;
 import org.hibernate.sql.ast.Clause;
 import org.hibernate.sql.ast.SqlAstNodeRenderingMode;
@@ -172,7 +175,6 @@ import org.hibernate.sql.ast.tree.expression.Every;
 import org.hibernate.sql.ast.tree.expression.Expression;
 import org.hibernate.sql.ast.tree.expression.ExtractUnit;
 import org.hibernate.sql.ast.tree.expression.Format;
-import org.hibernate.sql.ast.tree.expression.FunctionExpression;
 import org.hibernate.sql.ast.tree.expression.JdbcLiteral;
 import org.hibernate.sql.ast.tree.expression.JdbcParameter;
 import org.hibernate.sql.ast.tree.expression.Literal;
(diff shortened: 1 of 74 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstLookupStageWithPipeline.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * The pipeline form of <a
+ * href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/lookup/">{@code $lookup}</a>, which binds
+ * outer-document values into {@code let} variables and evaluates a sub-{@code pipeline} against the joined collection.
+ * Used for non-equijoin {@code ON} conditions, which the simple {@code localField}/{@code foreignField} form cannot
+ * express.
+ *
+ * @hidden
+ */
+public record AstLookupStageWithPipeline(String from, List<AstElement> let, List<AstStage> pipeline, String as)
```
strogiyotec: this pipeline doesn't support localField and foreighField
as both are used for equality `join on c.id = o.id`
This pipeline is only used for ranges 
AstLookupStage supports equality already

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1491,25 +1473,59 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
+                    resolveFieldPath(columns.outer()),
+                    columns.joined().getColumnExpression(),
+                    joinAlias);
+        }
+
+        // The $expr array is always [<outer>, <joined>]; invert the operator when Hibernate placed the joined column
+        // on the left so the operand order stays outer-then-joined.
+        var exprOperator = createAstExprComparisonFilterOperator(columns.joinedOnLeft() ? operator.invert() : operator);
+
+        var letVariable = "v0";
```
strogiyotec: the variable name was created by claude, I believe it's fine if we see it in query logs but happy to adjust it

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1524,9 +1540,25 @@ private EquijoinFields extractEquijoinFields(@Nullable Predicate predicate, Stri
...
-        var outerCr = lhsIsJoined ? rhsCr : lhsCr;
-        var innerCr = lhsIsJoined ? lhsCr : rhsCr;
-        return new EquijoinFields(resolveFieldPath(outerCr), innerCr.getColumnExpression());
+        return lhsIsJoined ? new JoinColumns(rhsCr, lhsCr, true) : new JoinColumns(lhsCr, rhsCr, false);
+    }
+
+    /**
+     * Maps the non-equality comparison operators supported for non-equijoin {@code ON} conditions to their
+     * aggregation-expression counterparts. {@code EQUAL} never reaches here — it is routed to the simple
+     * {@code $lookup} form by the caller — and {@code DISTINCT_FROM}/{@code NOT_DISTINCT_FROM} are not yet supported.
```
strogiyotec, line 1945: I added test cases that assert failure on district queries

strogiyotec (commented):
comments

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1836,25 +1838,59 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
     }
 
-    private EquijoinFields extractEquijoinFields(@Nullable Predicate predicate, String joinedAlias) {
+    /**
+     * Builds the {@code $lookup} stage for a join {@code ON} condition. An {@code EQUAL} comparison maps to the simple
+     * {@code localField}/{@code foreignField} form; the ordering and inequality operators ({@code <}, {@code <=},
+     * {@code >}, {@code >=}, {@code !=}) require the pipeline form, which binds the outer column into a {@code let}
+     * variable and compares it against the joined column with {@code $expr}.
+     */
+    private AstStage buildJoinLookupStage(@Nullable Predicate predicate, String joinedCollection, String joinedAlias) {
```
strogiyotec, line 1876: this will not throw NPE exception when predicate is null

strogiyotec (commented):
comment

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1836,25 +1838,59 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
+                    resolveFieldPath(columns.outer()),
+                    columns.joined().getColumnExpression(),
+                    joinAlias);
+        }
+
+        // The $expr array is always [<outer>, <joined>]; invert the operator when Hibernate placed the joined column
+        // on the left so the operand order stays outer-then-joined.
+        var exprOperator = toJoinExprComparisonOperator(columns.joinedOnLeft() ? operator.invert() : operator);
+
+        var letVariable = "v0";
```
jyemin: `AstLookupStageWithPipeline` reuses `AstElement` for its `let` entries, but `AstElement` is a `name → AstValue` pair intended for literal documents (`AstDocument`, e.g. insert payloads).  A `let` binding, in contrast, is a `name → AstExpression` pair (more similar to a projection), not a document element. That type mismatch is the only reason `AstFieldPathValue` had to be kept after the merge: the field path gets re-expressed as an `AstValue` instead of using HIBERNATE-86's `AstFieldPathExpression`.

I suggest the following changes:

- Model a variable declaration in the AST, by adding an `AstLetVariable(String name, AstExpression expression)` node and change `AstLookupStageWithPipeline.let` to `List<AstLetVariable>`.
- Similarly, model a variable reference in the AST by adding an `AstVariableExpression(String name)` that renders `$$name`, replacing `new AstFieldPathExpression("$" + letVariable)`. 
- Delete `AstFieldPathValue`.

The translator can then pass a bare path/variable name and let the AST nodes own the `$`/`$$` prefixing, so the convention is consistent, and is no longer representing an expression as a value.

jyemin (changes requested):

jyemin (dismissed):
LGTM!

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: I think a derived name would be more future-proof and easier to debug. We could replace the hardcoded `"v0"` with a name derived from the outer column, e.g. `v{n}_{qualifier}_{column}` — `v0_c1_0__id`, `v1_o1_0_total` - where `n` is a per-query counter, e.g. `"v" + counter + "_" qualifier + "_" + outer.getColumnExpression().replaceAll("[^a-zA-Z0-9_]", "_")`.

The `v0` -> `v1` -> `v{n}` counter guarantees uniqueness, while the sanitized qualifier/column suffix is only there for human-readable context.  This approach is similar to MongoSql https://github.com/mongodb/mongosql/blob/main/mongosql/src/translator/utils.rs#L117


Two reasons:

1. **Scalability.** Once HIBERNATE-164 adds composite `ON` conditions, a single `$lookup` will need multiple `let` entries. The counter makes each name unique without collision detection.

2. **Debuggability.** `"v0"` in every `$lookup` makes logs harder to read. Including the qualifier and column name makes it clear which outer column is being bound.

vbabanin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Addressed it [here ](https://github.com/mongodb/mongo-hibernate/pull/185/changes/ecabfa3a1ebac4f62820c4bd7b2f1daa5688d449#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1885)
here is an [example ](https://github.com/mongodb/mongo-hibernate/pull/185/changes/ecabfa3a1ebac4f62820c4bd7b2f1daa5688d449#diff-b32874b2cf9858eb3dc583e1300a2dfe186f34f376144b34c73ee5491627559dR496)of the test case

jyemin (dismissed):
LGTM!

src/test/java/com/mongodb/hibernate/query/select/RegularExpressionH2ReferenceTests.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * The same {@code LIKE} scenarios as {@link com.mongodb.hibernate.query.select}'s MongoDB
+ * {@code RegularExpressionBugTests}, but run against an in-memory H2 database instead of MongoDB.
+ *
+ * <p>H2 implements the SQL standard {@code LIKE} operator, so it acts as a reference oracle: these tests are expected
+ * to PASS, which proves that the result sets asserted by the MongoDB bug tests are exactly the SQL-standard behavior,
+ * and that the current MongoDB {@code LIKE} → {@code $regex} translation (PR #171) diverges from it.
+ */
+class RegularExpressionH2ReferenceTests {
```
vbabanin: I assume this may have been committed accidentally? If yes, let's remove it. If not, let’s discuss.

vbabanin (changes requested):

vbabanin (approved):
I partially reviewed the `AbstractMqlTranslator` changes. I did not review the tests in and other changes.

LGTM!

PR #186, opened 2026-07-16
HIBERNATE-204 run integration tests in parallel with per-fork database isolation

jyemin:
Run this module's integration tests across parallel Gradle test forks instead of sequentially in a single JVM.

## Mechanism

- **Per-fork database isolation.** Each Gradle test fork gets its own database (base name suffixed with the Gradle worker id). `MongoExtension` derives and exposes that name; `TestServiceContributor` points every test `SessionFactory` at it (chosen over `MongoServiceRegistryProducer` because it is the `ServiceContributor` that also reaches tests built via `new Configuration()`). Off-Gradle runs (IDE) fall back to the base name, so single-process behavior is unchanged.
- **Serial lane for global state.** `failCommand` fail points are a single mongod-global switch, so tests that toggle them (`ExceptionHandlingIntegrationTests`, `NativeBootstrappingIntegrationTests`) are `@Tag("serial")` and run in a dedicated single-fork `integrationTestSerial` task that never overlaps the parallel task.
- **Fork count** defaults to `cpus/2`, overridable per machine via `-PitForks=<n>`.
- **Scope.** Parallelism is configured in the root module only; the Spring Boot modules keep single-fork execution (their tests are not fork-isolated).

## Result

The integration suite runs roughly 2-3x faster locally (~54s → ~18-25s depending on fork count).

Note: the fastest fork count is machine-specific and does not track core count in a simple way; the cause is not established here.

HIBERNATE-204

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -72,6 +72,7 @@ functions:
         files:
           - src/build/test-results/test/TEST-*.xml
           - src/build/test-results/integrationTest/TEST-*.xml
+          - src/build/test-results/integrationTestSerial/TEST-*.xml
 
   cleanup:
     - command: subprocess.exec
--- a/src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
@@ -58,6 +58,25 @@ public final class MongoExtension implements BeforeAllCallback, BeforeEachCallba
 
     private static final State STATE = State.create();
 
+    /**
+     * The database used by the current test JVM. When the suite runs across parallel Gradle test forks, each fork gets
+     * its own database (suffixed with the Gradle worker id) so forks do not drop each other's data. The
+     * {@code SessionFactory} under test must be pointed at this same database; see
+     * {@link MongoServiceRegistryProducer}.
+     */
+    public static String databaseName() {
+        return STATE.mongoDatabase().getName();
+    }
+
+    /**
+     * Appends the Gradle test-worker id to {@code baseName} so parallel forks use isolated databases. Returns
+     * {@code baseName} unchanged when not running under Gradle (e.g. from an IDE), preserving single-process behavior.
+     */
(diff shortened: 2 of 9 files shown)


.evergreen/config.yml
```diff
@@ -72,6 +72,7 @@ functions:
         files:
           - src/build/test-results/test/TEST-*.xml
           - src/build/test-results/integrationTest/TEST-*.xml
+          - src/build/test-results/integrationTestSerial/TEST-*.xml
```
strogiyotec, line 75: should we update AGENTS.md ? 
```
Integration tests are split into two tasks: `integrationTest` (parallel, fork-isolated) and `integrationTestSerial` (single-fork, runs only `serial`-tagged tests). `integrationTest` finalizes `integrationTestSerial`, so `./gradlew integrationTest` runs the full suite — including when filtered with `--tests ClassName`.

**Tag any test that manipulates mongod-global state (e.g. the `failCommand` fail point) with `serial`** so it runs single-fork instead of flaking under parallel forks.
```
This statement was suggested by Claude when I asked if we should update agents.md ?

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -58,6 +58,25 @@ public final class MongoExtension implements BeforeAllCallback, BeforeEachCallba
 
     private static final State STATE = State.create();
 
+    /**
+     * The database used by the current test JVM. When the suite runs across parallel Gradle test forks, each fork gets
+     * its own database (suffixed with the Gradle worker id) so forks do not drop each other's data. The
+     * {@code SessionFactory} under test must be pointed at this same database; see
+     * {@link MongoServiceRegistryProducer}.
```
strogiyotec: I think the comment is wrong, this databaseName is not used by RegistryProducer
we should put this instead
> com.mongodb.hibernate.TestServiceContributor

 This is the only place I found that uses `databaseName`

strogiyotec (changes requested):

strogiyotec (approved):

PR #187, opened 2026-07-17
HIBERNATE-198: support expressions in HQL bulk UPDATE SET

jyemin:
[HIBERNATE-198](https://jira.mongodb.org/browse/HIBERNATE-198): support expressions in HQL bulk UPDATE SET


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -987,26 +989,38 @@ public void visitUpdateStatement(UpdateStatement updateStatement) {
...
+        for (var assignment : assignments) {
+            var fieldValue = acceptAndYieldExpression(assignment.getAssignedValue());
+            fieldUpdates.add(new AstComputedFieldUpdate(resolveAssignmentFieldPath(assignment), fieldValue));
+        }
+        return new AstPipelineUpdate(fieldUpdates);
+    }
+
+    private String resolveAssignmentFieldPath(Assignment assignment) {
+        var fieldReferences = assignment.getAssignable().getColumnReferences();
+        assertTrue(fieldReferences.size() == 1);
```
strogiyotec, line 1022: I couldn't hit this assertion error because `SqlTupples` are not supported
but once they are supported I would prefer to have a cleaner error message
The `assertTrue` from our assertions class does not support passing a message string

So instead could we do something like this ? 
```
 if (fieldReferences.size() != 1) {
                  "Updating multiple columns of field path [%s] in a single assignment is not supported; "
                          + "assign each component separately",
                  fieldReferences.get(0).getColumnExpression()));
      }
```

strogiyotec (dismissed):
Left a small comment on the error messaging , apart from it looks good, verfied the changes locally

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
```diff
@@ -573,13 +574,20 @@ private static WriteModel<BsonDocument> createUpdateModel(
...
+            if (updateModification instanceof BsonDocument updateDocument) {
+                return isMulti
+                        ? new UpdateManyModel<>(filter, updateDocument)
+                        : new UpdateOneModel<>(filter, updateDocument);
+            } else if (updateModification instanceof BsonArray updatePipelineArray) {
+                var updatePipeline =
+                        updatePipelineArray.stream().map(BsonValue::asDocument).toList();
+                return isMulti
+                        ? new UpdateManyModel<>(filter, updatePipeline)
+                        : new UpdateOneModel<>(filter, updatePipeline);
```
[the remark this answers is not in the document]
strogiyotec: invalid claim
Empty pipeline u: [] — MongoDB accepts this as a valid no-op ({n:1, nModified:0, ok:1}); it's the identity transform, not a deferred error
u: [1] — this already throws: BsonValue::asDocument raises a BsonInvalidOperationException (a BSONException)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: I think our practice, as documented in the translation skill, is to only add FeatureNotSupportedException for testable paths.  Since tuples are not support this is genuinely unreachable. WDYT?

strogiyotec (approved):

PR #189, opened 2026-07-20
HIBERNATE-204 run integration tests in parallel

jyemin:
Run integration tests concurrently within a single JVM via JUnit parallel execution (classes-concurrent, methods-same-thread). To avoid interaction between concurrently-running tests, each top-level test class is isolated to its own database and its own `CommandListener` instance.

Tests that manipulate mongod-global state, such as the `failCommand` fail point, are annotated `@Isolated` so JUnit does not run them concurrently.

On an M4 Max this drops the integration test time from ~60 sec to ~14 sec.


src/integrationTest/java/com/mongodb/hibernate/internal/jdbc/MongoPreparedStatementIntegrationTests.java
```diff
@@ -78,7 +79,13 @@ class MongoPreparedStatementIntegrationTests {
 
     @BeforeAll
     static void beforeAll() {
-        sessionFactory = new Configuration().buildSessionFactory();
+        var configuration = new Configuration();
+        configuration
+                .getProperties()
+                .put(
+                        MONGO_CONFIGURATION_CONTRIBUTOR_KEY,
+                        MongoExtension.configurationContributorForClass(MongoPreparedStatementIntegrationTests.class));
```
jyemin, line 87: All integration tests now require this, but most pick it up by implementing `MongoServiceRegistryProducer`.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
+     * {@linkplain #databaseNameFor database} and installs the {@link TestCommandListener}. Used by
+     * {@link MongoServiceRegistryProducer} and by tests that bootstrap Hibernate directly, so every test is isolated to
+     * its own database without any per-thread state.
+     */
+    public static MongoConfigurationContributor configurationContributorForClass(Class<?> testClass) {
+        var databaseName = databaseNameFor(testClass);
+        return configurator -> configurator
+                .databaseName(databaseName)
+                .applyToMongoClientSettings(
+                        builder -> builder.addCommandListener(STATE.commandListenerByDatabase.computeIfAbsent(
```
jyemin, line 69: This used to happen in the now-deleted `TestServiceContributor`, but now that we're creating a new `TestCommandListener` instance per `MongoClient` instance, it moves here so we have a place to track the instances.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -111,7 +200,21 @@ private void close() {
         }
     }
 
-    private static void disableFailPoint() {
-        STATE.adminDatabase().runCommand(DISABLE_FAIL_POINT_COMMAND);
+    private static final class TestCommandListener implements CommandListener {
```
jyemin: Moved here from a top-level class, now deleted.  No need for it to be public now that access needs to be routed through methods of `MongoExtension` anyway.

src/integrationTest/java/com/mongodb/hibernate/query/mutation/BatchUpdateIntegrationTests.java
```diff
@@ -46,11 +45,6 @@ class BatchUpdateIntegrationTests extends AbstractQueryIntegrationTests implemen
     @InjectMongoCollection(COLLECTION_NAME)
     private static MongoCollection<BsonDocument> collection;
 
-    @BeforeEach
```
jyemin, line 49: `MongoExtension` now does this for any class that uses it.

src/integrationTest/java/com/mongodb/hibernate/query/AbstractQueryIntegrationTests.java
```diff
@@ -68,29 +65,25 @@
...
 
-    private TestCommandListener testCommandListener;
-
     @Override
     public void injectSessionFactoryScope(SessionFactoryScope sessionFactoryScope) {
         this.sessionFactoryScope = sessionFactoryScope;
     }
 
-    @Override
-    public void injectServiceRegistryScope(ServiceRegistryScope serviceRegistryScope) {
```
jyemin, line 84: `MongoExtension` now handles this for us.

src/integrationTest/java/com/mongodb/hibernate/TestServiceContributor.java
```diff
@@ -1,35 +0,0 @@
...
- * limitations under the License.
- */
-
-package com.mongodb.hibernate;
-
-import com.mongodb.hibernate.cfg.spi.MongoConfigurationContributor;
-import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
-import org.hibernate.service.spi.ServiceContributor;
-
-public final class TestServiceContributor implements ServiceContributor {
```
jyemin, line 23: No longer useful now that `MongoExtension` is managing the command listeners.

src/integrationTest/java/com/mongodb/hibernate/query/AbstractQueryIntegrationTests.java
```diff
@@ -68,29 +65,25 @@
...
 
-    protected TestCommandListener getTestCommandListener() {
-        return testCommandListener;
+    protected List<BsonDocument> getCommands() {
+        return MongoExtension.getCommands(getClass());
+    }
+
+    protected void clearCommands() {
+        MongoExtension.clearCommands(getClass());
     }
```
vbabanin: I like the simplification in this PR where the listener is auto-cleared by the extension.

Could we inject `TestCommandListener` via `@InjectTestCommandListener` instead of using static lookups through `MongoExtension.getCommands(getClass())`?

The current approach threads the test class identity through static methods (`getCommands(getClass())` / `clearCommands(getClass())`) to find the listener in an internal map. I think the annotation-based approach would be cleaner for a few reasons:

1. **Consistency:** `MongoClient` / `MongoCollection` are already injected via annotations, and `MongoCollection` also follows the auto-cleanup pattern. Injecting `TestCommandListener` and letting the extension clear it automatically keeps this consistent with the existing test infrastructure. I like that part of the cleanup simplification proposed here.

2. **Less coupling to `MongoExtension` internals:** exposing `getCommands` / `clearCommands` is a new pattern in the codebase and makes tests depend on how the extension stores listeners. With injection, tests can use the listener directly via `commandListener.getCommands()` / `commandListener.clear()`.

The change should be small: extract `TestCommandListener` as a package-visible class, add `@InjectTestCommandListener`, and wire it in `beforeAll` alongside the existing annotation handling.

Suggested commit: vbabanin@ddc7a76

P.S. I like the simplification in this PR where the listener is auto-cleared by the extension.

src/integrationTest/java/com/mongodb/hibernate/SessionFactoryIntegrationTests.java
```diff
@@ -16,17 +16,21 @@
...
+import com.mongodb.hibernate.junit.MongoExtension;
 import java.util.Map;
 import org.hibernate.SessionFactory;
 import org.hibernate.boot.MetadataSources;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.service.spi.ServiceException;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 
+@ExtendWith(MongoExtension.class)
```
vbabanin: Do we need `MongoExtension` here? I don’t see this test class using `@InjectMongoClient` / `@InjectMongoCollection`, persisting data, or relying on database cleanup. It looks like it only verifies whether the `SessionFactory` builds or fails.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
+     * Clear the MongoDB commands initiated so far by this test class. Required only in special cases when the test
+     * initiates one or more commands that it does not want to assert on.
+     */
+    public static void clearCommands(Class<?> testClass) {
+        STATE.commandListenerByDatabase.get(databaseNameFor(testClass)).clear();
+    }
+
     /** Injects the {@linkplain InjectMongoClient client}, {@linkplain InjectMongoCollection collections}. */
     @Override
     public void beforeAll(ExtensionContext context) throws Exception {
```
vbabanin: There appears to be a latent race here if `@InjectMongoCollection` is ever used on an inherited static field.

`MongoExtension.beforeAll` uses `findAnnotatedFields`, which walks the class hierarchy. If a base class declares a static `@InjectMongoCollection` field and two concrete subclasses run concurrently under `classes-concurrent`, both subclasses would write to the same inherited static field. Whichever `beforeAll` runs last wins, so one class could persist into its own database but read through a collection handle pointing at another class’s database.

<details>
<summary>Minimal reproduction</summary>

```java
abstract class BaseRaceTests implements SessionFactoryScopeAware, MongoServiceRegistryProducer {

    @InjectMongoCollection("race_items")
    static MongoCollection<BsonDocument> collection; // shared static - race target

    ...
}

class RaceSubclassATests extends BaseRaceTests {
    @Test
    void insertAndVerifyViaCollection() {
        getSessionFactoryScope().inTransaction(session -> session.persist(new Item(1, "from_A")));

        // collection may point to B's database - finds nothing - fails
        assertThat(collection.find())
                .extracting(doc -> doc.getString("name").getValue())
                .contains("from_A");
    }
}

class RaceSubclassBTests extends BaseRaceTests {
    @Test
    void insertAndVerifyViaCollection() {
        getSessionFactoryScope().inTransaction(session -> session.persist(new Item(1, "from_B")));

        // collection may point to A's database - finds nothing - fails
        assertThat(collection.find())
                .extracting(doc -> doc.getString("name").getValue())
                .contains("from_B");
    }
}
```

</details>

This is not triggered today because no current base class declares a static `@InjectMongoCollection` field. But we do have base-class test patterns like `AbstractSelectionTemporalIntegrationTests`, so this could be introduced later by accident.

Could we consider forbidding static fields for database-scoped injected resources (`@InjectMongoCollection`, `@InjectTestCommandListener`) and injecting them through `TestInstancePostProcessor.postProcessTestInstance` instead? Since `@SessionFactory` forces `@TestInstance(PER_CLASS)`, instance fields still live for the full class lifetime, but avoid inherited-static-field clobbering.

`@InjectMongoClient` could remain static because the client is shared across all classes.

For example:

```java
@Override
public void postProcessTestInstance(Object testInstance, ExtensionContext context) throws Exception {
    for (var field : findAnnotatedFields(testInstance.getClass(), InjectMongoCollection.class)) {
        if (isStatic(field)) {
            throw new IllegalStateException(
                    "@InjectMongoCollection field must not be static: " + field
                            + " — use an instance field to avoid races under parallel execution");
        }
        field.setAccessible(true);
        field.set(testInstance, ...);
    }
}
```

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
- * href="https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution">sequentially</a>.
- *
- * <p>The {@linkplain MongoConfigurator#databaseName(String) database} is dropped
- * {@linkplain MongoExtension#beforeEach(ExtensionContext) before each} test and
- * {@linkplain MongoExtension#afterAll(ExtensionContext) after all} tests. The fail points are disabled
- * {@linkplain MongoExtension#beforeEach(ExtensionContext) before each} test and
- * {@linkplain MongoExtension#afterAll(ExtensionContext) after all} tests.
+ * Test classes run concurrently (JUnit parallel execution, classes-concurrent / methods-same-thread), so each top-level
+ * test class gets its own database; nested classes share their top-level class's database. ({@code SessionFactory}
+ * instances are not shared — the testing framework builds one per test method.) The database is dropped before each
```
vbabanin: `@SessionFactory` is annotated with `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`, so the session factory is created once per test class.

```suggestion
 * test class gets its own database; nested classes share their top-level class's database. {@code SessionFactory} is shared across all test methods in a test class —
 * {@code @SessionFactory} meta-annotates {@code @TestInstance(PER_CLASS)}, so one instance
 * is created per test class. The database is dropped before each
```

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
+ * test and after all tests of the class.
```
vbabanin: The database appears to be dropped only after all tests in the class; before each test, only the collections are cleaned up via `deleteMany`.

```suggestion
 * instances are not shared — the testing framework builds one per test method.) Collection documents are cleared before each test, and 
 * the database is dropped after all tests in the class.
```

src/integrationTest/java/com/mongodb/hibernate/query/AbstractQueryIntegrationTests.java
jyemin: This is great.  I took your commit and applied one small change on top: I extracted a `CommandHistory` interface to inject, instead of `TestCommandListener`, to keep the surface area smaller.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
jyemin: I took your changed and applied it.  Nice future-proofing.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
+                .databaseName(databaseName)
+                .applyToMongoClientSettings(
+                        builder -> builder.addCommandListener(STATE.commandListenerByDatabase.computeIfAbsent(
+                                databaseName, ignored -> new TestCommandListener())));
+    }
+
+    /** Gets the MongoDB commands initiated by the currently-executing test in this test class. */
+    public static List<BsonDocument> getCommands(Class<?> testClass) {
+        return STATE.commandListenerByDatabase.get(databaseNameFor(testClass)).getCommands();
+    }
```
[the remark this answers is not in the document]
jyemin: The injection of the command history indirectly addresses this.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -24,40 +24,64 @@
...
+
+    /**
+     * Clear the MongoDB commands initiated so far by this test class. Required only in special cases when the test
+     * initiates one or more commands that it does not want to assert on.
+     */
+    public static void clearCommands(Class<?> testClass) {
+        STATE.commandListenerByDatabase.get(databaseNameFor(testClass)).clear();
+    }
```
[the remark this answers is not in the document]
jyemin: The injection of the command history indirectly addresses this.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -111,7 +200,21 @@ private void close() {
...
+            commands.add(event.getCommand().clone());
+        }
+
+        private List<BsonDocument> getCommands() {
+            return List.copyOf(commands);
+        }
+
+        private void clear() {
+            commands.clear();
+        }
```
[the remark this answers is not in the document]
jyemin: It doesn't need to be thread safe anymore.

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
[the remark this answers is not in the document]
jyemin: Addressed.

src/integrationTest/java/com/mongodb/hibernate/SessionFactoryIntegrationTests.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/junit/MongoExtension.java
```diff
@@ -17,91 +17,199 @@
...
+                            databaseName, ignored -> new TestCommandListener()));
         }
     }
 
     @Override
     public void afterAll(ExtensionContext context) {
-        disableFailPoint();
-        STATE.mongoDatabase().drop();
+        currentDatabase(context).drop();
+    }
```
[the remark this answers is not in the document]
jyemin: This config prevents any funny business like that:

```
# Integration tests run concurrently within a single JVM: test classes run concurrently, but methods within a class
# stay on one thread, so each class's own database (assigned by MongoExtension) is created, used, and dropped
# sequentially. Fail-point and other mongod-global-state tests are @Isolated so they never run concurrently with
# anything.
junit.jupiter.execution.parallel.enabled=true
junit.jupiter.execution.parallel.mode.default=same_thread
junit.jupiter.execution.parallel.mode.classes.default=concurrent
junit.jupiter.execution.parallel.config.strategy=fixed
junit.jupiter.execution.parallel.config.fixed.parallelism=6
```

vbabanin (approved):
LGTM!

PR #190, opened 2026-07-20
HIBERNATE-66: Building index and unique constraint for Mongo collection

apmasell:


src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
```diff
@@ -0,0 +1,84 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate;
+
+import static com.mongodb.internal.operation.IndexHelper.generateIndexName;
```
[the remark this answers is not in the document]
strogiyotec: ~not relevant , this class is also internal~

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,84 @@
...
+import org.hibernate.tool.schema.spi.Exporter;
+
+public abstract class MongoIndexExporter<T extends Exportable> implements Exporter<T> {
+    private final boolean unique;
+
+    protected MongoIndexExporter(boolean unique) {
+        this.unique = unique;
+    }
+
+    protected abstract Table tableForExportable(T exportable);
```
strogiyotec: I really like the idea with `<T extends Exportable>` to support both index and unique index

But can we move implementations to the separate classes instead of anonymous classes ? 

My thinking process is that we don't have a unit test for `getSqlCreateStrings` as we do for records where we check the generated string command

Once we have two separate implementations in their own classes we can then add unit tests 

```
   @Test
      void testSingleKeyUnique() {
          var exporter = exporter(true, "books", Optional.of("uniq_isbn"), "", new BsonElement("isbn", new BsonInt32(1)));
          assertThat(exporter.getSqlCreateStrings(UNUSED, null, null))
                  .containsExactly(
                          "{\"createIndexes\": \"books\", \"indexes\": [{\"key\": {\"isbn\": 1}, \"name\": \"uniq_isbn\", \"unique\": true}]}");
      }

      @Test
      void testMultipleKeysNonUnique() {
          var exporter = exporter(
                  false,
                  "books",
                  Optional.of("idx_publisher_author"),
                  "",
                  new BsonElement("publisher", new BsonInt32(1)),
                  new BsonElement("author", new BsonInt32(1)));
          assertThat(exporter.getSqlCreateStrings(UNUSED, null, null))
                  .containsExactly(
                          "{\"createIndexes\": \"books\", \"indexes\": [{\"key\": {\"publisher\": 1, \"author\": 1}, \"name\": \"idx_publisher_author\",
  \"unique\": false}]}");
      }
```

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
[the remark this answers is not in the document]
strogiyotec: my bad, I though this method comes from hibernate, we should not use mongo  java driver internal classes
As a quick work around we can copy paste the generate index method to one of our own utility classes @apmasell

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,135 @@
...
+public class IndexIntegrationTests {
+
+    @Test
+    void testIndexCreated() {
+        final var registry = new StandardServiceRegistryBuilder()
+                .applySettings(Map.of(
+                        "hibernate.hbm2ddl.auto",
+                        "update",
+                        "jakarta.persistence.schema-generation.database.action",
+                        "drop-and-create",
```
strogiyotec: this test is passing because pipeline starts with a fresh collections everytime , 
we do not support drop (the drop silently doesn't drop the collection)
Let's discuss if we want to support drop within this ticket 

Also create doesn't do anything as for mongo collections are created automatically so I think it's better to come up with the list of database.action that we want to support and for everything else fail fast

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
```diff
@@ -245,7 +247,13 @@ public void clearWarnings() throws SQLException {
...
         checkClosed();
         closeLastOpenResultSet();
-        throw new SQLFeatureNotSupportedException("TODO-HIBERNATE-66 https://jira.mongodb.org/browse/HIBERNATE-66");
+        var command = AdminCommand.decode(
+                new JsonReader(mql), DecoderContext.builder().build());
+        try {
+            return command.execute(mongoDatabase);
+        } catch (RuntimeException exception) {
+            throw handleExecuteQueryOrUpdateException(exception);
+        }
```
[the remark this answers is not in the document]
strogiyotec: @apmasell  I will take a look if this claim is true but basically the suggestion is 
```

command.execute()...
return false
```

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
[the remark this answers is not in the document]
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -299,4 +327,62 @@ public void appendDatetimeFormat(SqlAppender appender, String format) {
...
+            protected Optional<String> indexNameForExportable(Index exportable) {
+                return Optional.ofNullable(exportable.getName());
+            }
+
+            @Override
+            protected Stream<BsonElement> keysForExportable(Index exportable) {
+                return exportable.getSelectables().stream()
+                        .filter(selectable -> !selectable.isFormula())
+                        .map(selectable -> new BsonElement(selectable.getText(), new BsonInt32(1)));
+            }
```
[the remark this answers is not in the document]
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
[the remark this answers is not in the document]
apmasell: Copied in to the exporter.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Per discussion, these aren't actually sent to the server, they are parsed by `AdminCommand`, so we don't need that kind of check. I've added a note.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Updated, though, per-discussion, `create` still issues drop commands.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -310,4 +327,69 @@ public void appendDatetimeFormat(SqlAppender appender, String format) {
...
+                return exportable.getTable();
+            }
+
+            @Override
+            protected Optional<String> indexNameForExportable(Index exportable) {
+                return Optional.ofNullable(exportable.getName());
+            }
+
+            @Override
+            protected Stream<BsonElement> keysForExportable(Index exportable) {
```
jyemin: This method needs to handle the direction of each column, e.g.

```
@Index(name = "idx_on_multi_cols", columnList = "publisher desc,author asc")
```

In MQL that's represented as `new BsonInt32(-1)` for each descending column.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -310,4 +327,69 @@ public void appendDatetimeFormat(SqlAppender appender, String format) {
     public SQLExceptionConversionDelegate buildSQLExceptionConversionDelegate() {
         return (sqlException, exceptionMessage, mql) -> new JDBCException(exceptionMessage, sqlException, mql);
     }
+
+    @Override
+    public Exporter<Table> getTableExporter() {
+        return new NoOpExporter<>();
```
jyemin: At minimum, this should throw `FeatureNotSupportedException("TODO HIBERNATE-NNN..."), for the drop strings, so the extension doesn't silently fail to drop the collection.  But I think we should just implement drop since it's simple.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,111 @@
...
+        final var result =
+                switch (name) {
+                    case "createIndexes" -> {
+                        final var collectionName = reader.readString();
+                        reader.readName("indexes");
+                        final var indexes =
+                                listOf(AdminCommand::decodeIndexModel).decode(reader, decoderContext);
+                        yield new CreateIndexesCommand(collectionName, indexes);
+                    }
+                    case "dropIndexes" -> {
```
jyemin: Just a note to delete this branch since its unreachable.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,111 @@
...
+            reader.readEndArray();
+            return results;
+        };
+    }
+
+    static IndexModel decodeIndexModel(BsonReader reader, DecoderContext decoderContext) {
+        reader.readStartDocument();
+        reader.readName("key");
+        var key = new BsonDocumentCodec().decode(reader, decoderContext);
+        var name = reader.readString("name");
```
jyemin: This implementation is fragile as it relies on the order of the fields, which is not supposed to be significant in JSON.  So it will fail on

```
key { unique: true, name="foo" }
```

It should be robust to different orders.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
...
+import org.hibernate.annotations.Formula;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.junit.jupiter.api.Test;
+
+public class IndexIntegrationTests {
+
+    @Test
+    void testIndexCreated() {
+        final var registry = new StandardServiceRegistryBuilder()
```
jyemin: Unfortunately this is not caught by the linter, but the prevailing style omits final on locals. Let's remove them in this commit.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -310,4 +327,69 @@ public void appendDatetimeFormat(SqlAppender appender, String format) {
...
+
+            @Override
+            protected String optionsForExportable(Index exportable) {
+                return exportable.getOptions();
+            }
+        };
+    }
+
+    @Override
+    public Exporter<UniqueKey> getUniqueKeyExporter() {
```
jyemin, line 384: This method also needs to handle the direction of each column, e.g.
 
```
@UniqueConstraint(name = "uniq_idx_on_multi_cols", columnList = "publisher desc,author asc", unique=true)
```

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
...
+                .applySettings(Map.of(
+                        "hibernate.hbm2ddl.auto",
```
jyemin: This setting should be removed.  It's being ignored because `jakarta.persistence.schema-generation.database.action` is also set, and if you comment out the jakarta one the test fails.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
jyemin: Let's DRY this up by moving the registry configuration into a `@BeforeEach`-annotated helper method.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,111 @@
...
+import org.bson.BsonNumber;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.hibernate.boot.Metadata;
+import org.hibernate.boot.model.relational.Exportable;
+import org.hibernate.boot.model.relational.SqlStringGenerationContext;
+import org.hibernate.mapping.Table;
+import org.hibernate.tool.schema.spi.Exporter;
+
+public abstract class MongoIndexExporter<T extends Exportable> implements Exporter<T> {
```
jyemin: Let's move this into the same package as MongoDialect (its only usage), and make it package-private.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
...
+                .contains(
+                        BsonDocument.parse("{ name: \"idx_on_single_col\", key: {publishYear: 1}}"),
+                        BsonDocument.parse("{ name: \"idx_on_multi_cols\", key: {publisher: 1, author: 1}}"),
+                        BsonDocument.parse("{ name: \"uniq_idx_on_single_col\", key: {isbn: 1}, unique: true}"),
+                        BsonDocument.parse(
+                                "{ name: \"uniq_idx_on_multi_cols\", key: {publisher: 1, title: 1}, unique: true}"));
+    }
+
+    @Test
+    void testForbiddenOptions() {
```
jyemin: Project convention is to put tests like this in an `@Nested Unsupported` nested class.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
...
+        final var registry = new StandardServiceRegistryBuilder()
+                .applySettings(Map.of(
+                        "hibernate.hbm2ddl.auto",
+                        "update",
+                        "jakarta.persistence.schema-generation.database.action",
+                        "create",
+                        "hibernate.hbm2ddl.halt_on_error",
+                        "true"))
+                .build();
+        assertThrows(FeatureNotSupportedException.class, () -> {
```
jyemin: Project convention is to also assert on the exception message.

src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
```diff
@@ -0,0 +1,111 @@
...
+    protected abstract Optional<String> indexNameForExportable(T exportable);
+
+    protected abstract Stream<BsonElement> keysForExportable(T exportable);
+
+    @Override
+    public final String[] getSqlCreateStrings(T exportable, Metadata metadata, SqlStringGenerationContext context) {
+        final var collectionName = tableForExportable(exportable).getName();
+        final var keys = new BsonDocument();
+        keysForExportable(exportable).forEach(e -> keys.put(e.getName(), e.getValue()));
+        final var indexName = indexNameForExportable(exportable).orElseGet(() -> generateIndexName(keys));
```
jyemin: There are no tests for the `orElseGet` branch, and when I tried it out I saw something surprising: Hibernate manufactures an index name if one is not supplied.  It's never null.  It generates a random looking index name (though not actually sure it's random).  

We should test with a UniqueConstraint/Index with no name to lock that in, in case Hibernate ever changes its behavior, and then we can delete the orElseGet` and `generateIndexName`.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
...
+import org.bson.BsonDocument;
+import org.hibernate.annotations.Formula;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.junit.jupiter.api.Test;
+
+public class IndexIntegrationTests {
+
+    @Test
+    void testIndexCreated() {
```
jyemin: Similar to our existing mutation tests, this test should also assert MongoDB state, in addition to the command assertions.  The pattern is 

```
public class IndexIntegrationTests
```

then

```
    private static MongoCollection<BsonDocument> booksCollection;
```

then assert on the return of `MongoCollection#listIndexes`.

This has the additional nice property that the database is dropped before each test, so it's starting with a clean slate each run. (I ran into this problem when I changed an index name, and the createIndexes command failed because there was already an existing index with the same keys, which is an error in MongoDB, and I had to drop the collection in the shell to work around it.)

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Moved.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Implemented.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Deleted.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Removed.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Removed.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Refactored.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Moved.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
apmasell: I changed a test. Let me know if that's what you intended.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Added.

jyemin:
I'm still reviewing the PR, but just want to note that CI is failing due to the HIBERNATE-204 commit that requires corresponding changes to the new integration test.

There are three things to do here, and each only surfaces once the previous is fixed:

1. TestCommandListener → @InjectCommandHistory CommandHistory, with commandHistory.getCommands() in place of registry.requireService(...).getStartedCommands().
2. @InjectMongoCollection fields must now be instance, not static — otherwise IllegalStateException: … must not be static — use an instance field to avoid races under parallel execution.
3. inRegistry builds its registry by hand, so it has to apply the per-class contributor itself:

```
  .applySetting(MONGO_CONFIGURATION_CONTRIBUTOR_KEY,
                MongoExtension.configurationContributorForClass(IndexIntegrationTests.class))
```

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
jyemin: Moved, thanks. The visibility half is still open: `public abstract class MongoIndexExporter` at :40 and `public static String generateIndexName` at :55. Both can be package-private now that MongoDialect in the same package is the only caller, and per the other thread generateIndexName can go away entirely.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
jyemin: Still not fixed. `IndexBinder.initializeColumns` strips a trailing ` asc` or ` desc` off each columnList entry before the Selectable is built, so `selectable.getText()` never contains a direction and the regex in `generateKeys` has nothing to match. The ordering is kept separately, in `Index.getSelectableOrderMap()` (Index.java:71), with values "asc", "desc", or null.

The example from my original comment still comes out ascending:

    @Index(name = "idx_on_multi_cols", columnList = "publisher desc,author asc")
      -> {"key": {"publisher": 1, "author": 1}}

I have written this up as a failing test rather than more prose: [`descendingIndexes`](https://github.com/jyemin/mongo-hibernate/blob/8f030d4878c6ef5fc071552b1648d9e7c7c1303e/src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java#L239-L253). Reading the order map also lets the regex go away, which fixes the space-in-a-field-name failure at the same time.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
jyemin: Also still not fixed, and the ordering is likewise available to read. On this branch `@Index(columnList = "isbn desc", unique = true)` binds to a UniqueKey with a populated order map, and still emits ascending:

    columns=['isbn'], ordering recorded: desc
    emitted: {"key": {"isbn": 1}, "name": "uk_b", "unique": true}

Source is `UniqueKey.getColumnOrderMap()` (UniqueKey.java:40).

One correction to my original comment: the syntax I wrote is not valid. `@UniqueConstraint` has no columnList and no direction grammar at all; only `@Index.columnList` has the BNF `index_column ::= column_name [ASC | DESC]`. Direction reaches a UniqueKey solely via `@Index(columnList = "... desc", unique = true)`, which IndexBinder.java:174 routes to getOrCreateUniqueKey.

Covered by [`uniqueDescendingIndexes`](https://github.com/jyemin/mongo-hibernate/blob/8f030d4878c6ef5fc071552b1648d9e7c7c1303e/src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java#L280-L298), which fails on this branch.

src/main/java/com/mongodb/hibernate/internal/translate/MongoIndexExporter.java
jyemin: Yes, that is the test I wanted, and it answers the question: the unnamed `@Index(columnList = "publisher,title", unique = true)` comes out as IDXbeyw7jm8ev66e1mbr0hggy13e, so Hibernate always manufactures a name.

That makes the second half actionable. `indexNameForExportable` is never empty, so the orElseGet at :97 and generateIndexName at :55 are unreachable rather than merely untested. Both can be deleted and the call collapsed to the name. Your test is what proves it, so deleting them loses no coverage.

Kept pinned as [`unnamedDeclarationsAreNamedByHibernate`](https://github.com/jyemin/mongo-hibernate/blob/8f030d4878c6ef5fc071552b1648d9e7c7c1303e/src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java#L321-L329), over its own entity so the asserted hash does not move when unrelated columns change. That one passes already.

jyemin (changes requested):
I was having trouble writing my findings in prose, so instead I decided to put what I understand the requirements to be into the integration test itself and pushed it to a branch on my fork. If we can first agree on the test surface, then the implementation will follow naturally from there:

https://github.com/jyemin/mongo-hibernate/blob/HIBERNATE-66-new-tests/src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java

There are thirteen tests in the updated class, of which six are failing.

Failing:

| Test | What is wrong |
|---|---|
| `descendingIndexes` | direction ignored on `@Index` |
| `uniqueDescendingIndexes` | direction ignored on `@Index(unique = true)` |
| `columnLevelUnique` | `@Column(unique = true)` emits no index at all |
| `indexesFollowTheQualifiedCollectionName` | indexes created on the unqualified collection name, so they land on a different collection from the data |
| `InvalidMappings.indexOnUnmappedColumn` | a `columnList` entry naming an unmapped column is not rejected |
| `InvalidMappings.uniqueConstraintOnUnmappedColumn` | the same through `columnNames` |

Passing, so that what already works is guarded rather than bundled together with what does not: `ascendingIndexes`, `uniqueConstraints`, `unnamedDeclarationsAreNamedByHibernate`, `createDropLifecycle`, and the three cases in `Unsupported`.

Three things about the shape of the tests, since they differ from what is on the branch now.

Each test asserts both the emitted `createIndexes` command and the resulting index set on the server, through `listIndexes`. "The server accepted it" and "the server built what the mapping asked for" are different claims, and the second is more interesting, but the command is what our code actually produces and some of it is invisible in server state: one `createIndexes` per index and several batched into one command look identical afterwards. A `createIndexes(collection, name, unique, key...)` helper builds the expected command, so asserting both costs one line per test.

Comparisons are order-preserving. Commands are compared as JSON and each key document is flattened to an ordered list of `field:direction` pairs, because `BsonDocument` implements `Map`, so comparing documents directly ignores field order, and on a compound index the field order decides which sorts and prefix queries the index can serve. On the current branch you can swap the expectation for `idx_on_multi_cols` to `{author: 1, publisher: 1}` and it still passes.

Each group of declarations gets its own entity and collection. Partly so one defect cannot hide the state of everything else, and partly because MongoDB rejects a second index with the same key pattern under a different name, so declarations that overlap on fields cannot share a collection. That is also why `@Column(unique = true)` needs a field no other uniqueness declaration touches. A `@BeforeEach` drops those collections, since `MongoExtension` empties collections between tests but leaves their indexes standing, and the drop half of `create-drop` does not run when a `SessionFactory` fails to open, which is what the negative tests provoke.

I also deleted `testGarbageDirection`. An invalid direction is not a reachable input: `IndexBinder.initializeColumns` recognizes only a trailing ` asc` or ` desc` and leaves anything else as part of the column name, so that test was really exercising the column-name regex and would fail as soon as the regex goes away. What it was standing in front of is a real gap, which is now `InvalidMappings`.

Nine of my earlier threads are resolved. Four are still open, with replies on each.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
jyemin: Now the public modifier can be removed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,113 @@
...
+                        new BsonArray(List.of(new BsonDocument(List.of(
+                                new BsonElement("key", keys),
+                                new BsonElement("name", new BsonString(indexName)),
+                                new BsonElement("unique", BsonBoolean.valueOf(unique)))))))));
+        // This intentionally looks like a Mongo command, but it is parsed by AdminCommand and is not sent directly to
+        // the server
+        return command.toJson(MongoConstants.EXTENDED_JSON_WRITER_SETTINGS);
+    }
+
+    static String nameForTable(Table table) {
```
jyemin: I found one more gap that the new integration tests miss.

This method prepends the schema from the table definition, so it doesn't account for `hibernate.default_schema`: `table.getSchema()` is null when no `@Table(schema = ...)` is present, but Hibernate applies the configured default when it renders the table name for data operations. So DDL and data disagree.

With `hibernate.default_schema = lib` and `@Table(name = "t", indexes = @Index(...))`:

    create        -> t
    createIndexes -> t
    insert        -> lib.t

The documents go to `lib.t`, while the indexes are created on `t`.

`SqlStringGenerationContext` is already a parameter of both exporters and is the provided mechanism for applying the default schema:

    var collectionName = context.format(table.getQualifiedTableName());

This will need a test case.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
```diff
@@ -0,0 +1,583 @@
...
+ * implements {@link Map}, so comparing key documents with {@code equals} would silently ignore order.
+ * {@link #indexesOf} flattens each key document to an ordered list to avoid that.
+ *
+ * <p>Each group of declarations gets its own entity and collection, one per test, for two reasons. Every test asserts
+ * the complete index set for its collection, so bundling unrelated declarations would let one defect hide the state of
+ * everything else. And MongoDB rejects a second index with the same key pattern under a different name
+ * ({@code IndexOptionsConflict}), so declarations that overlap on fields cannot share a collection.
+ */
+@ExtendWith(MongoExtension.class)
+class IndexIntegrationTests {
```
jyemin, line 72: I discovered that the unique index exporter allows us to support the `NaturalId` annotation.  Can you add a test that an entity like

```
    @Entity(name = "NaturalIdEntity")
    @Table(name = "natural_id")
    static class NaturalIdEntity {
        @Id
        int id;

        @NaturalId
        String isbn;
    }
```

gets a unique index on `isbn`, so that we lock in the support with a regression test?

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -310,4 +322,98 @@ public void appendDatetimeFormat(SqlAppender appender, String format) {
...
     }
+
+    @Override
+    public Exporter<Table> getTableExporter() {
+        return new Exporter<>() {
+            @Override
+            public String[] getSqlCreateStrings(
+                    Table exportable, Metadata metadata, SqlStringGenerationContext context) {
+                var collectionName = MongoIndexExporter.nameForTable(exportable);
+                return Stream.concat(
```
jyemin: It's a bit jarring to see index creation in the table exporter.  

Since we already have a way to create a unique key via the unique key exporter, we can re-use that code path for unique columns by adding each one as a `UniqueKey` at boot (in `MongoAdditionalMappingContributor`), using the method `org.hibernate.mapping.Table#getOrCreateUniqueKey`.

I prototyped it here if you want to take a look: https://github.com/jyemin/mongo-hibernate/compare/3abb909a...jyemin:mongo-hibernate:HIBERNATE-66-unique-column-experiment

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import org.bson.BsonDocument;
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.codecs.Decoder;
+import org.bson.codecs.DecoderContext;
+
+public abstract sealed class AdminCommand
```
jyemin: Let's make this class package private.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final Decoder<List<IndexModel>> INDEX_LIST =
+            listOf(MongoClientSettings.getDefaultCodecRegistry().get(Index.class), Index::intoIndexModel);
+
+    public record Index(String name, BsonDocument key, boolean unique) {
+        IndexModel intoIndexModel() {
+            return new IndexModel(key, new IndexOptions().name(name).unique(unique));
+        }
+    }
+
+    public static AdminCommand decode(BsonReader reader, DecoderContext decoderContext)
```
jyemin: Let's make this method package private

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.function.Function;
+import org.bson.BsonDocument;
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.codecs.Decoder;
+import org.bson.codecs.DecoderContext;
+
+public abstract sealed class AdminCommand
+        permits AdminCommand.CreateIndexesCommand, AdminCommand.CreateCollection, AdminCommand.DropCollection {
+    private static <T, R> Decoder<List<R>> listOf(Decoder<T> inner, Function<T, R> mapper) {
```
jyemin: There's no need for this method to be generic, as it's used in only one context.  It can simplify to

```
    private static final Decoder<Index> INDEX_DECODER =
            MongoClientSettings.getDefaultCodecRegistry().get(Index.class);

    private static List<IndexModel> decodeIndexes(BsonReader reader, DecoderContext decoderContext) {
        var indexes = new ArrayList<IndexModel>();
        reader.readStartArray();
        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
            indexes.add(INDEX_DECODER.decode(reader, decoderContext).intoIndexModel());
        }
        reader.readEndArray();
        return indexes;
    }
```

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
jyemin: Let's simplify this method signature to just take the mql as a String, rather than a BsonReader and a DecoderContext.  A JsonReader is Closeable, so it forces the caller to take responsibility for calling close, and the DecoderContext is not used, so there is no reason for a caller to supply one.  Since this method ir responsible for the full decoding lifecycle, it should create the JsonReader and close it.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,113 @@
...
+    protected abstract String optionsForExportable(T exportable);
+
+    @Override
+    public final String[] getSqlDropStrings(T exportable, Metadata metadata, SqlStringGenerationContext context) {
+        throw new IllegalStateException(
+                "HIBERNATE-66: Dropping indices was deemed something that Hibernate never called");
+    }
+
+    protected record IndexEntry(String name, String direction) {
+        void insert(BsonDocument document, Table table) {
```
jyemin: Let's split this method into two parts.

1. Hoist the column name check into the loop on line 79.
2. Move the conversion of the direction string to the integer to a private static helper method in the outer class, so it's a sibling of the createIndex method above.

Then this record becomes just a DTO that abstracts the difference between UniqueKey and Index in the Hibernate API.  This is more consistent with our style in other parts of the project, like the translator.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,113 @@
...
+
+public abstract class MongoIndexExporter<T extends Exportable> implements Exporter<T> {
+
+    private final boolean unique;
+
+    protected MongoIndexExporter(boolean unique) {
+        this.unique = unique;
+    }
+
+    static String createIndex(String collectionName, BsonValue keys, String indexName, boolean unique) {
```
jyemin: The name of this method makes it sound like it's actually creating the index. Let's rename to something like toCreateIndexesCommand.  That's closer to our project conventions, e,g, toFilter in the translator.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+            reader.readEndArray();
+            return results;
+        };
+    }
+
+    private static final Decoder<List<IndexModel>> INDEX_LIST =
+            listOf(MongoClientSettings.getDefaultCodecRegistry().get(Index.class), Index::intoIndexModel);
+
+    public record Index(String name, BsonDocument key, boolean unique) {
+        IndexModel intoIndexModel() {
```
jyemin: Let's name this method `toIndexModel` to match project conventions.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
```diff
@@ -0,0 +1,113 @@
...
+
+    static String nameForTable(Table table) {
+        if (table.getSchema() == null || table.getSchema().isBlank()) {
+            return table.getName();
+        } else {
+            return table.getSchema() + "." + table.getName();
+        }
+    }
+
+    protected abstract String indexNameForExportable(T exportable);
```
jyemin: Now that everything is in the same package, let's remove all the protected modifiers from this class.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
jyemin: Let's make the type of the keys parameter BsonDocument, to make the intent clearer.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+                        throw new SQLFeatureNotSupportedException(
+                                "Cannot decode command %s: unknown command".formatted(name));
+                };
+        reader.readEndDocument();
+        return result;
+    }
+
+    abstract void execute(MongoDatabase database);
+
+    static final class CreateCollection extends AdminCommand {
```
jyemin: Let's rename to `CreateCollectionCommand`, to match `CreateIndexesCommand` and `AdminCommand`.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
```diff
@@ -0,0 +1,121 @@
...
+            this.collectionName = collectionName;
+        }
+
+        @Override
+        void execute(MongoDatabase database) {
+            database.createCollection(collectionName);
+        }
+    }
+
+    static final class DropCollection extends AdminCommand {
```
jyemin: Let's rename to `DropCollectionCommand`, to match `CreateIndexesCommand` and `AdminCommand`.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
jyemin: Let's rename this method to `toAdminCommand` to match project conventions.

jyemin (changes requested):
Correctness guarantees look good now.  I found just one issue with default schema that needs a behavioral change.  

This review focuses more on design and project conventions

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Fixed.

src/integrationTest/java/com/mongodb/hibernate/query/IndexIntegrationTests.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: I assumed this was going to be necessary for future admin commds.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoIndexExporter.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/jdbc/AdminCommand.java
jyemin: As far as I can tell these are the only admin commands we will need to support.  And even if there are new ones they are not likely to have exactly the structure required by this method.

jyemin (changes requested):
A few of the comment threads have not been addressed yet, so sending this back.

jyemin (approved):
LGTM

strogiyotec (approved):

PR #192, opened 2026-07-23
HIBERNATE-164 compound join

strogiyotec:
Join `ON` conditions are now translated by delegating to the shared `EXPRESSION`-mode predicate
visitors — the same code that translates a `WHERE` predicate.

### Now supported 
Because ON delegates to the WHERE visitors, essentially any predicate WHERE supports now works in ON:


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,44 +1868,73 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-                new AstFieldPathExpression(columns.joined().getColumnExpression())));
+        var letVariables = new ArrayList<AstLetVariable>();
+        var expr = buildJoinFilterExpression(predicate, joinedAlias, letVariables);
         return new AstLookupStageWithPipeline(
-                joinedCollection,
-                List.of(new AstLetVariable(letVariable, new AstFieldPathExpression(resolveFieldPath(columns.outer())))),
-                List.of(new AstMatchStage(expr)),
-                joinAlias);
+                joinedCollection, letVariables, List.of(new AstMatchStage(new AstExprFilter(expr))), joinAlias);
+    }
```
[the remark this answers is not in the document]
strogiyotec: Addressed it [here](https://github.com/mongodb/mongo-hibernate/pull/192/changes/12c45d837753d89f7b6a93d42f27a93ae28fd100#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1882)

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -120,11 +136,15 @@ static class LineItem {
 
     @BeforeEach
     void beforeEach() {
-        var customers = List.of(new Customer(1, "Alice"), new Customer(2, "Bob"), new Customer(3, "Charlie"));
+        var customers =
+                List.of(new Customer(1, "Alice", 10), new Customer(2, "Bob", 20), new Customer(3, "Charlie", 30));
+        // Order 3 belongs to customer 2 (customerId=2) but carries a non-matching region (99), so a
+        // compound ON of "c.id = o.customerId AND c.region = o.region" selects only orders 1 and 2 —
+        // proving the second conjunct actually filters (HIBERNATE-164).
```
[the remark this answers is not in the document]
strogiyotec: addressed it [here](https://github.com/mongodb/mongo-hibernate/pull/192/changes/12c45d837753d89f7b6a93d42f27a93ae28fd100#diff-b32874b2cf9858eb3dc583e1300a2dfe186f34f376144b34c73ee5491627559dR141)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,44 +1868,83 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
+            @Nullable Predicate predicate, String joinedAlias, List<AstLetVariable> letVariables) {
+        if (predicate instanceof ComparisonPredicate cp) {
+            var columns = extractJoinColumns(cp, joinedAlias);
+            // The $expr array is always [<outer>, <joined>]; invert the operator when Hibernate placed the joined
+            // column on the left so the operand order stays outer-then-joined.
+            var exprOperator = toJoinExprComparisonOperator(
+                    columns.joinedOnLeft() ? cp.getOperator().invert() : cp.getOperator());
+            var letVariable = nextLetVariableName(columns.outer());
+            letVariables.add(
+                    new AstLetVariable(letVariable, new AstFieldPathExpression(resolveFieldPath(columns.outer()))));
```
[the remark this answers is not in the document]
jyemin: @strogiyotec this seems reasonable on the face of it.  WDYT?

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -876,235 +901,1140 @@ void testChainedNonEquijoin() {
     }
 
     @Nested
-    @DomainModel(annotatedClasses = {ManyToManyJoin.ItemA.class, ManyToManyJoin.ItemB.class})
-    class ManyToManyJoin extends AbstractQueryIntegrationTests {
```
strogiyotec, line 880: this class was moved below

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,93 +1884,103 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-     * Builds the {@code $lookup} stage for a join {@code ON} condition. An {@code EQUAL} comparison maps to the simple
-     * {@code localField}/{@code foreignField} form; the ordering and inequality operators ({@code <}, {@code <=},
-     * {@code >}, {@code >=}, {@code !=}) require the pipeline form, which binds the outer column into a {@code let}
-     * variable and compares it against the joined column with {@code $expr}.
+     * Builds the {@code $lookup} stage for a join {@code ON} condition.
+     *
+     * <p>A lone one-outer-one-joined equijoin uses the compact, index-friendly {@code localField}/{@code foreignField}
+     * form. Every other shape uses the pipeline form and is translated by delegating to the shared EXPRESSION-mode
+     * predicate visitors — the same ones that translate a {@code WHERE} predicate — so a given HQL predicate produces
+     * the same {@code $expr} whether it appears in {@code WHERE} or {@code ON}. The {@link #joinLookupContext} makes
```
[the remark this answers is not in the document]
strogiyotec: addressed it [here](https://github.com/mongodb/mongo-hibernate/pull/192/changes/a7debfeea88aab29896d50a6fbc9d2374a508cb1#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270df)

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -876,235 +901,1140 @@ void testChainedNonEquijoin() {
...
+                    }""",
+                    results -> assertIterableEq(
+                            List.of(new Object[] {1, 100}, new Object[] {2, 200}, new Object[] {3, null}), results),
+                    Set.of("Customer", "Order"));
+        }
+
+        @Test
+        void testNotEqualConjunct() {
+            // A <> conjunct maps to $ne alongside the $eq.
+            // but NOT (a = b) is not supported yet
```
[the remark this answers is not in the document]
strogiyotec: addressed it [here ](https://github.com/mongodb/mongo-hibernate/pull/192/changes/a7debfeea88aab29896d50a6fbc9d2374a508cb1#diff-b32874b2cf9858eb3dc583e1300a2dfe186f34f376144b34c73ee5491627559d)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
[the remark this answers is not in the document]
strogiyotec: I strongly believe the number of variables used in compound JOINS will always be small enough 
and duplicates only occur when the same outer column is referenced more than once, which is rare 
I think adding the state (like a HashMap) will add more complexity 

I don't have a strong opinion against it so @jyemin  if you think it's worth it I can address it in this PR

src/integrationTest/java/com/mongodb/hibernate/query/select/JoinSelectQueryIntegrationTests.java
```diff
@@ -1419,12 +2297,13 @@ static class ItemWithSecondaryTable {
...
             void testJoinFormulaThrows() {
+                // A @JoinFormula column is rejected wherever it is referenced (visitColumnReference), the same as
+                // a formula column in any other clause.
                 assertThat(assertThrows(RuntimeException.class, () -> getSessionFactoryScope()
                                 .getSessionFactory()))
                         .rootCause()
                         .isInstanceOf(FeatureNotSupportedException.class)
-                        .hasMessage(
-                                "TODO-HIBERNATE-182 https://jira.mongodb.org/browse/HIBERNATE-182 @JoinFormula is not supported");
+                        .hasMessage("Formula is not supported");
```
strogiyotec, line 2305: now the flow goes through `visitColumnReference`

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,93 +1884,103 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-                new AstVariableExpression(letVariable),
-                new AstFieldPathExpression(columns.joined().getColumnExpression())));
-        return new AstLookupStageWithPipeline(
-                joinedCollection,
-                List.of(new AstLetVariable(letVariable, new AstFieldPathExpression(resolveFieldPath(columns.outer())))),
-                List.of(new AstMatchStage(expr)),
-                joinAlias);
+        var letVariables = new ArrayList<AstLetVariable>();
+        var previousContext = joinLookupContext;
+        joinLookupContext = new JoinLookupContext(joinedAlias, letVariables);
```
strogiyotec: right now the previousContext is always null but won't be null once HIBERNATE-168 is implemented (according to Claude , every test I debugged always produced a null context)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,93 +1884,105 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-        }
-        if (!(predicate instanceof ComparisonPredicate cp)) {
-            throw new FeatureNotSupportedException("TODO-HIBERNATE-200 https://jira.mongodb.org/browse/HIBERNATE-200");
-        }
-        var operator = cp.getOperator();
-        var columns = extractJoinColumns(cp, joinedAlias);
         var joinAlias = JOIN_ALIAS_PREFIX + joinedAlias;
 
-        if (operator == ComparisonOperator.EQUAL) {
+        // Parentheses are semantically inert, so unwrap them before testing for the compact-form fast path.
```
jyemin: I don't understand this comment.  How would one go about following its instruction?

Probably ok to remove.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,93 +1884,105 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-                exprOperator,
-                new AstVariableExpression(letVariable),
-                new AstFieldPathExpression(columns.joined().getColumnExpression())));
-        return new AstLookupStageWithPipeline(
-                joinedCollection,
-                List.of(new AstLetVariable(letVariable, new AstFieldPathExpression(resolveFieldPath(columns.outer())))),
-                List.of(new AstMatchStage(expr)),
-                joinAlias);
+        var letVariables = new ArrayList<AstLetVariable>();
+        var previousContext = joinLookupContext;
```
jyemin: I saw your comment.  

Consider making this `assertNotNull(joinLookupContext)` instead.  Once we get to sub-queries, there will be a lot more to do than just this to make everything work, and we'll likely have to rework this code anyway. For example, `letVariableCounter` should also be saved/restored, along with most of the rest of the translator fields.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Started a new comment thread about this.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: the comment was about the line below
`        var simpleEquijoin = trySimpleEquijoinColumns(unwrapGrouped(predicate), joinedAlias);`
specifically the unwrap function call , for the HQL
```
SELECT c.id, o.total FROM Customer c JOIN Order o ON (c.id = o.id)
```
the parentheses around equijoin are semantically the same as equijoin without parentheses(hence the comment "are semantically inert") , without the unwrap it will create the pipeline aggregation
With unwrap the generated MQL is
```
{ "$lookup": { "from": "Order", "localField": "_id", "foreignField": "_id", "as": "#o1_0" } }
```

How about changing the comment to 
>         // unwrapGrouped call lets a parenthesized equijoin ON (a = b) use the compact form instead of pipeline.

strogiyotec: @jyemin  I [updated ](https://github.com/mongodb/mongo-hibernate/pull/192/changes/ad84372d606641ec9130be3c89b9c59eab31df2e#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1901)the comment itself, let me know if you agree

jyemin: oh... I completely misunderstood it! I saw the word "testing" and I thought of integration testing, not a translator test

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: `assertNotNull` doesn't accept the error message , I added a manual `fail`

strogiyotec: https://github.com/mongodb/mongo-hibernate/pull/192/changes/ad84372d606641ec9130be3c89b9c59eab31df2e#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1913

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1910,14 +1910,16 @@ private AstStage buildJoinLookupStage(@Nullable Predicate predicate, String join
...
+            throw fail("Nested join ON conditions are not supported");
+        }
         joinLookupContext = new JoinLookupContext(joinedAlias, letVariables);
         try {
             var expr = acceptAndYieldExpression(assertNotNull(predicate));
             return new AstLookupStageWithPipeline(
                     joinedCollection, letVariables, List.of(new AstMatchStage(new AstExprFilter(expr))), joinAlias);
         } finally {
-            joinLookupContext = previousContext;
+            joinLookupContext = null;
```
jyemin: With the above throw, this is no longer necessary so the `finally` can be removed, and therefore the `try` as well.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1910,14 +1910,16 @@ private AstStage buildJoinLookupStage(@Nullable Predicate predicate, String join
         }
 
         var letVariables = new ArrayList<AstLetVariable>();
-        var previousContext = joinLookupContext;
+        if (joinLookupContext != null) {
```
jyemin, line 1912: I'm really not sure when you should use this pattern vs `assertNull`.  I guess if you want to include a failure message?  Anyway, not sure if you considered the alternative, but feel free to resolve if you have.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1868,93 +1884,107 @@ private List<AstStage> buildJoinStages(TableGroup tableGroup) {
...
-        var expr = new AstExprFilter(new AstBinaryOperatorExpression(
-                exprOperator,
-                new AstVariableExpression(letVariable),
-                new AstFieldPathExpression(columns.joined().getColumnExpression())));
-        return new AstLookupStageWithPipeline(
-                joinedCollection,
-                List.of(new AstLetVariable(letVariable, new AstFieldPathExpression(resolveFieldPath(columns.outer())))),
-                List.of(new AstMatchStage(expr)),
-                joinAlias);
+        var letVariables = new ArrayList<AstLetVariable>();
```
jyemin: Nit: this would be a bit cleaner if JoinLookupContext creates the list (in a secondary constructor), and we just access it on line 1920 from there instead of a local variable.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I didn't find other utility methods that accept an assertion along with the message , for this I strongly want to have a meaningful error message

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Great idea thanks 
Moved the init into Constructor [here](https://github.com/mongodb/mongo-hibernate/pull/192/changes/7bfe77c436fed72aa13a08afb78515bd2a26a62d#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1934)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Can you also address this comment?

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I removed the try finally block [here](https://github.com/mongodb/mongo-hibernate/pull/192/changes/ac829cfe826e199b3fea85b2e6fa0fd922d0316c#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1916), reasoning is that  the translator is single-use per query,

strogiyotec: the `joinLookupContext ` now is just assigned  to null after `visitColumnReference`

jyemin (approved):
LGTM!

PR #193, opened 2026-07-27
HIBERNATE-210: support HQL row-value (tuple) predicates

jyemin:
Translate HQL row-value constructor predicates over multiple columns — (a, b) = (…), (a, b) <> (…), (a, b) in ((…), (…)) — to per-component MQL, in both filter ($match) and expression ($expr / SELECT) position.

- Field-vs-value components (in either order) use the compact form: {field: {$eq: value}} joined by $and, negated via $nor. Field-to-field (and value-to-value) comparisons render via $expr, reusing the expression-position decomposition (aggregation $and/$or/$not of $eq).
- IN decomposes to $or of per-row $and; NOT IN negates it.
- Ordering comparisons (>, <, >=, <=) are out of scope and throw FeatureNotSupportedException.


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -664,6 +664,46 @@ public void visitRelationalPredicate(ComparisonPredicate comparisonPredicate) {
...
+        if (lhsTuple != null) {
+            var operator = comparisonPredicate.getOperator();
+            if (operator != ComparisonOperator.EQUAL && operator != ComparisonOperator.NOT_EQUAL) {
+                throw new FeatureNotSupportedException(
+                        "TODO-HIBERNATE-211 https://jira.mongodb.org/browse/HIBERNATE-211");
+            }
+            var rhsTuple = assertNotNull(SqlTupleContainer.getSqlTuple(comparisonPredicate.getRightHandExpression()));
+            var lhsComponents = lhsTuple.getExpressions();
+            var rhsComponents = rhsTuple.getExpressions();
+            assertTrue(lhsComponents.size() == rhsComponents.size());
```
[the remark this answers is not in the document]
jyemin: Assertion is correct in this context as the path is unreachable.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1404,35 +1455,120 @@ public void visitSqlFragmentPredicate(SqlFragmentPredicate sqlFragmentPredicate)
...
-                            .map(item -> acceptAndYield(item, VALUE))
-                            .toList());
-            astVisitorValueHolder.yield(FILTER, new AstFieldOperationFilter(fieldPath, operation));
         }
     }
 
+    // Row-value IN in aggregation-expression position: OR of per-row (AND of per-component $eq); a single-row
+    // list collapses to the bare AND; a negated list is wrapped in $not. Shared by visitInListPredicate's
+    // expression branch and by createTupleInListFilter's $expr fallback, and called directly (never via a second
+    // visit of the predicate) so any parameter it contains is bound exactly once.
```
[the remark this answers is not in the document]
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/query/select/RowValuePredicateIntegrationTests.java
```diff
@@ -0,0 +1,723 @@
...
+                      }
+                    }
+                  ]
+                }""",
+                List.of(new Widget(1, 10, 20)),
+                Set.of(Widget.COLLECTION_NAME));
+    }
+
+    @Test
+    void testFieldEqField() {
```
strogiyotec, line 479: let's add a new method/class to check the null semantics
For example (claude generated test case)
```
    @Test
    void testNeReturnsRowsWithNullComponent() {
        assertSelectionQuery(
                "from NullableWidget w where (w.a, w.b) <> (:a, :b) order by w.id",
                NullableWidget.class,
                q -> q.setParameter("a", 10L).setParameter("b", 20L),
                """
                {
                  "aggregate": "row_value_nullable",
                  "pipeline": [
                    {
                      "$match": {
                        "$nor": [
                          {
                            "$and": [
                              { "a": { "$eq": { "$numberLong": "10" } } },
                              { "b": { "$eq": { "$numberLong": "20" } } }
                            ]
                          }
                        ]
                      }
                    },
                    { "$sort": { "_id": 1 } },
                    {
                      "$project": {
                        "_id": true,
                        "a": true,
                        "b": true
                      }
                    }
                  ]
                }""",
                // SQL three-valued logic would return UNKNOWN for rows 2 and 3 (null component) and exclude them;
                // MQL's $nor-of-$and returns them.
                List.of(
                        new NullableWidget(2, null, 20L),
                        new NullableWidget(3, 10L, null),
                        new NullableWidget(4, 30L, 40L)),
                Set.of(NullableWidget.COLLECTION_NAME));
    }

```
We are checking that rows with null as a value are also returned 
The class definition and presetup are here
```
    @Entity(name = "NullableWidget")
    @Table(name = NullableWidget.COLLECTION_NAME)
    static class NullableWidget {
        static final String COLLECTION_NAME = "row_value_nullable";

        @Id
        long id;

        Long a;

        Long b;

        NullableWidget() {}

        NullableWidget(long id, Long a, Long b) {
            this.id = id;
            this.a = a;
            this.b = b;
        }
    }

    @BeforeEach
    void seed() {
        getSessionFactoryScope().inTransaction(session -> {
            session.persist(new NullableWidget(1, 10L, 20L));
            session.persist(new NullableWidget(2, null, 20L)); // a is null
            session.persist(new NullableWidget(3, 10L, null)); // b is null
            session.persist(new NullableWidget(4, 30L, 40L));
        });
        getTestCommandListener().clear();
    }

```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1404,35 +1452,112 @@ public void visitSqlFragmentPredicate(SqlFragmentPredicate sqlFragmentPredicate)
...
     }
 
+    // Row-value IN in aggregation-expression position: OR of per-row (AND of per-component $eq); a single-row
+    // list collapses to the bare AND; a negated list is wrapped in $not.
+    private AstExpression toTupleInListExpression(InListPredicate inListPredicate) {
+        var keyExpressions = getSqlTuple(inListPredicate.getTestExpression()).getExpressions();
+        var rowExpressions = new ArrayList<AstExpression>(
+                inListPredicate.getListExpressions().size());
+        for (var rowExpression : inListPredicate.getListExpressions()) {
+            var rowValues = assertNotNull(getSqlTuple(rowExpression)).getExpressions();
```
strogiyotec, line 1517: instead of assertions can we throw not supported exception ? 
```
if (rhsTuple == null) {
    throw new FeatureNotSupportedException("Row-value comparison with a subquery is not supported"); // add a hibernate ticket number 
}
```

strogiyotec (dismissed):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: I don't think this is about sub-queries.  I think what this is asserting is that if the key expression is a tuple, then every row expression has to be a tuple as well.  For example:

```
(w.a, w.b) in ((10, 20), (30, 40))
```

the key expression is `(w.a, w.b)` which we've already is ensure is a tuple. Hibernate guarantees that `(10, 20)` and `(30, 40)` are also tuples, otherwise the HQL won't parse.

src/integrationTest/java/com/mongodb/hibernate/query/select/RowValuePredicateIntegrationTests.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: you are right ,had a convo with claude and proved that it's not a sub-query

strogiyotec (approved):

PR #194, opened 2026-07-27
HIBERNATE-87: Support HQL string contatenation

apmasell:
Adds HQL functions for most string operations, including concatenation. Also, build out infrastructure for binding Mongo operators as functions.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
@@ -21,11 +21,12 @@ import com.mongodb.hibernate.query.AbstractQueryIntegrationTests;
 import jakarta.persistence.Entity;
 import jakarta.persistence.Id;
 import jakarta.persistence.Table;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import org.hibernate.query.sqm.produce.function.FunctionArgumentException;
 import org.hibernate.testing.orm.junit.DomainModel;
 import org.hibernate.testing.orm.junit.SessionFactory;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -37,96 +38,451 @@ public class StringFunctionIntegrationTests extends AbstractQueryIntegrationTest
     private static final String COLLECTION_NAME = "items";
     private static final Item HELLO = new Item(1, " Hello ", "_Hello_");
 
-    private void assertQueryResult(String hql, Object expected) {
-        getSessionFactoryScope().inTransaction(session -> {
-            var selectionQuery = session.createSelectionQuery(hql, Object[].class);
-            var resultList = selectionQuery.getResultList();
-            Assertions.assertEquals(1, resultList.size());
-            Assertions.assertEquals(1, resultList.get(0).length);
-            Assertions.assertEquals(expected, resultList.get(0)[0]);
-        });
+    @SuppressWarnings("unchecked")
+    private <T> void assertQueryResult(String hql, T expected, String expectedMql) {
+        assertSelectionQuery(
+                hql, (Class<T>) expected.getClass(), expectedMql, List.of(expected), Set.of(COLLECTION_NAME));
     }
 
     @BeforeEach
     void beforeEach() {
-        getSessionFactoryScope().inTransaction(session -> {
-            session.persist(HELLO);
-        });
+        getSessionFactoryScope().inTransaction(session -> session.persist(HELLO));
     }
 
     @Test
     void testCharacterLength() {
-        assertQueryResult("select length(s) from Item", 7);
+        assertQueryResult(
+                "select length(s) from Item",
+                7,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$strLenCP": "$s"
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testConcat() {
-        assertQueryResult("select s||s from Item", " Hello  Hello ");
+        assertQueryResult(
+                "select s||s from Item",
+                " Hello  Hello ",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$concat": [
+                            {
+                              "$toString": "$s"
+                            },
+                            {
+                              "$toString": "$s"
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testConcatCoerce() {
-        assertQueryResult("select s||3 from Item", " Hello 3");
+        assertQueryResult(
+                "select s||3 from Item",
+                " Hello 3",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$concat": [
+                            {
+                              "$toString": "$s"
+                            },
+                            {
+                              "$toString": 3
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testConcatLiteral() {
-        assertQueryResult("select concat(s, '!') from Item", " Hello !");
+        assertQueryResult(
+                "select concat(s, '!') from Item",
+                " Hello !",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$concat": [
+                            {
+                              "$toString": "$s"
+                            },
+                            {
+                              "$toString": "!"
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLocate() {
-        assertQueryResult("select locate('o', s) from Item", 6);
+        assertQueryResult(
+                "select locate('o', s) from Item",
+                6,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$add": [
+                            {
+                              "$indexOfCP": [
+                                "$s",
+                                "o"
+                              ]
+                            },
+                            {
+                              "$literal": 1
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLocateMissing() {
-        assertQueryResult("select locate('z', s) from Item", 0);
+        assertQueryResult(
+                "select locate('z', s) from Item",
+                0,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$add": [
+                            {
+                              "$indexOfCP": [
+                                "$s",
+                                "z"
+                              ]
+                            },
+                            {
+                              "$literal": 1
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLocateOffset() {
-        assertQueryResult("select locate('H', s, 1, 2) from Item", 2);
+        assertQueryResult(
+                "select locate('H', s, 1, 2) from Item",
+                2,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$add": [
+                            {
+                              "$indexOfCP": [
+                                "$s",
+                                "H",
+                                {
+                                  "$subtract": [
+                                    1,
+                                    {
+                                      "$literal": 1
+                                    }
+                                  ]
+                                },
+                                {
+                                  "$add": [
+                                    {
+                                      "$subtract": [
+                                        1,
+                                        {
+                                          "$literal": 1
+                                        }
+                                      ]
+                                    },
+                                    2
+                                  ]
+                                }
+                              ]
+                            },
+                            {
+                              "$literal": 1
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLocateOffsetMissing() {
-        assertQueryResult("select locate('o', s, 1, 2) from Item", 0);
+        assertQueryResult(
+                "select locate('o', s, 1, 2) from Item",
+                0,
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$add": [
+                            {
+                              "$indexOfCP": [
+                                "$s",
+                                "o",
+                                {
+                                  "$subtract": [
+                                    1,
+                                    {
+                                      "$literal": 1
+                                    }
+                                  ]
+                                },
+                                {
+                                  "$add": [
+                                    {
+                                      "$subtract": [
+                                        1,
+                                        {
+                                          "$literal": 1
+                                        }
+                                      ]
+                                    },
+                                    2
+                                  ]
+                                }
+                              ]
+                            },
+                            {
+                              "$literal": 1
+                            }
+                          ]
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLower() {
-        assertQueryResult("select lower(s) from Item", " hello ");
+        assertQueryResult(
+                "select lower(s) from Item",
+                " hello ",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$toLower": "$s"
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLeftTrim() {
-        assertQueryResult("select ltrim(s) from Item", "Hello ");
+        assertQueryResult(
+                "select ltrim(s) from Item",
+                "Hello ",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$ltrim": {
+                            "input": "$s"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testLeftTrimChars() {
-        assertQueryResult("select ltrim(u, '_') from Item", "Hello_");
+        assertQueryResult(
+                "select ltrim(u, '_') from Item",
+                "Hello_",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$ltrim": {
+                            "chars": "_",
+                            "input": "$u"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testReplaceAll() {
-        assertQueryResult("select replace_all(s, ' ', '_') from Item", "_Hello_");
+        assertQueryResult(
+                "select replace_all(s, ' ', '_') from Item",
+                "_Hello_",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$replaceAll": {
+                            "find": " ",
+                            "input": "$s",
+                            "replacement": "_"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testReplaceOne() {
-        assertQueryResult("select replace_one(s, ' ', '_') from Item", "_Hello ");
+        assertQueryResult(
+                "select replace_one(s, ' ', '_') from Item",
+                "_Hello ",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$replaceOne": {
+                            "find": " ",
+                            "input": "$s",
+                            "replacement": "_"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testRightTrim() {
-        assertQueryResult("select rtrim(s) from Item", " Hello");
+        assertQueryResult(
+                "select rtrim(s) from Item",
+                " Hello",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$rtrim": {
+                            "input": "$s"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
     void testRightTrimChars() {
-        assertQueryResult("select rtrim(u, '_') from Item", "_Hello");
+        assertQueryResult(
+                "select rtrim(u, '_') from Item",
+                "_Hello",
+                """
+                {
+                  "aggregate": "items",
+                  "pipeline": [
+                    {
+                      "$project": {
+                        "#c_1": {
+                          "$rtrim": {
+                            "chars": "_",
+                            "input": "$u"
+                          }
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
     }
 
     @Test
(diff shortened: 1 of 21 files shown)


jyemin:
@apmasell I'm going to wait until you resolve all the Copilot-initiated comment threads before reviewing.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstNamedOperatorExpression.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A general-purpose Mongo operator that includes arguments as a document
+ *
+ * @param operator the name of the operator, including the <code>$</code>
+ * @param arguments a map of name-argument pairs that will be converted into a document
+ */
+public record AstNamedOperatorExpression(String operator, Map<String, AstExpression> arguments)
+        implements AstExpression {
+    @Override
```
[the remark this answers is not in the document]
apmasell: Not relevant.

jyemin (commented):
There are a few bugs in this PR related to parameter binder-related defects.  This is at least the third time our parameter binding design has bit us, and it points to bad design of the translator.  Rather than working around it yet again, I've proposed a fix to the design in https://github.com/mongodb/mongo-hibernate/pull/205.  If we merge that PR and then merge this one on top of it, it should address all the issues and we won't have to think about this source of bugs again.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public void render(
+            SqlAppender sqlAppender,
+            List<? extends SqlAstNode> arguments,
+            ReturnableType<?> returnType,
+            SqlAstTranslator<?> walker) {
+        var translator = AbstractMqlTranslator.cast(walker);
+        var padding = arguments.size() < 3
+                ? new AstLiteralExpression(new AstLiteral(new BsonString(" ")))
+                : new AstUnaryOperatorExpression("$toString", translator.acceptAndYield(arguments.get(2), EXPRESSION));
```
jyemin: This is the one function (besides concat) that coerces with "$toString". For consistency, let's remove it.  Hibernate will already prevent this function from being called on anything but a string anyway.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
```diff
@@ -0,0 +1,130 @@
...
+                                typeConfiguration, FunctionParameterType.STRING),
+                        StandardFunctionArgumentTypeResolvers.impliedOrInvariant(
+                                typeConfiguration, FunctionParameterType.INTEGER),
+                        StandardFunctionArgumentTypeResolvers.impliedOrInvariant(
+                                typeConfiguration, FunctionParameterType.STRING)));
+        this.left = left;
+    }
+
+    @Override
+    public void render(
```
jyemin, line 80: There are a few gotchas with `pad`.

1. `length < 1` should evaluate to empty string
2. `length < lenth(matchedString)` should truncate `matchedString` from the right 

So these tests should pass

```
@Test
    void testPadLeadingTruncatesWhenTargetIsShorterThanInput() {
        assertResult("select pad(s with 3 leading) from Item", " He");
    }

    @Test
    void testPadTrailingTruncatesWhenTargetIsShorterThanInput() {
        assertResult("select pad(s with 3 trailing) from Item", " He");
    }

    @Test
    void testPadWithZeroTargetReturnsEmptyString() {
        assertResult("select pad(s with 0 leading) from Item", "");
    }

    @Test
    void testPadWithNegativeTargetReturnsEmptyString() {
        assertResult("select pad(s with -1 leading) from Item", "");
    }
```

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -284,12 +303,85 @@ public void initializeFunctionRegistry(FunctionContributions functionContributio
...
+                        "locate",
+                        "$indexOfCP",
+                        typeConfiguration,
+                        StandardBasicTypes.INTEGER,
+                        FunctionParameterDefinition::addOne,
+                        allModifications(swap(0, 1), offsetToEnd(2, 3)),
+                        required(FunctionParameterType.STRING),
+                        required(FunctionParameterType.STRING),
+                        orMissing(FunctionParameterType.INTEGER).map(FunctionParameterDefinition::subtractOne),
+                        orMissing(FunctionParameterType.INTEGER)));
```
jyemin: I think we should stick to Hibernate's 2 or 3 argument `locate` function rather than add our own 4th argument.

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
```diff
@@ -0,0 +1,943 @@
...
+                        }
+                      }
+                    }
+                  ]
+                }
+                """);
+    }
+
+    @Test
+    void testLocateOffset() {
```
jyemin: Let's test the three-argument form as well (regardless of whether we remove the four-argument form).

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
```diff
@@ -0,0 +1,89 @@
...
+import org.hibernate.type.StandardBasicTypes;
+import org.hibernate.type.spi.TypeConfiguration;
+
+public final class MongoTrimFunction extends AbstractSqmSelfRenderingFunctionDescriptor implements ExpressionFunction {
+
+    public MongoTrimFunction(TypeConfiguration typeConfiguration) {
+        super(
+                "trim",
+                new ArgumentTypesValidator(
+                        StandardArgumentsValidators.between(2, 3),
```
jyemin: This can be `exactly(3)` since Hibernate guarantees it.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public void render(
+            SqlAppender sqlAppender,
+            List<? extends SqlAstNode> arguments,
+            ReturnableType<?> returnType,
+            SqlAstTranslator<?> walker) {
+        var translator = AbstractMqlTranslator.cast(walker);
+        if (arguments.size() != 2 && arguments.size() != 3) {
```
jyemin: By our convention, this should be an assertion since Hibernate should guarantee it, e.g. `assertTrue(arguments.size == 3)`.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
```diff
@@ -0,0 +1,89 @@
...
+            throw new IllegalArgumentException("Invalid number of arguments to trim");
+        }
+        final var namedArguments = new TreeMap<String, AstExpression>();
```
jyemin: Nit: remove `final`

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
```diff
@@ -0,0 +1,89 @@
...
+        namedArguments.put("input", translator.acceptAndYield(arguments.get(arguments.size() - 1), EXPRESSION));
```
jyemin: Hibernate guarantees three arguments so this can become `arguments.get(2)`

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
```diff
@@ -0,0 +1,89 @@
...
+
+        if (arguments.size() == 3) {
```
jyemin: The conditional can be removed since Hibernate guarantees three args.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -284,12 +303,85 @@ public void initializeFunctionRegistry(FunctionContributions functionContributio
...
+        functionRegistry.register("array_list", new MongoArrayConstructorFunction(true));
+        functionRegistry.register(
+                "character_length",
+                new MongoExpressionUnaryFunction(
+                        "character_length",
+                        "$strLenCP",
+                        typeConfiguration,
+                        StandardBasicTypes.INTEGER,
+                        FunctionParameterType.STRING));
+        functionRegistry.register(
```
jyemin, line 316: Let's require a minimum of 1 argument to match Hibernate's existing behavior.

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
```diff
@@ -0,0 +1,943 @@
...
+                                  }
+                                }
+                              }
+                            }
+                          ]
+                        }
+                """);
+    }
+
+    @Test
```
jyemin, line 2626: There's some edge cases that need to be accounted for with substring: SQL standard defines the result as the characters whose positions fall in [start, start + len) intersected with [1, length], so a start below 1 is (unfortunately, perhaps) legal and clips the length.

```
    @Test
    void testSubstringWithZeroStart() {
        assertResult("select substring(s, 0) from Item", " Hello ");
    }

    @Test
    void testSubstringFromNegativeStartForLength() {
        assertResult("select substring(s -1, 3) from Item", " ");
    }

    @Test
    void testSubstringWithZeroStartAndLength() {
        assertResult("select substring(s, 0, 3) from Item", " H");
    }
```

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -284,12 +303,85 @@ public void initializeFunctionRegistry(FunctionContributions functionContributio
...
+                        orMissing(FunctionParameterType.INTEGER)));
+        functionRegistry.register(
+                "lower",
+                new MongoExpressionUnaryFunction(
+                        "lower",
+                        "$toLower",
+                        typeConfiguration,
+                        StandardBasicTypes.STRING,
+                        FunctionParameterType.STRING));
+        functionRegistry.register("lpad", new MongoPadFunction(typeConfiguration, true));
```
jyemin, line 348: Let's not support `lpad` or `rpad`, with all their complications, and instead just support `pad`, which only allows a single-character padding string.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExpressionNamedFunction.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public void render(
+            SqlAppender sqlAppender,
+            List<? extends SqlAstNode> arguments,
+            ReturnableType<?> returnType,
+            SqlAstTranslator<?> walker) {
+        var translator = AbstractMqlTranslator.cast(walker);
+        final var namedArguments = new TreeMap<String, AstExpression>();
```
jyemin: Let's kill all the `final` on local variables.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
```diff
@@ -0,0 +1,469 @@
...
+
+    static final class PositionalCheck implements RelativePositionChecker {
+        private int required;
+        private int withDefault;
+        private int missing;
+
+        @Override
+        public void incrementRequired() {
+            if (withDefault > 0 || missing > 0) {
+                throw new IllegalArgumentException(
```
jyemin: I think by convention all the `IllegalArgumentException` in this PR should instead be assertions, via `MongoAssertions` methods.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
```diff
@@ -0,0 +1,469 @@
...
+     * Define an optional parameter that will have a default value inserted if not provided in the HQL
+     *
+     * <p>Note that the type of this parameter is inferred to match the type of the default value provided.
+     *
+     * @param name the property name that should be emitted in the MQL
+     * @param defaultValue the default value to insert if a matching argument is not supplied in the HQL
+     * @return a parameter definition
+     * @see MongoExpressionNamedFunction
+     */
+    public static FunctionParameterDefinition<String> orDefault(String name, BsonValue defaultValue) {
```
jyemin: This method is currently unused.  Let's delete it and add it later if it becomes necessary.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
```diff
@@ -0,0 +1,469 @@
...
+
+    /**
+     * Define an optional parameter that will be omitted if not provided in the HQL
+     *
+     * @param type the type of the parameter
+     * @return a parameter definition
+     * @see MongoExpressionNamedFunction
+     */
+    @SuppressWarnings("NullAway")
+    public static Missing<Void> orMissing(FunctionParameterType type) {
```
jyemin: Let's change both `orMissing` methods to return `FunctionParameterDefinition<Void>` and then make the `Missing` class package-private to match the rest of the `FunctionParameterDefinition` implementations.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
```diff
@@ -0,0 +1,469 @@
...
+    @SuppressWarnings("varargs")
+    static <N> FunctionArgumentTypeResolver typeResolverFromParameters(
+            TypeConfiguration typeConfiguration, FunctionParameterDefinition<N>... parameters) {
+        return StandardFunctionArgumentTypeResolvers.byArgument(Stream.of(parameters)
+                .map(parameter ->
+                        StandardFunctionArgumentTypeResolvers.impliedOrInvariant(typeConfiguration, parameter.type()))
+                .toArray(FunctionArgumentTypeResolver[]::new));
+    }
+
+    @SuppressWarnings("varargs")
```
jyemin: Looks like this annotation is no longer needed.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstLetBindingExpression.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.mongodb.hibernate.internal.translate.mongoast;
+
+import java.util.SortedMap;
+import java.util.function.Consumer;
+import org.bson.BsonWriter;
+import org.hibernate.sql.exec.spi.JdbcParameterBinder;
+
+/** @hidden */
+@SuppressWarnings("MissingSummary")
```
jyemin: Let's be consistent and add a summary, as in the other two new expressions.

jyemin (changes requested):

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
apmasell: Tests added and implementation updated.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Updated.

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
apmasell: Changed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
apmasell: This entire check isn't required and I've removed it.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
apmasell: I think I got all of them.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoTrimFunction.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Updated in the variadic function.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: I think I've ironed out the inconsistencies. Since Hibernate expects to delegate to these functions in the other dialects, I'd rather keep the separate implementations.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExpressionNamedFunction.java
apmasell: Fixed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
apmasell: I've removed most of these. I can't remove it from the `switch` because the assertion method returns void.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
apmasell: Updated.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstLetBindingExpression.java
apmasell: Added.

src/main/java/module-info.java
```diff
@@ -117,6 +117,7 @@
     requires transitive org.mongodb.driver.core;
     requires transitive org.mongodb.driver.sync.client;
     requires org.jspecify;
+    requires jdk.compiler;
```
jyemin: Not sure how this slipped in but it can be removed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
jyemin: Still seeing the toString here.  Did you remove a different one?

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
jyemin: Sounds good.

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
jyemin: Another edge case: `adjustedStart` clamps the start, but the length is not floored, so a start far enough below 1 makes the third argument to `$substrCP` negative. On `s = " Hello "`:

| HQL | result |
|---|---|
| `substring(s, -5, 2)` | `Location34454`, `$substrCP: length must be a nonnegative integer` |
| `substring(s, -10, 2)` | same |
| `substring(s, :st, :ln)` with -5 and 2 | same |
| `substring(s, <int field> - 10, 2)` | same |

`substring(s, -1, 3)` returns `" "` by the rule in my earlier comment, and that same rule makes all four of these `""`.

Please floor the length at 0, so that the `length` local emits

```
{"$max": [{"$add": ["$$len", {"$subtract": ["$$start", "$$adjustedStart"]}]}, 0]}
```

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
```diff
@@ -284,12 +300,75 @@ public void initializeFunctionRegistry(FunctionContributions functionContributio
...
+                new MongoExpressionPositionalFunction(
+                        "locate",
+                        "$indexOfCP",
+                        typeConfiguration,
+                        StandardBasicTypes.INTEGER,
+                        FunctionParameterDefinition::addOne,
+                        swap(0, 1),
+                        required(FunctionParameterType.STRING),
+                        required(FunctionParameterType.STRING),
+                        orMissing(FunctionParameterType.INTEGER).map(FunctionParameterDefinition::subtractOne)));
```
jyemin: `locate` subtracts one from the start and hands it straight to `$indexOfCP`, so a start below 1 reaches the server: `locate('o', s, 0)` gives `Location40097`, `$indexOfCP requires a nonnegative starting index, found: -1`.

Please floor that argument at 0, so the third element of `$indexOfCP` becomes `{"$max": [{"$subtract": [...]}, 0]}`.   Something like:

```
orMissing(FunctionParameterType.INTEGER)
          .map(FunctionParameterDefinition::subtractOne)
          .map(FunctionParameterDefinition::atLeastZero)
```

Test cases: `locate('o', s, 0)` and `locate('o', s, -3)` both give 6, while `locate(' ', s, 4)` stays 7

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
jyemin: All eight pad tests emit one of four pipelines: lpad and rpad crossed with default and explicit padding. The `$cond` tree is emitted whole every time and its arms are selected by the server from the three values in `vars`, so the open coverage question is which arms execute.

| arm | lpad | rpad |
|---|---|---|
| `targetLen <= 0` | `pad(s with -1 leading)` | missing |
| `strLenCP(baseStr) >= targetLen` | `pad(s with 3 leading)` | missing |
| `strLenCP(padding) == 0` | `lpad(s, 60, '')` | missing |
| pad, single-character padding | `with 20 leading[ '*']` | `with 20 trailing[ '*']` |
| pad, multi-character, non-exact division | `lpad('x', 6, 'ab')` | missing |
| pad, multi-character, exact division | missing | missing |
| pad, padding longer than the gap | missing | missing |

Please cover the missing arms. Results on `s = " Hello "`:

| HQL | result |
|---|---|
| `pad(s with -1 trailing)` | `""` |
| `pad(s with 0 leading)` | `""` |
| `pad(s with 3 trailing)` | `" He"` |
| `rpad(s, 60, '')` | `" Hello "` |
| `rpad('abc', 6, 'ab')` | `"abcaba"` |
| `lpad('ab', 6, 'ab')` | `"ababab"` |
| `rpad('ab', 6, 'ab')` | `"ababab"` |
| `lpad('x', 3, 'abcdef')` | `"abx"` |
| `rpad('x', 3, 'abcdef')` | `"xab"` |

These all emit a pipeline the existing tests already asserts, so we should try to avoid duplicating the pipelines in every new test. I suggest adding a `@ParameterizedTest` per pipeline shape, with one row per arm. `pad(s with N leading 'c')` and `lpad(a, n, p)` resolve to the same descriptor, so rows can mix both forms, which also covers `rpad` by name for the first time. 

`MongoSubstringFunction` has the same structure, two pipelines with the `adjustedStart` `$cond` chosen at runtime, so the clamping cases for `substring` fit the same treatment.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
apmasell: Oops. Yes. I think I got them all.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
apmasell: I've added the tests, but I'm unclear what you're asking for with `@ParameterizedTest`. Are you looking to have each test check both `pad` and `lpad`/`rpad` with the same paramters?

src/main/java/module-info.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoDialect.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoPadFunction.java
jyemin: Here's an example of what I mean:

```java

    @ParameterizedTest
    @CsvSource({
        "' ', 4, 7",
        "'o', 0, 6",
        "'o', -3, 6",
    })
    void testLocateWithStart(String pattern, int start, int expected) {
        assertQueryResult(
                "select locate('%s', s, %d) from Item".formatted(pattern, start),
                expected,
                """
                {
                  "aggregate": "items",
                  "pipeline": [
                    {
                      "$project": {
                        "#c_1": {
                          "$add": [
                            {
                              "$indexOfCP": [
                                "$s",
                                "%s",
                                {
                                  "$let": {
                                    "in": {
                                      "$cond": [
                                        {
                                          "$lt": [
                                            "$$x",
                                            {
                                              "$literal": 0
                                            }
                                          ]
                                        },
                                        {
                                          "$literal": 0
                                        },
                                        "$$x"
                                      ]
                                    },
                                    "vars": {
                                      "x": {
                                        "$subtract": [
                                          %d,
                                          {
                                            "$literal": 1
                                          }
                                        ]
                                      }
                                    }
                                  }
                                }
                              ]
                            },
                            {
                              "$literal": 1
                            }
                          ]
                        }
                      }
                    }
                  ]
                }
                """
                        .formatted(pattern, start));

```

The same transform applies to the 16 pad tests: They emit four pipelines, lpad and rpad crossed with default and explicit padding, so they reduce to four parameterized tests:

| pipeline | current tests |
|---|---|
| lpad, default padding | `testPadLeadingWS`, `testPadLeadingTruncatesWhenTargetIsShorterThanInput`, `testPadLeadingWithNegativeTargetReturnsEmptyString` |
| lpad, explicit padding | `testPadLeadingCh`, `testLpadPadsToRequestedLengthWithMultiCharacterPadding`, `testLpadWithEmptyPaddingReturnsInput`, `testLpadMuliCharExact`, `testLpadLongPadding` |
| rpad, default padding | `testPadTrailingWS`, `testPadTrailingTruncatesWhenTargetIsShorterThanInput`, `testPadTrailingWithNegativeTargetReturnsEmptyString` |
| rpad, explicit padding | `testPadTrailingCh`, `testRpadPadsToRequestedLengthWithMultiCharacterPadding`, `testRpadWithEmptyPaddingReturnsInput`, `testRpadMuliCharExact`, `testRpadLongPadding` |

The four three-argument substring tests, `testSubstringLen`, `testSubstringFromNegativeStartForLength`, `testSubstringWithZeroStartAndLength` and `testSubstringNegative`, reduce the same way.

apmasell: That is much more difficult to read, follow, and maintain.

jyemin: I don't agree, but it's a judgement call.  If you want to keep it like it is, then I will abide by that.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
```diff
@@ -223,6 +215,27 @@ public final FunctionParameterDefinition<N> map(Function<? super AstExpression,
         return new Mapped<>(this, mapper);
     }
 
+    /**
+     * Ensures that an input value, which must be a number, is greater than zero
+     *
+     * @param input the expression to clamp
+     * @return an expression which returns the original value or zero if the original value is less than zero
+     */
+    public static AstExpression atLeastZero(AstExpression input) {
```
jyemin, line 219: Let's replace this with [$max](https://www.mongodb.com/docs/manual/reference/operator/aggregation/max/#use-in-project-stage), like. 

```
{"$max": [<input>, 0]}
```

jyemin (changes requested):
One more simplification request.

src/main/java/com/mongodb/hibernate/internal/dialect/function/FunctionParameterDefinition.java
apmasell: Fixed.

jyemin (approved):
LGTM

src/integrationTest/java/com/mongodb/hibernate/query/function/string/StringFunctionIntegrationTests.java
jyemin: Sorry, just noticed this one has no response yet.

apmasell: This got an `atLeastZero`.

PR #196, opened 2026-07-30
HIBERNATE-213: fix release publishing for the multi-module build

jyemin:
Fixes three defects in the release process.

1. publish.sh expanded the Gradle task list as `"${TASK}"`, making the two release task names a single argv element, so Gradle looked for one task named `"publishArchives closeAndReleaseSonatypeStagingRepository"`. A bash array expanded with `"${TASKS[@]}"` passes them separately. Snapshot publishing was unaffected because `publishSnapshots` is a single word, which is why this went unnoticed until the first release attempt. `SYSTEM_PROPERTIES` becomes an array for the same reason, and RELEASE now uses `${RELEASE:-false}`.

2. `publishArchives` and `publishSnapshots` declared `dependsOn(tasks.named(...))`, which resolves to the root project's tasks alone, so the publications of the Spring Boot modules were never uploaded and the build still passed. Both lifecycle tasks now depend on the equivalent task in every project that applies `mongo-hibernate-publish`.

3. The starter module applies plain `java-library` rather than the Java conventions, so it published neither a sources nor a javadoc jar; Maven Central requires both, and the staging repository would have failed to close. Producing the two jars moves from `mongo-hibernate-java` to `mongo-hibernate-publish`, since the requirement follows from publishing a module rather than from how the module is compiled.

### DIFF
--- a/.evergreen/publish.sh
+++ b/.evergreen/publish.sh
@@ -9,7 +9,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 ############################################
 source java-config.sh
 
-RELEASE=${RELEASE:false}
+RELEASE=${RELEASE:-false}
 
 export ORG_GRADLE_PROJECT_nexusUsername=${NEXUS_USERNAME}
 export ORG_GRADLE_PROJECT_nexusPassword=${NEXUS_PASSWORD}
@@ -17,12 +17,12 @@ export ORG_GRADLE_PROJECT_signingKey="${SIGNING_KEY}"
 export ORG_GRADLE_PROJECT_signingPassword=${SIGNING_PASSWORD}
 
 if [ "$RELEASE" == "true" ]; then
-  TASK="publishArchives closeAndReleaseSonatypeStagingRepository"
+  TASKS=(publishArchives closeAndReleaseSonatypeStagingRepository)
 else
-  TASK="publishSnapshots"
+  TASKS=(publishSnapshots)
 fi
 
-SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true"
+SYSTEM_PROPERTIES=(-Dorg.gradle.internal.publish.checksums.insecure=true)
 
 ./gradlew -version
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info "${TASK}"
+./gradlew "${SYSTEM_PROPERTIES[@]}" --stacktrace --info "${TASKS[@]}"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,7 +15,6 @@
  */
 
 import java.time.Duration
-import java.util.concurrent.Callable
 import net.ltgt.gradle.errorprone.errorprone
 
 plugins {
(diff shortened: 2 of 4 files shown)


.evergreen/publish.sh
```diff
@@ -9,20 +9,20 @@ set -o errexit  # Exit the script with error if any of the commands fail
 ############################################
 source java-config.sh
 
-RELEASE=${RELEASE:false}
```
strogiyotec, line 12: should I create a follow up for java driver ? (maybe spark and kafka too) as java driver has the same issue
https://github.com/mongodb/mongo-java-driver/blob/main/.evergreen/publish.sh#L13
On the other hand, I checked that `RELEASE` variable is always set in pipeline so this fix should not have blocked us during the release

.evergreen/publish.sh
```diff
@@ -9,20 +9,20 @@ set -o errexit  # Exit the script with error if any of the commands fail
...
 else
-  TASK="publishSnapshots"
+  TASKS=(publishSnapshots)
 fi
 
-SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true"
+SYSTEM_PROPERTIES=(-Dorg.gradle.internal.publish.checksums.insecure=true)
 
 ./gradlew -version
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info "${TASK}"
```
strogiyotec, line 28: I believe the problem is is quotes `"` , java driver has the same pattern but doesn't use quotes instead of introducing the array , why do you think having array is better ?

.evergreen/publish.sh
```diff
@@ -9,20 +9,20 @@ set -o errexit  # Exit the script with error if any of the commands fail
...
 
 if [ "$RELEASE" == "true" ]; then
-  TASK="publishArchives closeAndReleaseSonatypeStagingRepository"
+  TASKS=(publishArchives closeAndReleaseSonatypeStagingRepository)
 else
-  TASK="publishSnapshots"
+  TASKS=(publishSnapshots)
 fi
 
-SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true"
```
strogiyotec, line 25: same here https://github.com/mongodb/mongo-hibernate/pull/196/changes#r3709481140

build.gradle.kts
```diff
@@ -212,14 +213,21 @@ val gitVersion: String by lazy {
         .getOrElse("UNKNOWN")
 }
 
+// A plain `tasks.named(name)` below would resolve to the root project's task alone, leaving the
+// publications of the subprojects unpublished. The Callable defers looking at the projects until the task
+// graph is built, by which point the subprojects have been evaluated and their plugins are visible.
+fun publishingTaskInEveryProject(name: String) = Callable {
```
strogiyotec: how about adding 1 validation layer 
```
fun publishingTaskInEveryProject(name: String) = Callable {
    val publishing = allprojects.filter { it.pluginManager.hasPlugin("mongo-hibernate-publish") }
    val withPublications = allprojects.filter {
        it.extensions.findByType<PublishingExtension>()?.publications?.isNotEmpty() == true
    }
    require((withPublications - publishing.toSet()).isEmpty()) {
        "Projects declare publications but don't apply mongo-hibernate-publish: ${withPublications - publishing.toSet()}"
    }
    publishing.map { it.tasks.named(name) }
}
```
This way we check that every project that want to be published must declare `mongo-hibernate-publish`

I am still debating if it will be useful, for example if I were to introduce a new module like Quarkus starter I would just copy paste the Spring boot starter

strogiyotec (dismissed):

.evergreen/publish.sh
rozza: ~I agree - just don't quote when calling `./gradlew` - follows the java driver convention~

Claude seems to think the array approach is safer because if quotes were added in the future it would still work. 🤷

.evergreen/publish.sh
```diff
@@ -9,20 +9,20 @@ set -o errexit  # Exit the script with error if any of the commands fail
 ############################################
 source java-config.sh
 
-RELEASE=${RELEASE:false}
+RELEASE=${RELEASE:-false}
```
rozza, line 12: 👍 Nice catch.

.evergreen/publish.sh
rozza: Its worth it, if you do a PR feel free to assign to me

build.gradle.kts
rozza: Nit: use `Provider` to follow convention else where and saves an import:

> `Callable` and `Provider` are both in the set of types it unpacks lazily (alongside `Task`, `TaskProvider`), and both get their body invoked at that point, not at registration.

```suggestion
fun publishingTaskInEveryProject(name: String) = provider {
```

rozza (dismissed):
Just one nit (optional). 

I wondered why not just register the task inside the convention - but this seems to be the better approach as there is one single publish (archives/snapshot) task registered - rather than one per sub project.

FYI - `buildSrc` is not under spotless - something you may want to add in the future.

.evergreen/publish.sh
jyemin: Yes, and this is exactly how it broke in the first place: quotes were added around `TASKS` as a drive-by change, probably in response to a warning in the IDE.

build.gradle.kts
jyemin: Great idea.  Tested, and applied.

jyemin: Done, and import removed.

rozza (approved):
LGTM!

PR #198, opened 2026-08-01
HIBERNATE-220: executeUpdate returns the JDBC row count for updates

jyemin:
JDBC's Statement#executeUpdate contract is the count of rows matched by the DML statement, not the count of rows whose values actually changed. The UPDATE arm of MongoStatement.getUpdateCount used getModifiedCount(), which under-reports when a bulk HQL UPDATE matches a document that already holds the target value. Switch to getMatchedCount().

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/mutation/UpdatingIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/mutation/UpdatingIntegrationTests.java
@@ -337,6 +337,52 @@ class UpdatingIntegrationTests extends AbstractQueryIntegrationTests {
                 Set.of(Book.COLLECTION_NAME));
     }
 
+    @Test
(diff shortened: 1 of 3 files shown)


apmasell (commented):
Is there a better source for the interface contract, because https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeUpdate-java.lang.String- does not indicate either behaviour.

jyemin:
> Is there a better source for the interface contract?

Good question, and I looked but couldn't find one.  The justification for this change is that it's both what Postgres does and what Hibernate expects.  It didn't bite us until now, but without this change we can't get `StatelessSession#upsert` to work properly.

apmasell:
That's worth documenting.

jyemin:
@apmasell documentation added as a comment.

apmasell (approved):

PR #199, opened 2026-08-01
HIBERNATE-217: support StatelessSession.upsert on unversioned entities

jyemin:
Translate the OptionalTableUpdate built for StatelessSession.upsert and upsertMultiple into a single update statement: filter from the key bindings, $set over value bindings that are both insertable and updatable, $setOnInsert for insertable-only ones, upsert: true, multi: false.

visitOptionalTableUpdate serves two callers that hand it the same node: the boot-time forced version increment wants a plain update, the dialect wants an upsert. The caller's intent travels as the requested value descriptor, tested with expects(...).

The dialect translates eagerly at SessionFactory build time and defers every rejection to performMutation, since an eager throw would fail boot for entities that work today for everything except upsert: @Version entities throw TODO-HIBERNATE-216, attributes that are updatable but not insertable and entities whose only persistent attribute is the identifier throw bare messages (MQL has no matched-path-only write), and optional tables keep the TODO-HIBERNATE-69 rejection.

AstUpdateCommand now holds a list of {q, u, upsert?, multi} statements and renders byte-identically for every existing caller; AstDocumentUpdate gains $setOnInsert, rendering each operator only when non-empty and asserting the update is never empty; MongoStatement accepts upsert in update statements and extends the JDBC row count to matched plus upserted.

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
+++ b/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
@@ -34,6 +34,7 @@ import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor
 import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.SELECT_RESULT;
 import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.SORT_FIELDS;
 import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.TUPLE;
+import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.UPSERT_MODEL_MUTATION_RESULT;
 import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.VALUE;
 import static com.mongodb.hibernate.internal.translate.mongoast.AstLiteral.FALSE;
 import static com.mongodb.hibernate.internal.translate.mongoast.AstLiteral.TRUE;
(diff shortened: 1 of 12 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1065,7 +1072,9 @@ public void visitUpdateStatement(UpdateStatement updateStatement) {
         astVisitorValueHolder.yield(
                 MUTATION_RESULT,
                 new MutationMqlTranslator.Result(
-                        new AstUpdateCommand(collection, filter, update), parameterBinders, affectedTableNames));
+                        new AstUpdateCommand(collection, List.of(new AstUpdateStatement(filter, update, false, true))),
```
strogiyotec: can we create 2 subclasses of `AstUpdateCommand` 
1. `MultiUpdate` which sets `upsert=false` and `multi=true`
2. `Upsert` which sets `upsert=true` and `multi=false`
3. We might need the third variant but I haven't seen a usage and a unit test that has both multi and upsert set to true, though according to documentation it's a valid setting

strogiyotec (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Introduced two factory methods instead, which required making `AstUpdateStatement` a class instead of a record, which is unfortunate.  But overall it's an improvement.

strogiyotec (approved):
LGTM

PR #201, opened 2026-08-05
Update dependencies

jyemin:
Got an alert about assertj vulnerability, figured it was time to update all of our out of date dependencies.


jyemin:
Since this is basically Dependabot work, I merged without review.

PR #202, opened 2026-08-06
HIBERNATE-83 Support HQL CASE expression

strogiyotec:
Support HQL CASE expression
According to Hibernate user guide https://docs.hibernate.org/orm/8.0/userguide/html_single
there are two types of case statements
1. [Simple case expressions](https://docs.hibernate.org/orm/8.0/userguide/html_single/#hql-simple-case-expressions) 
2. [Search case expressions ](https://docs.hibernate.org/orm/8.0/userguide/html_single/#hql-searched-case-expressions)

A CASE only yields an aggregation expression, so positions that need a bare field path will throw unsupported exception (I haven't made a new ticket to support it )

For example this test [case ](https://github.com/mongodb/mongo-hibernate/pull/202/changes#diff-346ff10eafa7c3aa3de77e7b8a2ae02c6ecabf34ede4457b8d80780ef21f98d9R979)
```
        void testCaseAsLikeMatchExpression() {
            assertSelectQueryFailure(
                    "from Item where (case when y > 5 then 'a' else 'b' end) like '%a%'",
                    Item.class,
                    "CASE expression is only supported in aggregation-expression position");
        }
```

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
+++ b/src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
@@ -73,6 +73,8 @@ import com.mongodb.hibernate.internal.translate.mongoast.AstLogicalOperatorExpre
 import com.mongodb.hibernate.internal.translate.mongoast.AstNode;
 import com.mongodb.hibernate.internal.translate.mongoast.AstParameterMarker;
 import com.mongodb.hibernate.internal.translate.mongoast.AstRegexMatchExpression;
+import com.mongodb.hibernate.internal.translate.mongoast.AstSwitchCase;
+import com.mongodb.hibernate.internal.translate.mongoast.AstSwitchExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstUnaryOperatorExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstValue;
 import com.mongodb.hibernate.internal.translate.mongoast.AstValueExpression;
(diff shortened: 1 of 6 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1299,12 +1301,58 @@ private static AstExpression divide(
...
-        throw new FeatureNotSupportedException("TODO-HIBERNATE-83 https://jira.mongodb.org/browse/HIBERNATE-83");
+        assertCaseExpressionPosition();
+        // Simple CASE: the fixture is compared for equality against each `when` value, so each branch's
+        // `case` is a {$eq: [fixture, value]} expression. The fixture renders once per branch, so it must
+        // be visited once per branch too — visiting once and reusing the result would emit a single
+        // parameter binder for a fixture that renders N times, misaligning positional binding.
+        var branches = new ArrayList<AstSwitchCase>(
+                caseSimpleExpression.getWhenFragments().size());
+        for (var whenFragment : caseSimpleExpression.getWhenFragments()) {
+            var fixture = acceptAndYieldExpression(caseSimpleExpression.getFixture());
```
strogiyotec, line 1332: this thing was flagged by review skill, originally `fixture` was created outside of the loop
but then claude flagged it using this test [case](https://github.com/mongodb/mongo-hibernate/pull/202/changes#diff-346ff10eafa7c3aa3de77e7b8a2ae02c6ecabf34ede4457b8d80780ef21f98d9R843)
```
                    "select case :p when 7 then 1 when 8 then 2 else 0 end from Item",
```
the `p` has to be rendered twice for both when cases , but it only rendered once and later gave an exception saying that parameter with index [2] is not set, moving the fixture inside the loop fixed the problem

jyemin: I wonder if we need a better parameter marker design that is more robust in the face of re-ordering or duplication.

jyemin (approved):
LGTM

PR #205, opened 2026-08-06
HIBERNATE-222: collect parameter binders while rendering the MQL AST

jyemin:
A JDBC parameter renders as BSON undefined and carries no identity into the emitted command. MongoPreparedStatement recovers the parameter positions by walking the parsed command depth-first, and Hibernate binds the i-th JdbcParameterBinder into the i-th of those positions. The binder list was built separately, in the order visitParameter ran, so the two orders were equal only by convention and nothing compared them. Permuting translated expressions, using one translated node in two positions, or collecting named arguments into a sorted map each broke the pairing, two of the three silently.

AstParameterMarker now carries its own binder and AstNode.render takes a Consumer<JdbcParameterBinder>, so the marker appends its binder in the same statement that writes the undefined. The binder list is produced by the rendering itself and the translator no longer accumulates one. Reordering, dropping and duplicating translated nodes are all correct as a result.

### DIFF
--- a/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstNodeAssertions.java
+++ b/src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstNodeAssertions.java
@@ -21,30 +21,43 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.io.IOException;
 import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
 import org.bson.json.JsonWriter;
+import org.hibernate.sql.exec.spi.JdbcParameterBinder;
+import org.junit.jupiter.params.shadow.de.siegmar.fastcsv.util.Nullable;
 
 public final class AstNodeAssertions {
 
     private AstNodeAssertions() {}
 
     public static void assertRendering(String expectedCanonicalExtendedJson, AstNode node) {
-        doAssertRendering(expectedCanonicalExtendedJson, node, AstNodeKind.DOCUMENT);
+        doAssertRendering(expectedCanonicalExtendedJson, null, node, AstNodeKind.DOCUMENT);
     }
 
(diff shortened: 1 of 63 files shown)


src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstNodeAssertions.java
```diff
@@ -54,7 +54,7 @@ private static void doAssertRendering(String expectedJson, AstNode node, AstNode
             if (nodeKind == AstNodeKind.VALUE) {
                 jsonWriter.writeName(ancillaryFieldName);
             }
-            node.render(jsonWriter);
+            node.render(jsonWriter, parameterBinder -> {});
```
apmasell: Looking at the test suite, I don't think any tests will trigger this `Consumer`. If they did, would the binding be relevant? Or effectively implied by the MQL?

apmasell (commented):
Looks good.

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstNodeAssertions.java
jyemin: I confirmed that there are two existing tests that trigger this `Consumer`:

1. `AstInsertCommandTests#testRendering`
2. `AstAllFilterOperationTests#testRendering`

But the triggering is orthogonal to those tests, so not really relevant.

However, your comment made me think a bit more about this and I realized that this method _should_ be augmented to (optionally) assert on the emitted binders, and also that there should be a dedicated unit test for `AstParameterMarker` which makes use of that assertion.  So I've done that.

Additionally, I refactored `ParameterBinderCollectionTests` to use this new ability instead of implementing the assertion itself.  This has the nice side effect of removing its dependency on a translator method that it shouldn't have been using in the first place, and was doing so just for expediency.

apmasell (approved):

strogiyotec (approved):
LGTM !

PR #207, opened 2026-08-10
HIBERNATE-207: support composite primary key

jyemin:
Support a composite primary key declared with @EmbeddedId on a plain @Embeddable, stored as the document's _id sub-document.

Boot (MongoAdditionalMappingContributor): name each id component column _id.<component> instead of rejecting a multi-column primary key, in a canonical by-name order so the field-order-sensitive _id unique index is stable. Reject three unsupported id shapes with FeatureNotSupportedException naming their support tickets: @IdClass (HIBERNATE-235, also resolving the HIBERNATE-230 boot NullPointerException), a non-scalar id component (HIBERNATE-236), and derived identity / @MapsId (HIBERNATE-237).

Translator (AbstractMqlTranslator): gather the _id.<component> leaf bindings into a nested _id sub-document on insert (managed persist and HQL bulk INSERT, aligning the id's spanning value expression to its columns), and AND the per-component key bindings in the mutation key filter. Filtering, ordering, and projection over the id need no composite-specific code: the _id.<component> column naming routes them through the existing path handling and tuple translation.


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -373,6 +375,29 @@ public <R> void yield(AstVisitorValueDescriptor<R> valueDescriptor, R value) {
         astVisitorValueHolder.yield(valueDescriptor, value);
     }
 
+    // Column bindings for a composite id are flattened by boot-time metadata into sibling
+    // "_id.<component>" columns; gather them back into a single leading "_id" sub-document, with
+    // components ordered by name for a canonical shape, since the "_id" unique index is sensitive
+    // to BSON field order.
+    private static List<AstElement> assembleWithIdSubdocument(List<AstElement> flat) {
+        var idComponents = new TreeMap<String, AstValue>();
```
strogiyotec: NIT: we don't really need a TreeMap here, we can use TreeSet instead to avoid this second loop that converts map into a list of 
```
        var idComponents = new TreeSet<>(comparing(AstElement::name));
        var result = new ArrayList<AstElement>(flat.size());
        for (var element : flat) {
            if (EmbeddedIdColumnName.isComponent(element.name())) {
                idComponents.add(new AstElement(EmbeddedIdColumnName.componentName(element.name()), element.value()));
            } else {
                result.add(element);
            }
        }
        if (!idComponents.isEmpty()) {
// no second loop 
            result.add(0, new AstElement(ID_FIELD_NAME, new AstDocument(idComponents)));
        }
        return result;
```

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -346,9 +415,15 @@ private static void setIdentifierColumnName(PersistentClass persistentClass) {
...
-            throw new FeatureNotSupportedException(format(
-                    "%s: %s does not support primary key spanning multiple columns %s",
-                    persistentClass, MONGO_DBMS_NAME, idColumns));
+            // Non-scalar id components (nested embeddable or collection) were already rejected by
+            // forbidNonScalarIdComponent, so each component of this composite id has exactly one column.
+            var idComponent = assertInstanceOf(identifier, Component.class);
+            for (var property : idComponent.getProperties()) {
+                var componentColumns = property.getValue().getColumns();
+                assertTrue(componentColumns.size() == 1);
+                componentColumns.get(0).setName(ID_FIELD_NAME + "." + property.getName());
```
strogiyotec: I think there is an overlap that we can abstract to avoid working with string manipulations 
1. in this case we build an embedded name manually
2. in MqlTranslator we then do `if (element.name().startsWith(ID_FIELD_NAME + "."))`
3. and then `element.name().substring(ID_FIELD_NAME.length() + 1)`
I suggest we abstract this logic away 

https://github.com/jyemin/mongo-hibernate/pull/4
I asked claude to draft my idea into a separate class and it came up with a utility class

WDYT ? Feel free to skip it , but I personally found it to be more readable what the same abstraction is grouped together

strogiyotec (changes requested):
First review

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Applied

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Merged your PR.  Thanks!

strogiyotec (approved):
LGTM  !

PR #209, opened 2026-08-14
HIBERNATE-88: Support HQL date functions

apmasell:
Add support for `format` and `extract` HQL functions.

### DIFF
--- a/src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
+++ b/src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
@@ -16,7 +16,8 @@
 
 package com.mongodb.hibernate.internal.dialect.function;
 
-import static com.mongodb.hibernate.internal.dialect.function.FunctionParameterDefinition.divideAndSomethingAsInt;
+import static com.mongodb.hibernate.internal.dialect.function.FunctionParameterDefinition.addOne;
+import static com.mongodb.hibernate.internal.dialect.function.FunctionParameterDefinition.ceilingDivideAsInt;
 import static com.mongodb.hibernate.internal.translate.AstVisitorValueDescriptor.EXPRESSION;
 import static org.hibernate.query.sqm.produce.function.FunctionParameterType.TEMPORAL;
 import static org.hibernate.query.sqm.produce.function.FunctionParameterType.TEMPORAL_UNIT;
@@ -25,6 +26,7 @@ import com.mongodb.hibernate.internal.FeatureNotSupportedException;
 import com.mongodb.hibernate.internal.translate.AbstractMqlTranslator;
 import com.mongodb.hibernate.internal.translate.mongoast.AstArithmeticExpressionOperator;
 import com.mongodb.hibernate.internal.translate.mongoast.AstBinaryOperatorExpression;
+import com.mongodb.hibernate.internal.translate.mongoast.AstExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstLetBindingExpression;
 import com.mongodb.hibernate.internal.translate.mongoast.AstLiteral;
 import com.mongodb.hibernate.internal.translate.mongoast.AstLiteralExpression;
@@ -193,7 +195,7 @@ public final class MongoExtractFunction extends AbstractSqmSelfRenderingFunction
                                 new TreeMap<>(Map.of("time", input)));
                     case NATIVE -> new AstUnaryOperatorExpression("$toDate", input);
                     case QUARTER ->
-                        divideAndSomethingAsInt(
+                        ceilingDivideAsInt(
                                 new AstNamedOperatorExpression(
                                         "$month",
                                         new TreeMap<>(Map.of(
@@ -202,8 +204,7 @@ public final class MongoExtractFunction extends AbstractSqmSelfRenderingFunction
                                                 "timezone",
                                                 new AstLiteralExpression(new AstLiteral(new BsonString(
                                                         ZoneId.systemDefault().getId())))))),
-                                3,
-                                "$ceil");
+                                3);
                     case SECOND ->
                         new AstLetBindingExpression(
                                 new AstBinaryOperatorExpression(
@@ -215,18 +216,10 @@ public final class MongoExtractFunction extends AbstractSqmSelfRenderingFunction
                                                         "$millisecond", new AstVariableExpression("time")),
                                                 new AstLiteralExpression(new AstLiteral(new BsonInt32(1000))))),
                                 new TreeMap<>(Map.of("time", input)));
-                    case WEEK_OF_YEAR ->
-                        new AstNamedOperatorExpression(
-                                "$isoWeek",
-                                new TreeMap<>(Map.of(
-                                        "date",
-                                        input,
-                                        "timezone",
-                                        new AstLiteralExpression(new AstLiteral(new BsonString(
-                                                ZoneId.systemDefault().getId()))))));
+                    case WEEK_OF_YEAR -> weekOfYear(input);
                     case WEEK ->
                         new AstNamedOperatorExpression(
-                                "$week",
+                                "$isoWeek",
                                 new TreeMap<>(Map.of(
                                         "date",
                                         input,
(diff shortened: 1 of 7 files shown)


jyemin (changes requested):
I shared a review generated using the code review skill in the repo.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
```diff
@@ -0,0 +1,311 @@
...
+                                                                "date",
+                                                                new AstVariableExpression("time"),
+                                                                "timezone",
+                                                                new AstLiteralExpression(
+                                                                        new AstLiteral(
+                                                                                new BsonString(
+                                                                                        ZoneId.systemDefault()
+                                                                                                .getId())))))),
+                                                new AstUnaryOperatorExpression(
+                                                        "$week",
```
jyemin: This `$week` expression is missing a timezone, which IIUC will be a bug in some time zones.
But instead of adding it, let's re-use the code we already have for WEEK_OF_YEAR. The existing `weekOfYear` method can be changed to 

```
private static AstExpression sundayBasedWeek(AstExpression input, String dayOfPeriodOperator) {
    ...
                          dateOperator(dayOfPeriodOperator, time),
                          dateOperator("$dayOfWeek", time)),
    ....
}
```

and the two cases then become:

```
     case WEEK_OF_YEAR -> sundayBasedWeek(input, "$dayOfYear");
     case WEEK_OF_MONTH -> sundayBasedWeek(input, "$dayOfMonth");
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1409,7 +1490,36 @@ public void visitExtractUnit(ExtractUnit extractUnit) {
...
+                inQuote = !inQuote;
+            } else if (inQuote) {
+                outputFormat.append(inputFormat.charAt(0));
+                if (inputFormat.charAt(0) == '%') {
+                    outputFormat.append('%');
+                }
+                inputFormat = inputFormat.substring(1);
+            } else {
+                for (var entry : DATE_FORMATS) {
+                    if (inputFormat.startsWith(entry.hql())) {
```
jyemin: Because this is matching on the input format prefix, it accepts patterns that it doesn't intend to.  In particular, it intends to accept `"Z"`, `"ZZ"`, and `"ZZZ"`, but it also accepts `"ZZZZ"`, `"ZZZZZ"`, etc.

I'm not sure the right way to fix it.  Perhaps it needs to first get the run of identical characters from the input format, and compare that against each `DATE_FORMATS` entry by equality?

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -274,13 +274,94 @@
...
+            new DateFormatRule("y", Optional.empty()),
+
+            // month of year
+            new DateFormatRule("MMMM", Optional.of("%B")),
+            new DateFormatRule("MMM", Optional.of("%b")),
+            new DateFormatRule("MM", Optional.of("%m")),
+            new DateFormatRule("M", Optional.empty()),
+
+            // week of year (this looks like %U, but does not match Java's implmentation)
+            new DateFormatRule("ww", Optional.empty()),
```
jyemin: IIUC `"%V"` matches HQL semantics for `"ww"`.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: That's fair. This does seem like it should be handled in the Hibernate validation logic. I'll redesign it.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: Added.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: I've made this throw `IllegalArgumentException` since it's not really a missing feature.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
apmasell: Sure. I was also missing it on the extract logic for second and milliseconds.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Looks like it throws `FeatureNotSupportedException` now, as discussed in standup.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
```diff
@@ -0,0 +1,174 @@
...
+            ReturnableType<?> returnType,
+            SqlAstTranslator<?> walker) {
+        var translator = AbstractMqlTranslator.cast(walker);
+
+        var input = translator.acceptAndYield(sqlAstArguments.get(1), EXPRESSION);
+        var unit = ((ExtractUnit) sqlAstArguments.get(0)).getUnit();
+        translator.yield(
+                EXPRESSION,
+                switch (unit) {
+                    case DATE ->
```
jyemin: This case is not tested, and when I tried to add a test it fails in the JDBC layer with

```
org.hibernate.JDBCException: Could not extract column [1] from JDBC ResultSet [getDate not implemented]
```

I suggest removing this case and letting it fall through to the default.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
```diff
@@ -0,0 +1,174 @@
...
+                                                        dateOperator("$millisecond", new AstVariableExpression("time")),
+                                                        new AstLiteralExpression(
+                                                                new AstLiteral(new BsonInt32(1_000_000)))),
+                                                new AstBinaryOperatorExpression(
+                                                        AstArithmeticExpressionOperator.MULTIPLY,
+                                                        dateOperator("$second", new AstVariableExpression("time")),
+                                                        new AstLiteralExpression(
+                                                                new AstLiteral(new BsonInt32(1_000_000_000)))))),
+                                new TreeMap<>(Map.of("time", input)));
+                    case NATIVE -> new AstUnaryOperatorExpression("$toDate", input);
```
jyemin: This case appears to be unreachable.  I suggest removing it.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1409,7 +1527,49 @@ public void visitExtractUnit(ExtractUnit extractUnit) {
...
+                        var code = dateFormats.get(length);
+                        if (code.isPresent()) {
+                            outputFormat.append(code.get());
+                        } else {
+                            throw new FeatureNotSupportedException("Unsupported date format: "
+                                    + inputFormat.substring(inputIndex, inputIndex + length));
+                        }
+                        inputIndex += length - 1;
+                    } else {
+                        throw new FeatureNotSupportedException("Format code %s is ambiguous."
```
jyemin, line 1564: Is it always (or ever) ambiguous, or is it sometimes just unsupported?  Suggest to change to something like "Unsupported format code..."

jyemin (changes requested):
Looking good.  Found a few more small items to address, but it's very close now.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
apmasell: Removed.

src/main/java/com/mongodb/hibernate/internal/dialect/function/MongoExtractFunction.java
apmasell: Removed and a unsupported test is added.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
apmasell: Hibernate's `SqlFormat.FORMAT` is as follows:
```
Pattern.compile( "('[^']+'|[:;/,.!@#$^&?~`|()\\[\\]{}<>\\-+*=]|\\s|G{1,2}|[yY]{1,4}|M{1,4}|w{1,2}|W|E{3,4}|e{1,2}|d{1,2}|D{1,3}|a|[Hhms]{1,2}|S{1,6}|[zZx]{1,3})*");
```
which are all covered by the values in the map and if its in the map with no Mongo equivalent, an unsupported error is produced. Because this regular expression allows these to be repeated, these "too long" format codes pass validation, but, I think, they are invalid. When such a code is passed to the Java date formatter, it rejects it:

```
jshell> java.time.format.DateTimeFormatter.ofPattern("ZZZZZZZZZZZZZZZZZZZZZZZZZZ")
|  Exception java.lang.IllegalArgumentException: Too many pattern letters: Z
|        at DateTimeFormatterBuilder.parsePattern (DateTimeFormatterBuilder.java:1970)
|        at DateTimeFormatterBuilder.appendPattern (DateTimeFormatterBuilder.java:1908)
|        at DateTimeFormatter.ofPattern (DateTimeFormatter.java:569)
|        at (#1:1)
```

and the Hibernate formatter is supposed to be a subset of the Java date formatter, based on my reading of the comments in `SqmFormat`. We also have a difference for the code `E` since Hibernate restricts to `EEE` or `EEEE`, but the Java date formatter takes `E` just fine:

```
jshell> java.time.format.DateTimeFormatter.ofPattern("E")
$2 ==> Text(DayOfWeek,SHORT)
```

So, it is ambiguous in the Hibernate definition since it is a concatenation of valid codes and it is illegal in the Java reference implementation.

jyemin (approved):
LGTM!

PR #218, opened 2026-08-21
HIBERNATE-248: publish the SSDLC compliance report and CodeQL SARIF files

jyemin:
Adds the two SSDLC release assets to publish-snapshot and publish-release, uploaded to the release-assets bucket alongside the papertrail trace.

download-codeql-sarifs.py polls the code-scanning analyses API for the commit being published until every CodeQL category has an analysis, then downloads and validates each SARIF. Evergreen and the CodeQL workflow react to the same push concurrently, so a lookup without polling may find nothing and upload an empty directory. The script fails the task instead.

ssdlc-report.sh renders template_ssdlc_compliance_report.md, ported from mongo-java-driver. The findings section describes CodeQL alerts and where their dismissal reasons live, and the dependency section lists Hibernate ORM, the Java driver and Spring Boot. The release creator comes from HEAD, which is the published commit on both paths; deriving it from the version leaves the field blank whenever git describe reports a dirty tree.

Keys the CodeQL concurrency group on the commit as well as the ref, giving every commit a group of its own so that its analysis run survives whatever lands next. A publish task waits for the analyses of the exact commit it publishes, and while all of main shared one group, the next push cancelled that run and left the task to fail on its poll timeout. Cancellation now only collapses two runs of the same commit.

### DIFF
--- a/.evergreen/template_ssdlc_compliance_report.md
+++ b/.evergreen/template_ssdlc_compliance_report.md
@@ -1,0 +1,67 @@
+# ${product_name} SSDLC compliance report
+
+This report is available at
+<https://d-9067613a84.awsapps.com/start/#/console?account_id=857654397073&role_name=Drivers.User&destination=https%3a%2f%2fus-west-1.console.aws.amazon.com%2fs3%2fobject%2fjava-driver-release-assets%3fregion%3dus-west-1%26bucketType%3dgeneral%26prefix%3d${product_name}%2f${product_version}%2fssdlc_compliance_report.md>.
+
+<table>
+  <tr>
+    <th>Product name</th>
+    <td><a href="https://github.com/mongodb/mongo-hibernate">${product_name}</a></td>
+  </tr>
+  <tr>
+    <th>Product version</th>
+    <td>${product_version}</td>
+  </tr>
+  <tr>
+    <th>Release creator</th>
+    <td>
+        ${product_release_creator}
+        <p>
+            Refer to data in Papertrail for more details.
+            There is currently no official way to serve that data.
+        </p>
+    </td>
+  </tr>
+  <tr>
+    <th>Report date, UTC</th>
+    <td>${report_date_utc}</td>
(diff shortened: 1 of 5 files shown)


.evergreen/template_ssdlc_compliance_report.md
```diff
@@ -0,0 +1,67 @@
...
+
+The MongoDB SSDLC policy is available at
+<https://docs.google.com/document/d/1u0m4Kj2Ny30zU74KoEFCN4L6D_FbEYCaJ3CQdCYXTMc>.
+
+## Third-party dependency information
+
+This product depends on Hibernate ORM and the MongoDB Java driver, and its Spring Boot
+starter modules depend on Spring Boot.
+Our [SBOM](https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/docs/CYCLONEDX/) lite
+is <https://github.com/mongodb/mongo-hibernate/blob/r${product_version}/sbom.json>.
```
strogiyotec, line 43: we don't have a sbom.json file yet
The java driver has it [here](https://github.com/mongodb/mongo-java-driver/blob/9af9be1fa428effe0ea10e05c0e1256d2286ec3c/sbom.json#L4)
https://github.com/mongodb/mongo-java-driver/pull/1387 in this PR Valentin introduced an empty sbom which is not necessary as we won't be using SBOM but from the jira ticket comments it's still recomended to add an empty sbom file 
```
{
  "serialNumber": "urn:uuid:a291eaa6-9c96-4c46-9fb1-474f745cf6f5",
  "version": 1,
  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.5"
}
```
and replace serialNumber with any random uuid

strogiyotec (changes requested):

strogiyotec (approved):

PR #220, opened 2026-08-24
HIBERNATE-224: unwrap domain values in @Struct fields and HQL literals

jyemin:
`ValueConversions.toBsonValue` dispatches on the runtime class of a value and recognizes only JDBC-level classes. Every write path reaches it through BasicBinder.getBindValue, which first unwraps the domain value to the `JdbcType`'s preferred Java type, so what reaches `toBsonValue` is always a class it recognizes. Two paths skipped that step and passed the domain value straight through: the fields of a @Struct aggregate embeddable, and an inlined HQL literal. A type whose domain class is not itself a JDBC-level class therefore worked in every position except those two.

The HQL literal path now unwraps through the same `ValueBinder` that binding a parameter of that type already uses. The @Struct write path delegates to Hibernate's own `StructHelper.getJdbcValues`, which flattens the domain value to one JDBC value per column and applies each column's `ValueBinder` on the way, subsuming both the per-column unwrap and the existing array special case.

Delegating in this way also promotes a shape we previously refused: a flattened @Embeddable nested in a @Struct one now stores as fields of the enclosing subdocument and reads back intact, because StructHelper recurses into an embeddable-valued attribute.

The @Struct read path has a similar defect and gets the inverse change: read the JDBC-level value the binder would have written, then wrap it back through the mapped JavaType. This is not symmetric in form with the write side, because an ordinary column has `BasicExtractor.doExtract` do the wrapping while the @Struct read path has no `ValueExtractor.` A JdbcType reporting no preferred Java type binds the domain value unchanged, so it is read back unchanged. Materializing the embeddable itself, rather than reading its fields, uses `StructHelper.getAttributeValues`, the inverse of `getJdbcValues`, so that an attribute spanning more than one column is recomposed from its columns rather than taken one column per attribute.

Additionally the types java.time.Duration, java.time.Year, java.time.ZoneId, java.time.ZoneOffset and java.util.TimeZone already mapped correctly, but now they are documented and tested.

 Note this is a pre-condition for HIBERNATE-225 because `ZonedDateTime/OffsetDateTime` will fail to convert when in a @Struct embeddable field or an HQL literal.

HIBERNATE-224

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/QueryLiteralIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/QueryLiteralIntegrationTests.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2025-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.query.select;
+
+import static java.lang.String.format;
+
+import com.mongodb.hibernate.query.AbstractQueryIntegrationTests;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import java.math.BigDecimal;
+import java.time.Instant;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Stream;
+import org.bson.types.ObjectId;
+import org.hibernate.testing.orm.junit.DomainModel;
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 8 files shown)


src/integrationTest/java/com/mongodb/hibernate/query/select/QueryLiteralIntegrationTests.java
```diff
@@ -0,0 +1,132 @@
...
+        BigDecimal bigDecimalValue;
+        ObjectId objectIdValue;
+        Instant instantValue;
+
+        Item() {}
+
+        /** Item 1 holds every constant's value; item 2 holds values none of the constants match. */
+        Item(int id) {
+            this.id = id;
+            var matching = id == 1;
```
apmasell: What's the reason for using this matching logic vs creating two instances with different constructor parameters? Since the alternate values are also constants, shouldn't they also be in `QueryLiteralConstants`?

jyemin: Yeah, that was a bit lazy.  Fixed by adding two factory methods, one for each Item.  Let me know if that reads better.

I left the non-matching values inline rather than in `QueryLiteralConstants`. That class only holds the values referenced by fully-qualified name from HQL, which is the mechanism being tested. The non-matching values are never named in a query, so putting them there would make the class's purpose ambiguous.

As an example, this is the first HQL formatted in `testLiteralPredicate`:

```
select id from Item where stringValue = com.mongodb.hibernate.query.select.QueryLiteralConstants.STRING
```

apmasell (approved):

PR #221, opened 2026-08-25
HIBERNATE-208 exclude _id from $project when unspecified

jyemin:
Inject _id: 0 into the terminal $project when it does not already specify _id, in the shared MongoStatement.executeQuery path, and stop appending _id to the result-set field names. Result-set columns now match exactly the projected fields, for both generated and native queries, and MongoDB no longer returns an unused _id per document.

Prototype: only one integration-test MQL assertion (testProjectUsingAlias) is updated; the remaining ~53 full-pipeline assertions still expect the old projection and will fail until updated.

HIBERNATE-208


vbabanin (dismissed):
LGTM!

vbabanin (approved):

PR #222, opened 2026-08-25
HIBERNATE-241: Support Expression in GROUP BY and HAVING

vbabanin:
## What this adds

A GROUP BY key may be an arbitrary expression, not only a column reference, and a reference to that key in SELECT or HAVING resolves to the key.

```
select b.primitiveInt + 1 from Item as b GROUP BY b.primitiveInt + 1
```

```json
{"$group":   {"_id": {"k0": {"$add": ["$primitiveInt", 1]}}}},
{"$project": {"#c_1": "$_id.k0"}}
```

A key may also appear nested inside a larger expression, more than once, or in HAVING:

```
select (b.primitiveInt + 1) * 2 from Item as b
GROUP BY b.primitiveInt + 1 HAVING b.primitiveInt + 1 > 0
```

```json
{"$group":   {"_id": {"k0": {"$add": ["$primitiveInt", 1]}}}},
{"$match":   {"_id.k0": {"$gt": 0}}},
{"$project": {"#c_1": {"$multiply": ["$_id.k0", 2]}}}
```

Function keys and CASE keys work the same way.

## How a reference is matched to a key

Matching is structural: an expression matches a key when it is written identically to it. `1 + x` does not match a key of `x + 1` (like PostgreSQL). Expressions are identified by value number, so a key and its occurrences compare in constant time rather than by repeated subtree comparison, and a key occurring several times in SELECT resolves to a single `_id` sub-key.

A query parameter is identified by its parameter id rather than by its binder, so one parameter occurring in both the GROUP BY and the SELECT is one expression.

## Columns that are not keys

A column appearing in SELECT, HAVING or ORDER BY that is neither a GROUP BY key nor contained in one is rejected with `FeatureNotSupportedException`, rather than emitting a pipeline that reads a field `$group` did not produce.

## Not supported yet

| Shape | Ticket |
| --- | --- |
| Aggregate functions | HIBERNATE-239 |
| ORDER BY an expression that is a GROUP BY key | HIBERNATE-251 |
| SELECT DISTINCT together with GROUP BY | HIBERNATE-205 |

## Implementation

Substitution runs as a tree rewrite over the MQL AST. `AstRewriter` walks a node once, applying the optional pre-rules top down and the optional post-rule bottom up. `GroupBySubstitutionRule` replaces a key occurrence with its `_id` sub-key, and `ExprToMatchDowngradeRule` turns the resulting `{$expr: {$gt: ["$_id.k0", 0]}}` into the compact `{"_id.k0": {"$gt": 0}}` form that is equivalent to the rendering of the WHERE clause.

Descent is a per-node responsibility: `AstNode.mapChildren` returns a copy of the node with each child rewritten, so a new node type cannot compile without deciding what its children are. Rules are typed per hierarchy, so a replacement has to stay in the hierarchy it replaces.

Each expression describes its own structure as a `StructuralKey` of its kind, its data and its children, and `VNRegistry` assigns the numbers.

Note that with the subsequent refactorings, the design doc is stale in some places.

**Intractable walkthrough:** [link](https://pages.monglify.aix.prod.corp.mongodb.com/p/tiger-rocket-buffalo)


jyemin (dismissed):

vbabanin:
//TODO remove **docs/superpowers/specs/2026-07-20-hibernate-groupby-expression-keys-design.md** before merging.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+ * Assigns a canonical value number (VN) to each distinct expression structure, so that two structurally equal
+ * expressions compare in O(1) once their children are numbered, and GROUP BY key matching needs no repeated subtree
+ * comparison.
+ *
+ * <p>An expression describes itself as a {@link StructuralKey}; numbering and caching those descriptions is this
+ * class's business.
+ *
+ * <p>Matching is structural, not algebraic: {@code a + 1} and {@code 1 + a} get distinct value numbers. This is
+ * structural value numbering (Alpern, Wegman, and Zadeck, "Detecting equality of variables in programs", POPL '88), in
+ * tree form rather than over SSA. The interning table is a hash-cons (Ershov, "On programming of arithmetic
```
apmasell, line 34: I'm unclear on what the Ershov paper is contributing to this. Can you elaborate?

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstMapChildrenAssertions.java
```diff
@@ -0,0 +1,235 @@
...
+            assertEquals(
+                    sorted(rewriter.produced),
+                    sorted(childrenOf(result)),
+                    "mapChildren must rebuild the node from what the rewriter returned");
+        }
+    }
+
+    /** Ordered by rendering, so a node is free to visit its children in whatever order suits it. */
+    private static List<String> sorted(List<AstNode> nodes) {
+        return nodes.stream().map(Object::toString).sorted().toList();
```
apmasell, line 91: I think this is fine, but if you're trying to show that the set of AST nodes are visited a consistent number of times, I think having something like `Map<AstNode, Integer>` and incrementing the values is clearer and then you can compare the maps for equality.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java
jyemin: @vbabanin this one's for you.

src/test/java/com/mongodb/hibernate/internal/translate/mongoast/AstMapChildrenAssertions.java
jyemin: The intention was not to show that the set of AST nodes are visited a consistent number of times. So an implementation that rewrote the same node twice is not something it's trying to catch.  However, in looking at this I realized there are bugs that this set of assertions will not catch.  For example:

```
public record AstSwitchCase(AstExpression caseExpression, AstExpression thenExpression) implements AstNode {
    @Override
    public AstSwitchCase mapChildren(AstNodeRewriter rewriter) {
        return new AstSwitchCase(rewriter.rewrite(thenExpression), rewriter.rewrite(caseExpression));
    }
 ```

Transposing the two rewritten nodes passes all assertions. I tightened it up to catch this case, as well as a few others.

**Review by strogiyotec — DISMISSED** [review]

**Review by jyemin — DISMISSED** [review]

File: src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java (hunk 1 above)
Comment (vbabanin) [reply src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java:34]: The registry is a modern restatement of the A2 algorithm from that paper. Ershov’s downstream problem was different from ours - he was doing common-subexpression elimination in a BESM compiler, so repeated arithmetic subexpressions would not result in duplicate machine instructions. But the mechanism he introduced for recognizing that “these two subtrees are the same” is essentially what we use:

* Key each expression by `(operator, canonical-child-IDs)`, where the child IDs are already-canonicalized VNs.
* Look it up in a hash table.
* On a miss, allocate a new ID; on a hit, after tuple equality is confirmed, reuse the existing one.
* Build bottom-up so equality at each node is O(1): compare operand IDs rather than walking the subtrees.

Ershov used open-addressed linear probing in a fixed BPO table and treated the table slot address itself as the canonical ID. We use a `HashMap` and assign sequence numbers, but the underlying approach is essentially the same.

The paper is also cited in later literature as the earliest published instance of this kind of structural hash-consing/value-numbering technique. That is the property `GroupBySubstitutionRule` relies on when deciding that, for example, `SELECT x + 1` matches `GROUP BY x + 1`.

There are secondary sources supporting that interpretation:
ACL2 / University of Texas: 

> “It appears that hash consing was first conceived by A. P. Ershov in 1957, to speed up the recognition of common subexpressions.”
https://www.cs.utexas.edu/~moore/acl2/v6-3/HONS-AND-MEMOIZATION.html

Henry Baker, ACM Lisp Pointers (1992): 
> “Hash consing was invented by Ershov in 1958 … to speed up the recognition of common subexpressions, and popularized by Goto [1974].”
https://citeseerx.ist.psu.edu/document?doi=86f676eaab5de044996d5eb55066fbc157390c41&repid=rep1&type=pdf

File: src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+ * operations", CACM 1(8), 1958; Filliatre and Conchon, "Type-safe modular hash-consing", ML Workshop 2006), except that
+ * it stores the canonical integer rather than the canonical object.
```
vbabanin, line 36:
```suggestion
 * <p>Matching is structural, not algebraic: {@code a + 1} and {@code 1 + a} get distinct value numbers. This is                                                            
   * value numbering (Cocke & Schwartz, <i>Programming Languages and Their Compilers</i>, NYU 1970) in tree form: key                                                         
   * each expression by its operator and the canonical identifiers of its already-numbered operands, so per-node                                                              
   * equality is a constant-time tuple comparison rather than a subtree walk. The interning mechanism is hash-cons which is similar to Ershov's                                                             
   * ("On programming of arithmetic operations", CACM 1(8), 1958).                                                                                            
```

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/VNRegistry.java
vbabanin: Related: https://github.com/mongodb/mongo-hibernate/pull/222/changes#r3910185217

jyemin: @apmasell if you agree with the edit, please accept it and resolve this thread.

apmasell: Sure, that sounds good.

apmasell (approved):

jyemin (approved):

PR #225, opened 2026-09-03
Attach test results from all modules in Evergreen; fix naming test cleanup

jyemin:
## Attach test results from all modules in Evergreen

The `upload-test-results` function attached only the root module's `build/test-results` paths, so the spring-boot-autoconfigure module's unit and integration test results never appeared in Evergreen's test-results view even though the tasks executed them. This adds the `src/*/build/test-results` globs, matching the convention the `trace-artifacts` function already uses.

## Drop the namingbook collection via the Mongo client in naming test cleanup

`repository.deleteAll()` hydrates every document in the namingbook collection, which the two naming tests share while storing different field keys (`publishYear` vs `publish_year`) under their different physical naming strategies. A document written by one test leaves the other test's primitive `int publishYear` null when hydrated, and Hibernate's generated access optimizer throws NullPointerException unboxing it. Once such a document is left behind by an interrupted run, every later run of both tests fails, because the failing cleanup never removes the offending documents. Dropping the collection through the client never hydrates anything, so cleanup is unaffected by documents written under either naming strategy.

### DIFF
--- a/mongodb-hibernate-spring-boot-autoconfigure/src/integrationTest/java/com/mongodb/hibernate/naming/NamingTestSupport.java
+++ b/mongodb-hibernate-spring-boot-autoconfigure/src/integrationTest/java/com/mongodb/hibernate/naming/NamingTestSupport.java
@@ -27,6 +27,18 @@ final class NamingTestSupport {
 
     private NamingTestSupport() {}
 
+    // Drops the collection via the Mongo client rather than emptying it through the repository: the
+    // collection is shared by the two naming tests, whose naming strategies store different field keys,
+    // and repository.deleteAll would fail to hydrate documents written under the other strategy.
+    static void dropBooksCollection(String connectionString) {
+        try (var client = MongoClients.create(connectionString)) {
+            var databaseName = Objects.requireNonNull(
+                    new ConnectionString(connectionString).getDatabase(),
+                    "connection string must include a database name");
+            client.getDatabase(databaseName).getCollection("namingbook").drop();
+        }
+    }
+
     // Reads the raw stored document straight from MongoDB (bypassing Hibernate) so the test inspects the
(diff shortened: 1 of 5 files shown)


mongodb-hibernate-spring-boot-autoconfigure/src/integrationTest/java/com/mongodb/hibernate/naming/NamingTestSupport.java
```diff
@@ -27,6 +27,18 @@ final class NamingTestSupport {
...
 
+    // Drops the collection via the Mongo client rather than emptying it through the repository: the
+    // collection is shared by the two naming tests, whose naming strategies store different field keys,
+    // and repository.deleteAll would fail to hydrate documents written under the other strategy.
+    static void dropBooksCollection(String connectionString) {
+        try (var client = MongoClients.create(connectionString)) {
+            var databaseName = Objects.requireNonNull(
+                    new ConnectionString(connectionString).getDatabase(),
+                    "connection string must include a database name");
+            client.getDatabase(databaseName).getCollection("namingbook").drop();
```
vbabanin: We usually keep collection names as constants on the entity so they can be reused, e.g. `Book.COLLECTION_NAME` https://github.com/vbabanin/mongo-hibernate/blob/193c5c7a501aa1c90eb6c3caa582c992707bc684/src/integrationTest/java/com/mongodb/hibernate/query/Book.java#L27.

Should we do the same for `NamingBook` and define the collection name as a constant there?

jyemin: Done

vbabanin (approved):
LGTM!

PR #227, opened 2026-09-03
HIBERNATE-233: support sequence-style identifier generation for @GeneratedValue

jyemin:
Back @GeneratedValue with a MongoDB counter collection. MongoSequenceSupport generates the four commands Hibernate ORM's SequenceStyleGenerator asks a dialect for, and MongoStatement runs the allocation as a findAndModify whose $setOnInsert seeds the counter and whose pipeline update advances it by the increment.

Counters live in one document per sequence in hibernate_sequences, keyed by the sequence's formatted name:

  {"_id": "<schema.>name", "next_value": <int64>, "increment": <int64>}

Allocation deliberately does not join the caller's transaction. Sharing the session would make two concurrent allocations conflict, and a rollback would rewind the counter and hand the same identifier out twice.

Because a counter document is ordinary data that a deployment running without schema management may create by hand, the shape above is enforced rather than assumed. An allocation filter requires next_value and increment to be int64, so a malformed or partial document fails loudly instead of silently returning the same identifier forever. MongoSequenceInformationExtractor reports each counter's increment, which is what lets Hibernate ORM reject a mapping whose allocationSize disagrees with the counter it would allocate from.

Sequence qualifiers follow the rules tables already follow. A catalog is rejected, reaching a sequence either through @SequenceGenerator(catalog) or through a three-part sequenceName that Hibernate ORM reads as catalog, schema, sequence. A '.' in a sequence schema or name is rejected, which makes the counter key unambiguous: two sequences can no longer format alike and share one counter. These checks live in an Integrator because sequences are registered after AdditionalMappingContributor runs.

Rejected at boot, each with a message naming the alternative: IDENTITY, TABLE and UUID generation, BigInteger identifiers, generators that cannot be introspected, sequence options, and an entity mapped onto hibernate_sequences.

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
@@ -1,0 +1,940 @@
+/*
+ * Copyright 2026-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.id;
+
+import static com.mongodb.hibernate.internal.MongoConstants.MONGO_CONFIGURATION_CONTRIBUTOR_KEY;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+import com.mongodb.client.MongoCollection;
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
+import com.mongodb.hibernate.junit.CommandHistory;
+import com.mongodb.hibernate.junit.InjectCommandHistory;
+import com.mongodb.hibernate.junit.InjectMongoCollection;
+import com.mongodb.hibernate.junit.MongoExtension;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.SequenceGenerator;
+import jakarta.persistence.Table;
+import jakarta.persistence.TableGenerator;
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.function.Function;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+import org.assertj.core.api.InstanceOfAssertFactories;
+import org.bson.BsonDocument;
+import org.hibernate.MappingException;
+import org.hibernate.Session;
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Sequence-style identifier generation, from schema export through allocation to drop.
+ *
+ * <p>Each test boots its own {@code SessionFactory} with {@code create-drop}. {@link MongoExtension} empties every
+ * collection in this class's database before each test, {@code hibernate_sequences} among them, so a
+ * {@code SessionFactory} shared across the class would leave later tests with no counter document. A pooled optimizer
+ * also holds its block in the generator, which lives in the {@code SessionFactory}, so any assertion about round trips
+ * needs a fresh one.
+ */
+@ExtendWith(MongoExtension.class)
+class SequenceGeneratedIdIntegrationTests {
+
+    private static final Map<String, Object> BASE_SETTINGS = Map.of(
+            "jakarta.persistence.schema-generation.database.action",
+            "create-drop",
+            "hibernate.hbm2ddl.halt_on_error",
+            "true");
+
+    @InjectCommandHistory
+    private CommandHistory commandHistory;
+
+    @InjectMongoCollection("hibernate_sequences")
+    private MongoCollection<BsonDocument> sequences;
+
+    @InjectMongoCollection("books")
+    private MongoCollection<BsonDocument> books;
+
+    /** What one booted {@code SessionFactory} produced: the commands sent, and whatever the body observed. */
+    private record Run<T>(List<BsonDocument> commands, T observed) {}
+
+    /** The identifiers a persist-then-reload round trip observed, and the counter document at that point. */
+    private record PersistedBook(long id, long reloadedId, BsonDocument sequenceDocument) {}
+
+    /**
+     * Boots a {@code SessionFactory} for {@code entityClasses} with {@code create-drop}, runs {@code body} while it is
+     * open, and returns the commands sent along with the body's result.
+     *
+     * <p>The registry is built by hand so it can apply the contributor that points the {@code SessionFactory} at this
+     * class's own database and installs that database's command listener.
+     */
+    private <T> Run<T> inRegistry(Function<Session, T> body, Class<?>... entityClasses) {
+        return inRegistry(Map.of(), body, entityClasses);
+    }
+
+    private <T> Run<T> inRegistry(
+            Map<String, Object> additionalSettings, Function<Session, T> body, Class<?>... entityClasses) {
+        try (var registry = new StandardServiceRegistryBuilder()
+                .applySettings(BASE_SETTINGS)
+                .applySettings(additionalSettings)
+                .applySetting(
+                        MONGO_CONFIGURATION_CONTRIBUTOR_KEY,
+                        MongoExtension.configurationContributorForClass(SequenceGeneratedIdIntegrationTests.class))
+                .build()) {
+            T observed;
+            var metadataSources = new MetadataSources();
+            for (var entityClass : entityClasses) {
+                metadataSources.addAnnotatedClass(entityClass);
+            }
+            try (var sessionFactory = metadataSources.buildMetadata(registry).buildSessionFactory();
+                    var session = sessionFactory.openSession()) {
+                observed = body.apply(session);
+            }
+            return new Run<>(commandHistory.getCommands(), observed);
+        }
+    }
+
+    /** The commands sent whose name matches {@code commandName}. */
+    private static List<BsonDocument> commandsNamed(List<BsonDocument> commands, String commandName) {
+        return commands.stream()
+                .filter(command -> commandName.equals(command.getFirstKey()))
+                .toList();
+    }
+
+    /**
+     * Asserts that exactly one command named {@code commandName} was sent, and that it carries every field of
+     * {@code expected}. A subset check rather than equality, because the driver adds session and cluster metadata
+     * ({@code lsid}, {@code $db}, {@code txnNumber}, and the like) that {@code expected} does not name.
+     */
+    private static void assertOneCommandNamed(List<BsonDocument> commands, String commandName, String expected) {
+        var matching = commandsNamed(commands, commandName);
+        assertThat(matching).hasSize(1);
+        assertThat(matching.get(0))
+                .asInstanceOf(InstanceOfAssertFactories.MAP)
+                .containsAllEntriesOf(BsonDocument.parse(expected));
+    }
+
+    @Entity(name = "Book")
+    @Table(name = "books")
+    static class Book {
+        @Id
+        @GeneratedValue
+        Long id;
+
+        String title;
+    }
+
+    /** A primitive identifier, to pin which class {@code getReturnedClass()} reports for one. */
+    @Entity(name = "Pamphlet")
+    @Table(name = "pamphlets")
+    static class Pamphlet {
+        @Id
+        @GeneratedValue
+        long id;
+
+        String title;
+    }
+
+    @Entity(name = "Ledger")
+    @Table(name = "ledgers")
+    @SequenceGenerator(name = "ledgerSeq", allocationSize = 1)
+    static class Ledger {
+        @Id
+        @GeneratedValue(generator = "ledgerSeq")
+        Long id;
+    }
+
+    @Entity(name = "Invoice")
+    @Table(name = "invoices")
+    @SequenceGenerator(
+            name = "invoice_numbers",
+            sequenceName = "invoice_numbers",
+            initialValue = 1000,
+            allocationSize = 10)
+    static class Invoice {
+        @Id
+        @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "invoice_numbers")
+        Long id;
+    }
+
+    @Entity(name = "Receipt")
+    @Table(name = "receipts")
+    @SequenceGenerator(name = "receiptSeq", schema = "billing", sequenceName = "receipt_numbers", allocationSize = 1)
+    static class Receipt {
+        @Id
+        @GeneratedValue(generator = "receiptSeq")
+        Long id;
+    }
+
+    /**
+     * Declares a generator that {@link SharedSequenceConsumer} references without declaring it itself. Both end up on
+     * the default allocation size, which they must, since the consumer has no way to state one.
+     */
+    @Entity(name = "SharedSequenceDeclarer")
+    @Table(name = "sharedSequenceDeclarers")
+    @SequenceGenerator(name = "shared_seq", sequenceName = "shared_seq")
+    static class SharedSequenceDeclarer {
+        @Id
+        @GeneratedValue(generator = "shared_seq")
+        Long id;
+    }
+
+    @Entity(name = "SharedSequenceConsumer")
+    @Table(name = "sharedSequenceConsumers")
+    static class SharedSequenceConsumer {
+        @Id
+        @GeneratedValue(generator = "shared_seq")
+        Long id;
+    }
+
+    @Entity(name = "Note")
+    @Table(name = "notes")
+    @SequenceGenerator(name = "noteSeq", allocationSize = 1)
+    static class Note {
+        @Id
+        @GeneratedValue(generator = "noteSeq")
+        Integer id;
+    }
+
+    @Test
+    void persistsAndReloadsWithAGeneratedIdentifier() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var book = new Book();
+                    book.title = "War and Peace";
+                    session.persist(book);
+                    session.getTransaction().commit();
+                    session.clear();
+                    var reloaded = session.find(Book.class, book.id);
+                    assertThat(reloaded.title).isEqualTo("War and Peace");
+                    // Read while the SessionFactory is still open: create-drop deletes this document when it closes.
+                    var sequenceDocument = sequences
+                            .find(BsonDocument.parse("{\"_id\": \"books_SEQ\"}"))
+                            .first();
+                    return new PersistedBook(book.id, reloaded.id, sequenceDocument);
+                },
+                Book.class);
+
+        assertThat(run.observed().id()).isEqualTo(1L);
+        assertThat(run.observed().reloadedId()).isEqualTo(1L);
+        assertThat(run.observed().sequenceDocument())
+                .isEqualTo(
+                        BsonDocument.parse(
+                                """
+                                {
+                                  "_id": "books_SEQ",
+                                  "next_value": {"$numberLong": "51"},
+                                  "increment": {"$numberLong": "50"}
+                                }"""));
+
+        assertOneCommandNamed(
+                run.commands(),
+                "update",
+                """
+                {
+                  "update": "hibernate_sequences",
+                  "updates": [
+                    {
+                      "q": {"_id": "books_SEQ"},
+                      "u": {
+                        "$setOnInsert": {
+                          "next_value": {"$numberLong": "1"},
+                          "increment": {"$numberLong": "50"}
+                        }
+                      },
+                      "upsert": true
+                    }
+                  ]
+                }""");
+        assertOneCommandNamed(
+                run.commands(),
+                "findAndModify",
+                """
+                {
+                  "findAndModify": "hibernate_sequences",
+                  "query": {
+                    "_id": "books_SEQ",
+                    "next_value": {"$type": "long"},
+                    "increment": {"$type": "long"}
+                  },
+                  "update": [
+                    {"$set": {"next_value": {"$add": ["$next_value", "$increment"]}}}
+                  ],
+                  "new": false,
+                  "fields": {"_id": 0, "next_value": 1}
+                }""");
+        assertThat(commandsNamed(run.commands(), "findAndModify").get(0)).doesNotContainKey("nonTransactional");
+
+        // Two, not one: the schema tool's "create" action does a defensive drop before creating, and "create-drop"
+        // drops again at teardown, so the sequence's drop command is sent twice with identical content.
+        var deleteCommands = commandsNamed(run.commands(), "delete");
+        assertThat(deleteCommands).hasSize(2);
+        var expectedDelete = BsonDocument.parse(
+                """
+                {
+                  "delete": "hibernate_sequences",
+                  "deletes": [
+                    {"q": {"_id": "books_SEQ"}, "limit": 1}
+                  ]
+                }""");
+        assertThat(deleteCommands).allSatisfy(command -> assertThat(command)
+                .asInstanceOf(InstanceOfAssertFactories.MAP)
+                .containsAllEntriesOf(expectedDelete));
+    }
+
+    @Test
+    void persistsWithAPrimitiveGeneratedIdentifier() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var pamphlet = new Pamphlet();
+                    pamphlet.title = "On Liberty";
+                    session.persist(pamphlet);
+                    session.getTransaction().commit();
+                    return pamphlet.id;
+                },
+                Pamphlet.class);
+
+        assertThat(run.observed()).isEqualTo(1L);
+    }
+
+    @Test
+    void persistsWithAnIntegerGeneratedIdentifier() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var note = new Note();
+                    session.persist(note);
+                    session.getTransaction().commit();
+                    return note.id;
+                },
+                Note.class);
+
+        assertThat(run.observed()).isEqualTo(1);
+    }
+
+    @Test
+    void allocationSizeOneAllocatesPerRow() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var ids = new ArrayList<Long>();
+                    for (var i = 0; i < 3; i++) {
+                        var ledger = new Ledger();
+                        session.persist(ledger);
+                        ids.add(ledger.id);
+                    }
+                    session.getTransaction().commit();
+                    return ids;
+                },
+                Ledger.class);
+
+        assertThat(run.observed()).containsExactly(1L, 2L, 3L);
+        assertThat(commandsNamed(run.commands(), "findAndModify")).hasSize(3);
+    }
+
+    @Test
+    void pooledAllocationAmortizesTheRoundTrip() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var ids = new ArrayList<Long>();
+                    for (var i = 0; i < 51; i++) {
+                        var book = new Book();
+                        book.title = "Volume " + i;
+                        session.persist(book);
+                        ids.add(book.id);
+                    }
+                    session.getTransaction().commit();
+                    return ids;
+                },
+                Book.class);
+
+        assertThat(run.observed())
+                .isEqualTo(LongStream.rangeClosed(1, 51).boxed().toList());
+        assertThat(commandsNamed(run.commands(), "findAndModify")).hasSize(2);
+    }
+
+    /**
+     * {@link #allocationSizeOneAllocatesPerRow()} and {@link #pooledAllocationAmortizesTheRoundTrip()} assert exact
+     * identifiers for {@code none} and {@code pooled}. These three produce different numbers: {@code hilo} multiplies
+     * the stored value by the increment. This test checks only that identifiers are distinct and increasing.
+     */
+    @ParameterizedTest
+    @ValueSource(strings = {"pooled-lo", "hilo", "legacy-hilo"})
+    void optimizersYieldDistinctIncreasingIdentifiers(String optimizer) {
+        var run = inRegistry(
+                Map.of("hibernate.id.optimizer.pooled.preferred", optimizer),
+                session -> {
+                    session.getTransaction().begin();
+                    var ids = new ArrayList<Long>();
+                    for (var i = 0; i < 51; i++) {
+                        var book = new Book();
+                        book.title = "Volume " + i;
+                        session.persist(book);
+                        ids.add(book.id);
+                    }
+                    session.getTransaction().commit();
+                    return ids;
+                },
+                Book.class);
+
+        assertThat(run.observed()).doesNotHaveDuplicates().isSorted();
+    }
+
+    /**
+     * JPA generator names are global, so an entity may reference a {@link SequenceGenerator} declared on a different
+     * entity. {@code forbidUnintrospectableGenerator} rejects a {@code generator()} name it cannot resolve, and this is
+     * the only shape that reaches its global-registration lookup rather than the member or class lookup.
+     */
+    @Test
+    void aGeneratorNamedOnAnotherEntityIsAccepted() {
+        var run = inRegistry(
+                session -> {
+                    session.getTransaction().begin();
+                    var consumer = new SharedSequenceConsumer();
+                    session.persist(consumer);
+                    session.getTransaction().commit();
+                    return consumer.id;
+                },
+                SharedSequenceDeclarer.class,
+                SharedSequenceConsumer.class);
+
+        assertThat(run.observed()).isEqualTo(1L);
+    }
+
+    @Test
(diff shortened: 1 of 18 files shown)


src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
```diff
@@ -249,13 +302,19 @@ public void clearWarnings() throws SQLException {
...
+            executeUpdate(parsedCommand);
+        } else {
+            var command = AdminCommand.toAdminCommand(mql);
+            try {
+                command.execute(mongoDatabase);
+            } catch (RuntimeException exception) {
+                throw handleExecuteQueryOrUpdateException(exception);
+            }
         }
+        return false;
```
vbabanin, line 317: Just a thought for sometime later, not something to consider in this PR.

The `find() != null && isUpdate()` stitching feels like a sign that we have two parallel representations of the same command concept.

Longer-term, i think this could be modeled as a single `Command` abstraction where each command owns both parsing and execution. For example, `Insert`, `Update`, `Delete`, `Aggregate`, `FindAndModify`, `CreateCollection`, `CreateIndexes`, `Drop`, etc. could each parse their MQL shape and execute themselves.

Then the call site becomes closer to:

```java
parse(mql).execute(...);
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,940 @@
...
+
+    /** What one booted {@code SessionFactory} produced: the commands sent, and whatever the body observed. */
+    private record Run<T>(List<BsonDocument> commands, T observed) {}
+
+    /** The identifiers a persist-then-reload round trip observed, and the counter document at that point. */
+    private record PersistedBook(long id, long reloadedId, BsonDocument sequenceDocument) {}
+
+    /**
+     * Boots a {@code SessionFactory} for {@code entityClasses} with {@code create-drop}, runs {@code body} while it is
+     * open, and returns the commands sent along with the body's result.
```
vbabanin, line 98: [nit]
```suggestion
     * open, and returns the commands sent along with the {@code body}'s result.
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,940 @@
...
+     *
+     * <p>The registry is built by hand so it can apply the contributor that points the {@code SessionFactory} at this
```
vbabanin, line 100: [nit]
```suggestion
     * <p>The registry is built programmatically so it can apply the contributor that points the {@code SessionFactory} at this
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,940 @@
...
+     * class's own database and installs that database's command listener.
+     */
+    private <T> Run<T> inRegistry(Function<Session, T> body, Class<?>... entityClasses) {
+        return inRegistry(Map.of(), body, entityClasses);
+    }
```
vbabanin, line 105: I think the naming is getting a bit hard to follow here.

There are two similar patterns in `IndexIntegrationTests` and `SequenceGeneratedIdIntegrationTests`, but they use different names for the same concepts:

* `IndexIntegrationTests` uses `observer` / `observed`
* `SequenceGeneratedIdIntegrationTests` uses `body`
* This PR also introduces `Run`, which seems identical to the existing `Export.

I think `body` is clearer than `observer`, but this whole helper shape seems similar to Hibernate’s test helpers like `inTransaction(...)` / `inSession(...)`, where the parameter is usually named `action`.

Could we move the duplicated `InRegistry` logic from `IndexIntegrationTests` and `SequenceGeneratedIdIntegrationTests` into a shared test helper and align the naming there?

For example:
* lambda parameter: `action`
* returned/captured value: `value`
* wrapper type: maybe `ActionResult` / `Result` instead of `Run` / `Export`

The helper could take the final merged `Map<String, Object> settings` and stay agnostic about what is “base” vs test-specific. Each test class can keep its own `BASE_SETTINGS`, merge any per-test overrides, and pass the final settings map into the helper.


For example:

```
      var result = inServiceRegistry(
              getClass(),
              withSetting(Map.of()),
              session -> {
                  session.getTransaction().begin();
                  var book = new Book();
                  book.title = "War and Peace";
                  session.persist(book);
                  session.getTransaction().commit();
                  session.clear();
                  var reloaded = session.find(Book.class, book.id);
                  var seq = sequences.find(BsonDocument.parse("{\"_id\": \"books_SEQ\"}")).first();
                  return new PersistedBook(book.id, reloaded.id, seq);
              },
              Book.class);
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
vbabanin, line 100: [nit]
```suggestion
     * <p>The {@link ServiceRegistry} is built by hand so it can apply the contributor that points the {@code SessionFactory} at this
```

jyemin: Done

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
jyemin: Yeah, it's gotten a bit hard to understand.

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
jyemin: Yeah, good idea to DRY it up.  As for the names:

* I chose MongoRegistry for the location of the helper method.  RegistryHelper would also work.
* I used `action` as the name
* I kept `observed` and `Run`.  `ActionResult` didn't read write since it contains more than the result, but I'm not stuck on it.
* I moved the base settings into `MongoRegistry` as a named constant.  We can add more of them as needed.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -381,10 +401,193 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+                    persistentClass, identifierType.getTypeName()));
+        }
+        throw new FeatureNotSupportedException(format(
+                "%s: a generated identifier of type [%s] is not supported;"
+                        + " supported types are [short], [int] and [long] and their boxed forms",
+                persistentClass, identifierType.getTypeName()));
+    }
+
+    private static boolean isSupportedGeneratedIdentifierType(Class<?> identifierType) {
+        return identifierType == Short.class || identifierType == Integer.class || identifierType == Long.class;
```
vbabanin: `Short` generated identifiers are currently declared as supported, so Hibernate boots successfully, but they fail at runtime:

```text
JDBCException: Unable to bind parameter #2 - 1 [setShort not implemented]
```

We decided not to support `Short` when implementing CRUD operations, so this is currently inconsistent with the rest of the implementation.

I think we should drop `Short` support from sequences to keep this aligned. What do you think?

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -381,10 +401,193 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+        }
+    }
+
+    /**
+     * Any name that is neither a legacy non-sequence name nor a globally registered {@link GenericGenerator} either
+     * matches a declared {@link SequenceGenerator} or becomes an implicit sequence, and both are supported. A declared
+     * {@link SequenceGenerator} is not looked up here because generator names are global: an entity may name one
+     * declared on a different entity.
+     */
+    private static boolean namesANonSequenceGenerator(String generatorName, InFlightMetadataCollector metadata) {
```
vbabanin: It appears to be the only boolean helper using an article in the name. Existing helpers follow the `isX... pattern`, for example:

```
isCompactTupleInList      
isFieldPathExpression
isMongoDialect                                                                                                                                                                                                                                                                                      
isComponent               
isFieldValueComparison
isParameterMarker                                                                                                                                                                                                                                                                                       
isExcludeProjectSpecification  
isFieldValuePair
isStructOrArrayType                                                                                                                                                                                                                                                                                                   
isJoinedColumn           
isSupportedGeneratedIdentifierType                                                                                                                                                                                                                                                                                      
isTestMongoDialectSubclass 
isTimeoutException       
isUsed                                                                                                                                                                                                                                                                                           
isValueExpression
isUpdate 
isQuery
```
Since this is a boolean predicate, could we rename it to `isNonSequenceGeneratorName`?

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.boot;
+
+import static java.lang.String.format;
+
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
+
+final class NameChecks {
```
vbabanin, line 25: Let's apply `@hidden` and `@SuppressWarnings("MissingSummary")` for this class well.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -381,10 +401,193 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+            return;
+        }
+        // getDirectAnnotationUsage returns null when the annotation is absent, though it is declared without
+        // @Nullable, so an IDE reports this check as always false. Removing it would run the switch below on every
+        // entity, including the ones with no @GeneratedValue at all.
+        var generatedValue = memberDetails.getDirectAnnotationUsage(GeneratedValue.class);
+        if (generatedValue == null) {
+            return;
+        }
+        forbidUnintrospectableGenerator(persistentClass, memberDetails, generatedValue, metadata);
```
vbabanin: Placing a generator on a `@MappedSuperclass` is a valid pattern (based on section 3.7.9, Example 3.187 of [Hibernate user guide](https://docs.hibernate.org/orm/7.4/userguide/html_single/#settings-compliance)).

`forbidUnintrospectableGenerator` currently resolves the entity class via `persistentClass.getClassName()` and checks annotations on that class and on the `@Id` member. When both the identifier and generator annotation are declared on a `@MappedSuperclass`, this misses the generator: the entity class is not the declaring class, and the concrete entity member has no generator annotation.

In this case, boot succeeds, Hibernate ORM resolves a real `org.hibernate.id.enhanced.TableGenerator` from the superclass annotation, and the first insert fails later in the JDBC layer:

```text
JDBCException: Error performing isolated work
  [Invalid MQL: [select next_val from super_table_gen
                 where sequence_name={"$undefined": true} for update]]
```
instead of failing early with the intended `FeatureNotSupportedException` for `[TABLE]` generators.

Reproducer:

```java
@MappedSuperclass
@TableGenerator(name = "superTableGen", table = "super_table_gen")
abstract static class TableGenSuper {
    @Id
    @GeneratedValue
    Long id;
}

@Entity(name = "SuperTableGenItem")
@Table(name = "superTableGenItems")
static class SuperTableGenItem extends TableGenSuper {
    String title;
}

   @Test
    void sequenceGeneratorOnMappedSuperclass() {
        var id = inRegistry(
                session -> {
                    session.getTransaction().begin();
                    var item = new SuperSequenceGenItem();
                    item.title = "Handbill";
                    session.persist(item);
                    session.getTransaction().commit();
                    return item.id;
                },
                SuperSequenceGenItem.class);

        assertThat(id).isEqualTo(1L);
    }
```

`@GenericGenerator(strategy = "org.hibernate.id.enhanced.TableGenerator")` in the same position fails the same way.

I think we should walk the mapped-superclass hierarchy starting from the identifier member’s declaring type. That should also help catch other unsupported features declared on mapped superclasses in the future.

What do you think?

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
jyemin: Isn't it one or the other?  I also thought it is not necessary for a non-public class.

jyemin: Added `@SuppressWarnings("MissingSummary")`

vbabanin: Oh yes, i have not noticed it is `package-private`. "hidden" is not needed -  resolved.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Yes, agreed.  Nice find.

Done.

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
jyemin: And `@hidden`.  I see there is precedent for it.

jyemin: So roll it back?  Or leave it?

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -381,10 +403,196 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+        }
+        if (identifierType == BigInteger.class || identifierType == BigDecimal.class) {
+            throw new FeatureNotSupportedException(format(
+                    "%s: a generated identifier of type [%s] is not supported."
+                            + " TODO-HIBERNATE-253 https://jira.mongodb.org/browse/HIBERNATE-253",
+                    persistentClass, identifierType.getTypeName()));
+        }
+        throw new FeatureNotSupportedException(format(
+                "%s: a generated identifier of type [%s] is not supported;"
+                        + " supported types are [short], [int] and [long] and their boxed forms",
```
vbabanin: This message appears to be stale now that we no longer support `Short` / `short`.

```suggestion
                        + " supported types are [int] and [long] and their boxed forms",
```

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
```diff
@@ -0,0 +1,44 @@
...
+
+package com.mongodb.hibernate.internal.boot;
+
+import static java.lang.String.format;
+
+import com.mongodb.hibernate.internal.FeatureNotSupportedException;
+
+/** @hidden */
+@SuppressWarnings("MissingSummary")
+final class NameChecks {
```
vbabanin, line 25: Should we either move the `#` / `$` / `.` field-name checks into `NameChecks` for coherence? Right now `NameChecks` only covers one check.

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * A schema folds into the name a collection or a sequence counter is stored under, as {@code schema.name}, so the
+     * '.' is the extension's own separator. A '.' written by the user makes the folded name ambiguous:
+     * {@code @Table(schema = "a", name = "b")} and {@code @Table(name = "a.b")} would resolve to the same collection,
+     * and nothing downstream could tell the two qualifiers apart.
+     */
+    static void forbidDot(String name, String kind) {
```
vbabanin: The class/method naming makes this look like a general name-validation utility, but `forbidDot` is safe to use only for schema/collection names.

Should we reflect that in the method name, e.g. `forbidDotInSchemaName` / `forbidDotInCollectionName`, so it is clear this should not be reused for field.

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
```diff
@@ -0,0 +1,44 @@
...
+        if (name.contains(".")) {
+            throw new FeatureNotSupportedException(format(
+                    "The character [.] in a %s name is not supported, but is present in [%s]. A schema folds into the"
+                            + " name as [schema.name], so a '.' written by the user would make the folded name"
```
vbabanin:
```suggestion
                    "The character [.] in a %s name is not supported, but is present in [%s]. A schema and a name are"
                            + " joined with a '.' as [schema.name], so a '.' in a name collides with that join.",
```

src/main/java/com/mongodb/hibernate/internal/dialect/MongoSequenceSupport.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Returns the pre-image, which is the value being handed out, and narrows the reply to one field because
+     * {@code SequenceStructure} reads column 1 positionally.
+     *
+     * <p>Carries {@link com.mongodb.hibernate.internal.MongoConstants#NON_TRANSACTIONAL_COMMAND_FIELD_NAME} because a
+     * native SQL sequence advances outside the caller's transaction and {@code SequenceStructure} relies on that:
+     * joining the transaction would make concurrent allocations conflict and let a rollback rewind the counter.
+     */
+    @Override
```
vbabanin: Maybe just STring?

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+                        {
+                          "_id": "books_SEQ",
+                          "next_value": {"$numberLong": "501"},
+                          "increment": {"$numberLong": "50"}
+                        }"""));
+
+        // The JPA action "create" (unlike "create-drop") does not drop before creating, so re-export is the only
+        // thing under test here: no teardown-drop to muddy whether the counter merely survived being untouched.
+        var run = inRegistry(
+                Book.class, Map.of("jakarta.persistence.schema-generation.database.action", "create"), session -> {
```
vbabanin, line 486: I think we should also add `Unsupported` tests for the `update` and `drop` schema-generation actions.

`update` currently fails with an error:

```text
org.hibernate.JDBCException: Error accessing table metadata [getTables not implemented] [n/a]
```

Example test shape:
```java
@Test
void updateSchemaActionIsRejectedWithFeatureNotSupportedException() {
    books.insertOne(BsonDocument.parse("{\"_id\": 500, \"title\": \"pre-existing\"}"));

    assertThatThrownBy(() -> inRegistry(
                    Map.of("jakarta.persistence.schema-generation.database.action", "update"),
                    session -> {
                        session.getTransaction().begin();
                        var book = new Book();
                        book.title = "After resync";
                        session.persist(book);
                        session.getTransaction().commit();
                        return book.id;
                    },
                    Book.class))
            .isInstanceOf(FeatureNotSupportedException.class);
}
```

Should we reject it with `FeatureNotSupportedException`?

Related comment: https://github.com/mongodb/mongo-hibernate/pull/227/changes#r4043038277

src/main/java/com/mongodb/hibernate/internal/dialect/MongoSequenceSupport.java
```diff
@@ -0,0 +1,166 @@
...
+                true);
+    }
+
+    @Override
+    public String getRestartSequenceString(String sequenceName, long startWith) {
+        return updateCommand(
+                sequenceName,
+                new BsonDocument("$set", new BsonDocument(NEXT_VALUE_FIELD_NAME, new BsonInt64(startWith))),
+                false);
+    }
```
vbabanin, line 122: This method seems to be never executed because `"jakarta.persistence.schema-generation.database.action", "update"` throws in getTableSchemas() of `MongoPreparedStatement`.

We should forbid update here with FeatureUnsupportedException or implement support in `getTables`. User guide explicitly calls out as "not suitable for a production environment" - use Flyway/Liquibase instead, i would lean towards not supporting it.

Ideally, we should intercept `jakarta.persistence.schema-generation.database.action", "update"` configuration and throw so we forbid it across the board, i assume Indexes might be affected as well.


Related comment: https://github.com/mongodb/mongo-hibernate/pull/227/changes#r4042696651

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -381,10 +403,196 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+    }
+
+    /**
+     * Rejects the annotation combinations that resolve, via Hibernate's own generator lookup, to a generator other than
+     * a MongoDB-backed sequence: a {@link TableGenerator} or {@link GenericGenerator} localized on the identifier
+     * member or on the entity class or one of its superclasses (a {@code @MappedSuperclass} can carry the generator), a
+     * meta-annotation annotated with {@link IdGeneratorType}, and a non-blank {@code generator()} that names none of
+     * {@code @SequenceGenerator} localized on the member/class or declared globally in the metadata.
+     */
+    private static void forbidUnintrospectableGenerator(
```
vbabanin: A `@TableGenerator` declared on one entity and referenced by name from another entity’s `@GeneratedValue` can currently escape boot validation.

There seem to be two gaps:

1. `forbidUnsupportedIdentifierGeneration` returns early when an entity’s id has no `@GeneratedValue`, so the declaring entity is never checked if its own id is a plain `@Id`.
2. `isNonSequenceGenerator` checks `getGenericGeneratorRegistrations()`, but not `getTableGeneratorRegistrations()`, so the consumer-side `generator = "..."` reference to a `TableGenerator` is not caught either.

Hibernate then resolves the global `TableGenerator`, emits `select ... for update`, and the JDBC adapter fails to parse it at first insert.

Reproducer:

```java
@Entity
@Table(name = "declarers")
@TableGenerator(name = "globalTableGen", table = "global_table_gen", allocationSize = 1)
static class Declarer {
    @Id Long id; // no @GeneratedValue -> forbidUnsupportedIdentifierGeneration returns early
    String title;
}

@Entity
@Table(name = "consumers")
static class Consumer {
    @Id
    @GeneratedValue(generator = "globalTableGen")
    Long id;
}
```

with `hibernate.jpa.compliance.global_id_generators=true`.

Boot succeeds, and the first `persist(new Consumer())` fails with:

```text
org.hibernate.JDBCException: Error performing isolated work
  [Invalid MQL: [select next_val from global_table_gen where sequence_name={"$undefined": true} for update]]
```

I think we have two possible fixes:

1. Check `getTableGeneratorRegistrations()` in `isNonSequenceGenerator`. It returns a `Map<String, TableGeneratorRegistration>` keyed by generator name. If the `@GeneratedValue(generator = "...")` name is present in that map, it resolves to a `TableGenerator` - potentially a global one - so we should reject it during boot.

2. Validate unsupported generator declarations (`@TableGenerator` / `@GenericGenerator` / `@IdGeneratorType`) regardless of whether the declaring entity’s own id uses them. These declarations register into the global generator namespace and can be referenced by another entity’s `@GeneratedValue`, so the declaration check should not be gated behind:

```java
var generatedValue = memberDetails.getDirectAnnotationUsage(GeneratedValue.class);
if (generatedValue == null) {
    return;
}
```

I slightly prefer option 2 because it blocks unsupported generator declarations at the source and avoids adding more name-resolution logic.

What do you think?

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+
+        @Test
+        void sequenceOptionsAreRejected() {
+            assertThatThrownBy(() -> inRegistry(OptionsItem.class, session -> null))
+                    .isInstanceOf(FeatureNotSupportedException.class)
+                    .hasMessageContaining("cache 20");
+        }
+
+        @Test
+        void entityMappedToTheSequenceCollectionIsRejected() {
```
vbabanin:
```suggestion
        void entityMappedToSequenceCollectionIsRejected() {
```

src/test/java/com/mongodb/hibernate/internal/dialect/MongoSequenceSupportTests.java
```diff
@@ -0,0 +1,163 @@
...
+                                    {
+                                      "q": {"_id": "books_SEQ"},
+                                      "u": {"$set": {"next_value": {"$numberLong": "500"}}}
+                                    }
+                                  ]
+                                }"""));
+    }
+
+    @Test
+    void dropRemovesOnlyTheSequenceDocument() {
```
vbabanin:
```suggestion
    void dropRemovesOnlySequenceDocument() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+                    {"q": {"_id": "books_SEQ"}, "limit": 1}
+                  ]
+                }""");
+        assertThat(deleteCommands).allSatisfy(command -> assertThat(command)
+                .asInstanceOf(InstanceOfAssertFactories.MAP)
+                .containsAllEntriesOf(expectedDelete));
+    }
+
+    @Test
+    void persistsWithAPrimitiveGeneratedIdentifier() {
```
vbabanin:
```suggestion
    void persistsWithPrimitiveGeneratedIdentifier() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+                });
+
+        // The pooled optimizer reads the returned value as the upper bound of the block it just reserved, the same
+        // as a native sequence would: the fresh optimizer's first identifier is next_value - increment + 1 = 452,
+        // not 501.
+        assertThat(run.observed()).isEqualTo(452L);
+    }
+
+    @Test
+    void allocationSurvivesARolledBackTransaction() {
```
vbabanin:
```suggestion
    void allocationSurvivesRolledBackTransaction() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+                        Map.of("jakarta.persistence.schema-generation.database.action", "none"),
+                        session -> null))
+                .isInstanceOf(MappingException.class)
+                .hasMessage(
+                        "The increment size of the [books_SEQ] sequence is set to [50] in the entity mapping but the"
+                                + " mapped database sequence increment size is [1]");
+    }
+
+    @Test
+    void allocationWithoutASeededCounterFails() {
```
vbabanin:
```suggestion
    void allocationWithoutSeededCounterFails() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+            return sequences
+                    .find(BsonDocument.parse("{\"_id\": \"billing.receipt_numbers\"}"))
+                    .first();
+        });
+
+        assertThat(run.observed()).isNotNull();
+    }
+
+    @Test
+    void reExportDoesNotResetALiveSequence() {
```
vbabanin:
```suggestion
    void reExportDoesNotResetLiveSequence() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+                    return consumer.id;
+                },
+                SharedSequenceDeclarer.class,
+                SharedSequenceConsumer.class);
+
+        assertThat(run.observed()).isEqualTo(1L);
+    }
+
+    @Test
+    void persistsWithASequenceDeclaredOnAMappedSuperclass() {
```
vbabanin:
```suggestion
    void persistsWithASequenceDeclaredOnMappedSuperclass() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+    }
+
+    @Entity(name = "SequenceSuperItem")
+    @Table(name = "sequenceSuperItems")
+    static class SequenceSuperItem extends SequenceSuper {
+        String title;
+    }
+
+    @Test
+    void persistsAndReloadsWithAGeneratedIdentifier() {
```
vbabanin: Nit: could we rename this to `persistsAndReloadsWithGeneratedIdentifier`? The article 'A' does not add much here, and the shorter form is more consistent with the existing names in the codebase.

```suggestion
    void persistsAndReloadsWithGeneratedIdentifier() {
```

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
```diff
@@ -0,0 +1,975 @@
...
+     *
+     * <p>Each shape is caught in a different place, which is why each asserts its own message. A missing
+     * {@code increment} reads back as {@code 0} during Hibernate ORM's boot-time increment-mismatch check and fails
+     * there. A wrong BSON type fails in the same check, while reading the value. A missing {@code next_value} leaves
+     * {@code increment} matching {@code allocationSize}, so boot succeeds and the allocation's own shape guard catches
+     * it instead.
+     */
+    @ParameterizedTest
+    @MethodSource("malformedCounterDocuments")
+    void allocationFromAMalformedCounterFails(String counterDocument, String expectedMessage) {
```
vbabanin:
```suggestion
    void allocationFromMalformedCounterFails(String counterDocument, String expectedMessage) {
```

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
vbabanin: The error message reads a bit awkwardly around "written by the user." Could we make it input-focused, e.g. “an additional `.` in the name” or “a `.` in the provided name”?

That'd keep the message focused on what is invalid rather than who provided it. What do you think?

```suggestion
        if (name.contains(".")) {
            throw new FeatureNotSupportedException(format(
                    "The character [.] in a %s name is not supported, but is present in [%s]. A schema folds into the"
                            + " name as [schema.name], so a '.' in the provided name would make the folded name"
```

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
jyemin: I'd prefer to leave it alone.  The only reason to break this one out is because it's used from two classes.

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
jyemin: I think we should, but this error is pre-existing not introduced by this PR.  I would prefer a separate ticket for it.  WDYT?

src/main/java/com/mongodb/hibernate/internal/boot/NameChecks.java
jyemin: Went with your suggestion in Slack of `forbidDotInQualifiedName`

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: I also like Option 2.

src/main/java/com/mongodb/hibernate/internal/dialect/MongoSequenceSupport.java
jyemin: Fair enough.  Replaced with `FeatureUnsupportedException`

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/jdbc/MongoStatement.java
jyemin: As agreed, I think, deferring.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Done

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
jyemin: Done

src/integrationTest/java/com/mongodb/hibernate/id/SequenceGeneratedIdIntegrationTests.java
vbabanin: Yeah, agreed. I think this is also resolved by this comment: https://github.com/mongodb/mongo-hibernate/pull/227#discussion_r4043038277.

src/main/java/com/mongodb/hibernate/internal/boot/MongoAdditionalMappingContributor.java
```diff
@@ -384,10 +406,202 @@ private static void forbidColumnFragmentAnnotations(PersistentClass persistentCl
...
+                .forSelfAndEachSuper(classDetails ->
+                        forbidUnsupportedGeneratorAnnotations(classDetails, persistentClass, modelsContext));
+    }
+
+    /**
+     * Rejects a non-blank {@code generator()} that resolves, via Hibernate's own generator lookup, to a generator other
+     * than a MongoDB-backed sequence: a legacy non-sequence name, or a globally registered {@link GenericGenerator}. A
+     * declared {@code @SequenceGenerator} is not looked up here because generator names are global: an entity may name
+     * one declared on a different entity.
+     */
```
vbabanin, line 493:
```suggestion
    /**
     * Rejects a non-blank {@code generator()} that resolves, via Hibernate's own generator lookup, to a generator other
     * than a MongoDB-backed sequence: a legacy non-sequence name, or a globally registered {code @GenericGenerator}. 
     * A {@link @SequenceGenerator} is not looked up here; invalid {@code @SequenceGenerator} declarations are already 
     * rejected in {@link #forbidUnsupportedGeneratorDeclarations}.
     */
```

vbabanin (approved):

strogiyotec (approved):
approved without the review to bypass the permissions

PR #229, opened 2026-09-08
HIBERNATE-239

strogiyotec:
HIBERNATE-239, supporting

- count() (including count(*))
- avg()
- min()
- max()
- sum()

Statistical aggregates are not supported and tracked at HIBERNATE-257

### DIFF
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
@@ -1541,6 +1541,108 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
         }
 
         @Test
+        void orderByAggregateAlias() {
+            assertSelectionQuery(
+                    "select b.string, sum(b.primitiveInt) as total from Item as b GROUP BY b.string"
+                            + " ORDER BY total DESC, b.string",
+                    Object[].class,
+                    """
+                    {
+                      "aggregate": "Item",
+                      "pipeline": [
+                        {
+                          "$group": {
+                            "_id": {"string": "$string"},
+                            "#acc_0": {"$sum": {"$toLong": "$primitiveInt"}}
+                          }
+                        },
+                        {"$sort": {"#acc_0": -1, "_id.string": 1}},
+                        {"$project": {"_id#string": "$_id.string", "total": "$#acc_0", "_id": 0}}
+                      ]
+                    }
+                    """,
+                    results -> assertThat((Iterable<Object[]>) results)
+                            .containsExactly(new Object[] {"c", 7L}, new Object[] {"a", 4L}, new Object[] {"b", 4L}),
+                    Set.of(COLLECTION_NAME));
+        }
+
+        @Test
+        void orderByAggregateOrdinal() {
+            assertSelectionQuery(
+                    "select b.string, sum(b.primitiveInt) from Item as b GROUP BY b.string ORDER BY 2 DESC, 1",
+                    Object[].class,
+                    """
+                    {
+                      "aggregate": "Item",
+                      "pipeline": [
+                        {
+                          "$group": {
+                            "_id": {"string": "$string"},
+                            "#acc_0": {"$sum": {"$toLong": "$primitiveInt"}}
+                          }
+                        },
+                        {"$sort": {"#acc_0": -1, "_id.string": 1}},
+                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                      ]
+                    }
+                    """,
+                    results -> assertThat((Iterable<Object[]>) results)
+                            .containsExactly(new Object[] {"c", 7L}, new Object[] {"a", 4L}, new Object[] {"b", 4L}),
+                    Set.of(COLLECTION_NAME));
+        }
+
+        /**
+         * An alias naming an aggregate that is already in SELECT must resolve to the accumulator SELECT registered, not
+         * add a second one.
+         */
+        @Test
+        void orderByAggregateAliasSharesTheSelectAccumulator() {
+            assertSelectionQuery(
+                    "select b.string, sum(b.primitiveInt) as total from Item as b GROUP BY b.string"
+                            + " HAVING sum(b.primitiveInt) > 4 ORDER BY total",
+                    Object[].class,
+                    """
+                    {
+                      "aggregate": "Item",
+                      "pipeline": [
+                        {
+                          "$group": {
+                            "_id": {"string": "$string"},
+                            "#acc_0": {"$sum": {"$toLong": "$primitiveInt"}}
+                          }
+                        },
+                        {"$match": {"#acc_0": {"$gt": 4}}},
+                        {"$sort": {"#acc_0": 1}},
+                        {"$project": {"_id#string": "$_id.string", "total": "$#acc_0", "_id": 0}}
+                      ]
+                    }
+                    """,
+                    results -> assertThat((Iterable<Object[]>) results).containsExactly(new Object[] {"c", 7L}),
+                    Set.of(COLLECTION_NAME));
+        }
+
+        /**
+         * Ordering by an expression is unsupported however it is named, but it must now fail with a ticket rather than
+         * with the message-less exception an unresolved alias used to produce.
+         */
+        @ParameterizedTest(name = "[{index}] {0}")
+        @MethodSource("orderByExpressionQueries")
+        void orderByExpressionIsRejected(String hql) {
+            assertSelectQueryFailure(hql, Object[].class, FeatureNotSupportedException.class, "TODO-HIBERNATE-251");
+        }
+
+        static Stream<String> orderByExpressionQueries() {
+            return Stream.of(
+                    // written out
+                    "select b.string, avg(b.primitiveInt) from Item as b GROUP BY b.string"
+                            + " ORDER BY avg(b.primitiveInt) + 1",
+                    // by alias
+                    "select b.string, avg(b.primitiveInt) + 1 as a from Item as b GROUP BY b.string ORDER BY a",
+                    // by ordinal
+                    "select b.string, avg(b.primitiveInt) + 1 from Item as b GROUP BY b.string ORDER BY 2");
+        }
+
+        @Test
         void aggregateWithoutGroupByIsRejected() {
             assertThatThrownBy(() -> getSessionFactoryScope().inTransaction(session -> session.createSelectionQuery(
                                     "select sum(b.primitiveInt) from Item as b", Object.class)
@@ -1576,6 +1678,109 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                                     Object[].class)
                             .getResultList()))
                     .isInstanceOf(FeatureNotSupportedException.class);
+        }
+    }
+
+    /**
+     * SQL's {@code COUNT(x)} counts only the rows where {@code x} is not null, which is the whole reason the translator
+     * emits a {@code $switch} rather than a bare {@code $sum: 1}. The shared dataset has no nulls, so this nest brings
+     * its own.
+     */
+    @Nested
+    @DomainModel(annotatedClasses = {Item.class})
+    class CountNullSemantics extends AbstractQueryIntegrationTests {
+
+        @BeforeEach
+        void beforeEach() {
+            getSessionFactoryScope().inTransaction(session -> {
+                session.createMutationQuery("delete from Item").executeUpdate();
+                List.of(
+                                // primitiveInt 1: two non-null strings and one null
+                                new Item(1, 1, "x", true, new ItemStruct(1)),
+                                new Item(2, 1, null, true, new ItemStruct(1)),
+                                new Item(3, 1, "y", true, new ItemStruct(1)),
+                                // primitiveInt 2: every string null, so count(string) must be 0 while count(*) is 2
+                                new Item(4, 2, null, false, new ItemStruct(2)),
+                                new Item(5, 2, null, false, new ItemStruct(2)))
+                        .forEach(session::persist);
+            });
+        }
+
+        @Test
+        void countOfNullableColumnExcludesNulls() {
+            assertSelectionQuery(
+                    "select b.primitiveInt, count(*), count(b.string) from Item as b"
+                            + " GROUP BY b.primitiveInt ORDER BY b.primitiveInt",
+                    Object[].class,
+                    """
+                    {
+                      "aggregate": "Item",
+                      "pipeline": [
+                        {
+                          "$group": {
+                            "_id": {"primitiveInt": "$primitiveInt"},
+                            "#acc_0": {"$sum": {"$toLong": 1}},
+                            "#acc_1": {
+                              "$sum": {
+                                "$toLong": {
+                                  "$switch": {
+                                    "branches": [{"case": {"$eq": ["$string", null]}, "then": 0}],
+                                    "default": 1
+                                  }
+                                }
+                              }
+                            }
+                          }
+                        },
+                        {"$sort": {"_id.primitiveInt": 1}},
+                        {
+                          "$project": {
+                            "_id#primitiveInt": "$_id.primitiveInt",
+                            "#c_2": "$#acc_0",
+                            "#c_3": "$#acc_1",
+                            "_id": 0
+                          }
+                        }
+                      ]
+                    }
+                    """,
+                    results -> assertThat((Iterable<Object[]>) results)
+                            .containsExactly(new Object[] {1, 3L, 2L}, new Object[] {2, 2L, 0L}),
+                    Set.of(COLLECTION_NAME));
+        }
+
+        /** A HAVING on a non-null count filters on the counted rows, not on the group size. */
+        @Test
+        void havingOnCountOfNullableColumn() {
+            assertSelectionQuery(
+                    "select b.primitiveInt from Item as b GROUP BY b.primitiveInt HAVING count(b.string) > 0",
+                    Object.class,
+                    """
+                    {
+                      "aggregate": "Item",
+                      "pipeline": [
+                        {
+                          "$group": {
+                            "_id": {"primitiveInt": "$primitiveInt"},
+                            "#acc_0": {
+                              "$sum": {
+                                "$toLong": {
+                                  "$switch": {
+                                    "branches": [{"case": {"$eq": ["$string", null]}, "then": 0}],
+                                    "default": 1
+                                  }
+                                }
+                              }
+                            }
+                          }
+                        },
+                        {"$match": {"#acc_0": {"$gt": 0}}},
+                        {"$project": {"_id#primitiveInt": "$_id.primitiveInt", "_id": 0}}
+                      ]
+                    }
+                    """,
+                    List.of(1),
+                    Set.of(COLLECTION_NAME));
         }
     }
 
--- a/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
+++ b/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
@@ -25,6 +25,7 @@ import com.mongodb.hibernate.internal.FeatureNotSupportedException;
 import com.mongodb.hibernate.junit.InjectMongoCollection;
 import com.mongodb.hibernate.junit.MongoExtension;
 import com.mongodb.hibernate.query.AbstractQueryIntegrationTests;
+import com.mongodb.hibernate.query.Book;
 import jakarta.persistence.Embeddable;
 import jakarta.persistence.Embedded;
 import jakarta.persistence.Entity;
@@ -32,6 +33,7 @@ import jakarta.persistence.FetchType;
 import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
+import java.math.BigDecimal;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
@@ -647,15 +649,8 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {
-                          "$group": {
-                            "_id": {
-                              "k0": {"$switch": {"branches": [{"case": {"$gt": ["$primitiveInt", 2]}, "then": 1}], "default": 0}}
-                            },
-                            "#acc_0": {"$sum": {"$toLong": 1}}
-                          }
-                        },
-                        {"$project": {"#c_1": "$_id.k0", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$group": {"_id": {"k0": {"$switch": {"branches": [{"case": {"$gt": ["$primitiveInt", 2]}, "then": 1}], "default": 0}}}, "#acc_0": {"$sum": 1}}},
+                        {"$project": {"#c_1": "$_id.k0", "#c_2": {"$toLong": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1241,9 +1236,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": {"$toLong": 1}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": 1}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toLong": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1261,9 +1256,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": {"$toLong": {"$switch": {"branches": [{"case": {"$eq": ["$primitiveInt", null]}, "then": 0}], "default": 1}}}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": {"$switch": {"branches": [{"case": {"$eq": ["$primitiveInt", null]}, "then": 0}], "default": 1}}}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toLong": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1281,9 +1276,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": {"$toLong": "$primitiveInt"}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": "$primitiveInt"}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toLong": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1301,9 +1296,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$avg": {"$toDouble": "$primitiveInt"}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$avg": "$primitiveInt"}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toDouble": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1321,9 +1316,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$min": {"$toInt": "$primitiveInt"}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$min": "$primitiveInt"}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toInt": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1341,9 +1336,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$max": {"$toInt": "$primitiveInt"}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$max": "$primitiveInt"}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toInt": "$#acc_0"}, "_id": 0}}
                       ]
                     }
                     """,
@@ -1362,9 +1357,9 @@ public class GroupByHavingIntegrationTests extends AbstractQueryIntegrationTests
                     {
                       "aggregate": "Item",
                       "pipeline": [
-                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": {"$toLong": 1}}, "#acc_1": {"$sum": {"$toLong": "$primitiveInt"}}, "#acc_2": {"$avg": {"$toDouble": "$primitiveInt"}}, "#acc_3": {"$min": {"$toInt": "$primitiveInt"}}, "#acc_4": {"$max": {"$toInt": "$primitiveInt"}}}},
+                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": 1}, "#acc_1": {"$sum": "$primitiveInt"}, "#acc_2": {"$avg": "$primitiveInt"}, "#acc_3": {"$min": "$primitiveInt"}, "#acc_4": {"$max": "$primitiveInt"}}},
                         {"$sort": {"_id.string": 1}},
-                        {"$project": {"_id#string": "$_id.string", "#c_2": "$#acc_0", "#c_3": "$#acc_1", "#c_4": "$#acc_2", "#c_5": "$#acc_3", "#c_6": "$#acc_4", "_id": 0}}
+                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toLong": "$#acc_0"}, "#c_3": {"$toLong": "$#acc_1"}, "#c_4": {"$toDouble": "$#acc_2"}, "#c_5": {"$toInt": "$#acc_3"}, "#c_6": {"$toInt": "$#acc_4"}, "_id": 0}}
                       ]
                     }
                     """,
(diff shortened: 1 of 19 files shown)


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1263,6 +1287,14 @@ public void visitSortSpecification(SortSpecification sortSpecification) {
 
     private AstSortField createAstSortField(Expression sortExpression, AstSortOrder astSortOrder) {
         if (!isFieldPathExpression(sortExpression)) {
+            // An aggregate function is orderable under a GROUP BY: it resolves to the accumulator field $group
+            // computes, shared with SELECT/HAVING when they name the same function, so nothing is recomputed here.
+            if (sortExpression instanceof SelfRenderingFunctionSqlAstExpression<?> function) {
+                var accumulatorReference = tryRegisterAccumulator(function);
+                if (accumulatorReference != null) {
+                    return new AstSortField(accumulatorReference.fieldPath(), astSortOrder);
```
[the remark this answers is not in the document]
strogiyotec: Introduced a new method here 
https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1321

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1748,6 +1785,118 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        var one = new AstValueExpression(new AstLiteral(new BsonInt32(1)));
+        if (argument instanceof Star) {
+            return one;
+        }
+        var isNull = new AstBinaryOperatorExpression(
+                AstComparisonExpressionOperator.EQ,
+                acceptAndYieldArgument(argument),
+                new AstValueExpression(new AstLiteral(BsonNull.VALUE)));
+        return new AstSwitchExpression(
+                List.of(new AstSwitchCase(isNull, new AstValueExpression(new AstLiteral(new BsonInt32(0))))), one);
```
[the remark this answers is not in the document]
strogiyotec: Fixed and verified using these tests 
https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-561eed13d4b5c83c438633a05cf66f6f2578b2faec6e37ecb5347df1606df6c4R1691

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @see AstAccumulatorExpression
+ * @hidden
+ */
+@SuppressWarnings("MissingSummary")
+public enum AstAccumulatorOperator {
+    /**
+     * See <a href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/sum/">{@code $sum}</a>. Sum will
+     * also be used for count aggregate(E.g {"$sum": {"$toLong": 1}}) as MQL doesn't have a build in COUNT aggregate
+     */
```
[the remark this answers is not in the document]
strogiyotec: Changed the comment

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStage.java
```diff
@@ -36,22 +37,34 @@
...
+    public AstGroupStage(Collection<? extends AstGroupStageSpecification> specifications) {
+        this(specifications, List.of());
+    }
+
     @Override
     public AstStage mapChildren(AstNodeRewriter rewriter) {
-        return new AstGroupStage(specifications.stream().map(rewriter::rewrite).toList());
+        return new AstGroupStage(
+                specifications.stream().map(rewriter::rewrite).toList(),
+                accumulatorSpecifications.stream().map(rewriter::rewrite).toList());
```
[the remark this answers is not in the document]
strogiyotec: Covered this test already

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorExpression.java
```diff
@@ -0,0 +1,58 @@
...
+        implements AstExpression {
+
+    @Override
+    public StructuralKey structuralKey() {
+        return new StructuralKey("Accumulator", List.of(operator, argument));
+    }
+
+    @Override
+    public AstExpression mapChildren(AstNodeRewriter rewriter) {
+        return new AstAccumulatorExpression(operator, rewriter.rewrite(argument));
```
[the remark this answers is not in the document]
strogiyotec: Added the new test case here

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1748,6 +1785,118 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        var expressionType = expression.getExpressionType();
+        if (expressionType == null || expressionType.getJdbcTypeCount() != 1) {
+            return null;
+        }
+        return switch (expressionType.getSingleJdbcMapping().getJdbcType().getDdlTypeCode()) {
+            case SqlTypes.TINYINT, SqlTypes.SMALLINT, SqlTypes.INTEGER -> AstConversionExpressionOperator.TO_INT;
+            case SqlTypes.BIGINT -> AstConversionExpressionOperator.TO_LONG;
+            case SqlTypes.REAL, SqlTypes.FLOAT, SqlTypes.DOUBLE -> AstConversionExpressionOperator.TO_DOUBLE;
+            case SqlTypes.DECIMAL, SqlTypes.NUMERIC -> AstConversionExpressionOperator.TO_DECIMAL;
+            default -> null;
```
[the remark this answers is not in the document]
strogiyotec: SimpleSelectQueryIntegrationTests.testBigDecimal() already covers big decimal

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStage.java
```diff
@@ -36,22 +37,34 @@
...
  * </pre>
  *
+ * @param specifications the {@code _id} sub-key specifications, one per GROUP BY key; must be non-empty
+ * @param accumulatorSpecifications the accumulator specifications, rendered as siblings of {@code _id}; may be empty
  * @hidden
  */
-public record AstGroupStage(Collection<? extends AstGroupStageSpecification> specifications) implements AstStage {
+public record AstGroupStage(
+        Collection<? extends AstGroupStageSpecification> specifications,
+        Collection<? extends AstGroupStageSpecification> accumulatorSpecifications)
```
jyemin: Let's create a dedicated type for the accumulators, something that enforces the constraint that the expression must be a `AstAccumulatorExpression`, e.g.
```
record AstGroupStageAccumulatorSpecification(String key, AstAccumulatorExpression accumulator)
```

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorExpression.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.List;
+import java.util.function.Consumer;
+import org.bson.BsonWriter;
+import org.hibernate.sql.exec.spi.JdbcParameterBinder;
+
+/**
+ * A MongoDB {@code $group}-stage accumulator applied to a single argument, e.g. {@code {$sum: <expr>}}. Rendered in
+ * value position so it can appear as the right-hand side of an
+ * {@link com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstGroupStageSpecification} in the
```
jyemin: If you take the suggest on `AstGroupStage`, this sentence will need to change (or just be removed) to match.

src/test/java/com/mongodb/hibernate/internal/translate/rewrite/GroupBySubstitutionRuleTests.java
```diff
@@ -68,7 +69,7 @@ private static AstExpression add(AstExpression l, AstExpression r) {
 
     private AstRewriter rewriterWithKeys(Map<AstExpression, String> keys) {
         keys.forEach((expression, subKey) -> groupKeyVN.put(vn.valueNumber(expression), subKey));
-        return new AstRewriter(new GroupBySubstitutionRule(groupKeyVN, vn), null);
+        return new AstRewriter(new GroupBySubstitutionRule(groupKeyVN, vn, Set.of()), null);
```
jyemin: Could we add unit tests for the `accumulatorFields.contains` conditionals?

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -422,13 +427,23 @@ public abstract class AbstractMqlTranslator<T extends JdbcOperation> implements
     private @Nullable AstRewriter astRewriter;
 
     /**
-     * Per-query GROUP BY state: the VN registry and the group-key-to-sub-key map. Created by {@link #createGroupStage}
-     * when the query has a GROUP BY clause; {@code null} otherwise. When sub-queries are supported, this should be
-     * pushed/popped on a stack.
+     * Per-query GROUP BY state: the VN registry, the group-key-to-sub-key map, and the accumulators registered while
+     * translating SELECT/HAVING/ORDER BY. Created by {@link #prepareGroupBy} when the query has a GROUP BY clause;
+     * {@code null} otherwise. When sub-queries are supported, this should be pushed/popped on a stack.
```
jyemin: > When sub-queries are supported, this should be pushed/popped on a stack.

This is true for almost all the fields on the translator, so no need to state it here explicitly.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -422,13 +427,23 @@ public abstract class AbstractMqlTranslator<T extends JdbcOperation> implements
...
     static final class GroupByContext {
         final VNRegistry vnRegistry = new VNRegistry();
         final Map<Integer, String> registeredGroupKeyByVN = new HashMap<>();
+
+        /**
+         * Registered accumulators keyed by the value number of the {@link AstAccumulatorExpression}, which is what
+         * makes the same aggregate function appearing in both SELECT and HAVING resolve to one {@code $group} field.
+         * Ordered so that {@code $group} renders them in registration order, keeping the pipeline deterministic.
+         */
+        final LinkedHashMap<Integer, AstGroupStageSpecification> registeredAccumulatorsByVN = new LinkedHashMap<>();
```
jyemin: For consistency, the field type can be `Map` rather than `LinkedHashMap`

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1738,6 +1790,11 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
         if (selfRenderingExpression instanceof SelfRenderingFunctionSqlAstExpression<?> sqlAstExpression) {
             if (astVisitorValueHolder.expects(EXPRESSION)
                     && !(sqlAstExpression.getFunctionRenderer() instanceof ExpressionFunction)) {
+                var accumulatorReference = tryRegisterAccumulator(sqlAstExpression);
+                if (accumulatorReference != null) {
+                    astVisitorValueHolder.yield(EXPRESSION, accumulatorReference);
+                    return;
```
jyemin: I think we try to avoid early `return` like this.  Let's replace with an `else`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1748,6 +1805,134 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        if (STATISTICAL_AGGREGATE_FUNCTION_NAMES.contains(
+                function.getFunctionName().toLowerCase(Locale.ROOT))) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-257 https://jira.mongodb.org/browse/HIBERNATE-257");
+        }
+        var ctx = groupByContext;
+        if (ctx == null || !(function instanceof AggregateFunctionExpression aggregate)) {
+            return null;
+        }
+        // FILTER (WHERE ...) has no `$group` counterpart, and every accumulator below takes exactly one argument.
+        if (aggregate.getFilter() != null || function.getArguments().size() != 1) {
```
jyemin: Is there a way to test the `true` branches of this condition?  Are they reachable?

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1262,18 +1286,46 @@ public void visitSortSpecification(SortSpecification sortSpecification) {
...
+     * <p>This has to happen before the sort target is classified, because {@link #isFieldPathExpression} treats the
+     * wrapper itself as a field path. Left wrapped, {@code ORDER BY total} over {@code sum(x) as total} would take the
+     * field-path branch, where the aggregate is visited with {@code FIELD_PATH} expected instead of {@code EXPRESSION},
+     * so it would never be offered to {@link #tryRegisterAccumulator} and would fail without a diagnostic. Unwrapping
+     * changes nothing for an alias or ordinal that names a plain column: the wrapper's only behaviour is to delegate to
+     * the same expression.
+     */
+    private static Expression resolveSelectedExpression(Expression expression) {
+        var resolved = expression;
+        while (resolved instanceof SqlSelectionExpression selection) {
```
jyemin: This is correct, but no tests cause it to iterate more than once.  For clarity, consider changing the `while` to an `if`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1748,6 +1805,134 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        if (ctx == null || !(function instanceof AggregateFunctionExpression aggregate)) {
+            return null;
+        }
+        // FILTER (WHERE ...) has no `$group` counterpart, and every accumulator below takes exactly one argument.
+        if (aggregate.getFilter() != null || function.getArguments().size() != 1) {
+            return null;
+        }
+        var argument = function.getArguments().get(0);
+        if (argument instanceof Distinct) {
+            throw new FeatureNotSupportedException("DISTINCT within an aggregate function is not supported");
```
jyemin: Should this reference HIBERNATE-205?

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
```diff
@@ -637,6 +637,33 @@ void beforeEach() {
             });
         }
 
+        @Test
+        void groupByCaseKeyWithAccumulator() {
```
jyemin: This is the only accumulator test outside of the `Accumulators` nested class.  Is it testing anything new, or is it a replacement for the negative test removed from `ExpressionIntegrationTests`?  If only the latter, consider removing it.  If the former, consider moving it into `Accumulators`.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1748,6 +1805,134 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        // requires the accumulator itself to be the outermost operator.
+        var resultCast = resultTypeCast(function);
+        var accumulator =
+                switch (function.getFunctionName().toLowerCase(Locale.ROOT)) {
+                    case "count" ->
+                        new AstAccumulatorExpression(
+                                AstAccumulatorOperator.SUM, cast(countedValue(argument), resultCast));
+                    case "sum" ->
+                        new AstAccumulatorExpression(
+                                AstAccumulatorOperator.SUM, cast(acceptAndYieldArgument(argument), resultCast));
```
jyemin: Agent review pointed out an edge case here that bears consideration: MongoDB behavior for `$sum` for the case where there are no non-null fields to sum is to yield an int, which makes the cast to long ineffective, as it never executes. In that scenario, an int will be projected, Hibernate will try to read a long from the result set, and it will throw.  

It can be fixed by moving the cast from the accumulator expression to the projection, so we'd end up with a pipeline like:

```
                        {"$group": {"_id": {"string": "$string"}, "#acc_0": {"$sum": "$primitiveInt"}}},
                        {"$project": {"_id#string": "$_id.string", "#c_2": {"$toLong": "$#acc_0"}, "_id": 0}}
```

instead.  If we go this way, we should move all the casts to the projection, since the mechanism is the same for all of them, and it expresses intent better (we're casting only to make Hibernate's result set read work as expected).

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I prefer to keep it as LHM just to emphasize that we do care about the ordering, and changing the implementation won't make it backward compatible

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorExpression.java
strogiyotec: introduced a new class here https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-8e8466643ad0f26fa69911f3d8957108eece30232a3abedc86288e310593cbcaR44
and removed the comment

src/test/java/com/mongodb/hibernate/internal/translate/rewrite/GroupBySubstitutionRuleTests.java
strogiyotec: Added https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-fb77d8b28ad2286be71aae8a2b873ab46dc6f7808c252e37b14c62a1bea1d96aR208 to test not contains and https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-fb77d8b28ad2286be71aae8a2b873ab46dc6f7808c252e37b14c62a1bea1d96aR189 to test contains

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Removed the comment

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I replaced it with ternary operator 
https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-257f67e5bca71af984ee33f6e40928ee5e4ff607ee992c8cb306db90b8b270dfR1346

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: no HIBERNATE-205 is general distinct , this exception only triggers when distinct is used within aggregation
E.g count(distinct x)
https://github.com/mongodb/mongo-hibernate/blob/1b26c5c49f6ad195d9c03ced030465fb06db15d8/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java#L1699

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
strogiyotec: It replaces this test that I removed
https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-346ff10eafa7c3aa3de77e7b8a2ae02c6ecabf34ede4457b8d80780ef21f98d9L1020
 but it's also the only coverage of a CASE expression as a GROUP BY key , what it pins is the _id: {k0: {$switch: …}} key that's why I'd keep it in ExpressionKeys rather

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Good catch
Fixed by a new `AccumulatorResultCastRule`, applied to `$project` only as a second rewrite pass
https://github.com/mongodb/mongo-hibernate/pull/229/changes#diff-757219ef98180f04b74819f45e589bf6b1e42f63d77a3f78fbb0d1f273106fb5R41

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
jyemin: Got it.  But still, the test name implies that there is something newly testable about combining case with an accumulator.  If that's not the case (pun!), then let's remove the accumulator and rename the test.

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
```diff
@@ -647,15 +649,8 @@ void groupByCaseKeyWithAccumulator() {
...
-                        {
-                          "$group": {
-                            "_id": {
-                              "k0": {"$switch": {"branches": [{"case": {"$gt": ["$primitiveInt", 2]}, "then": 1}], "default": 0}}
-                            },
-                            "#acc_0": {"$sum": {"$toLong": 1}}
-                          }
-                        },
-                        {"$project": {"#c_1": "$_id.k0", "#c_2": "$#acc_0", "_id": 0}}
+                        {"$group": {"_id": {"k0": {"$switch": {"branches": [{"case": {"$gt": ["$primitiveInt", 2]}, "then": 1}], "default": 0}}}, "#acc_0": {"$sum": 1}}},
```
jyemin: FWIW I find these harder to read when compressed to one line. Do we have a convention for this from other test files?

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStageAccumulatorSpecification.java
```diff
@@ -0,0 +1,58 @@
...
+import org.hibernate.sql.exec.spi.JdbcParameterBinder;
+
+/**
+ * One accumulator field of a {@code $group} stage, rendered as a sibling of {@code _id}, e.g. {@code "#acc_0": {"$sum":
+ * "$sales"}}.
+ *
+ * <p>Distinct from {@link AstGroupStageSpecification}, which holds an arbitrary
+ * {@link com.mongodb.hibernate.internal.translate.mongoast.AstExpression} for an {@code _id} sub-key, because the two
+ * slots of a {@code $group} stage accept opposite things and MongoDB does not reject both mistakes the same way. A
+ * non-accumulator among the siblings of {@code _id} fails loudly ({@code Location40234: The field '...' must be an
```
jyemin: Let's avoid characterizations like "loudly" in comments.  Saying if fails is sufficient.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStageAccumulatorSpecification.java
```diff
@@ -0,0 +1,58 @@
...
+ * accumulator object}), but an accumulator inside {@code _id} is silently accepted and means something else entirely
```
jyemin: I don't think this prevents an accumulator from making it's way into `_id`, since an  `AstAccumulatorExpression` is an `AstExpression`

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorExpression.java
```diff
@@ -22,13 +22,13 @@
...
- * {@link com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstGroupStageSpecification} in the
- * accumulator slot of an {@link com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstGroupStage}.
+ * A MongoDB {@code $group}-stage accumulator applied to a single argument, e.g. {@code {$sum: <expr>}}.
  *
- * <p>An accumulator is only meaningful as the top-level operator of such a specification: MongoDB does not allow one to
- * be nested inside an ordinary aggregation expression. The translator never builds one anywhere else.
+ * <p>Only meaningful as the accumulator of an
+ * {@link com.mongodb.hibernate.internal.translate.mongoast.command.aggregate.AstGroupStageAccumulatorSpecification},
+ * which is the only place the translator builds one. MongoDB requires an accumulator to be the outermost operator of a
+ * {@code $group} field and rejects one nested inside an ordinary aggregation expression: {@code {"$toLong": {"$count":
```
jyemin: > and rejects one nested...

This is implied by the previous clause.  I think it can be removed.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: I didn't expect this to be done by the rewriter. Nice.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Let's add a new ticket for count(distinct ...) and reference it here.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Looks like you handled this.  Resolving.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
jyemin: Good point.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1843,39 +1868,37 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
         if (ctx == null || !(function instanceof AggregateFunctionExpression aggregate)) {
             return null;
         }
-        // FILTER (WHERE ...) has no `$group` counterpart, and every accumulator below takes exactly one argument.
-        if (aggregate.getFilter() != null || function.getArguments().size() != 1) {
+        // HQL's FILTER (WHERE ...) restricts which rows an aggregate sees, which no `$group` accumulator can express
```
jyemin: This is actually supportable, e.g.

```
sum(x) filter (where n > 1) → {"$sum": {"$cond": [{"$gt": ["$n", 1]}, "$n", null]}}
```

Let's file a ticket for it and throw here with a reference to it.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1738,16 +1814,150 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+     */
+    private @Nullable AstFieldPathExpression tryRegisterAccumulator(SelfRenderingFunctionSqlAstExpression<?> function) {
+        // Checked before anything else so that the diagnostic is the same with or without a GROUP BY clause: the
+        // function itself is the blocker either way.
+        if (STATISTICAL_AGGREGATE_FUNCTION_NAMES.contains(
+                function.getFunctionName().toLowerCase(Locale.ROOT))) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-257 https://jira.mongodb.org/browse/HIBERNATE-257");
+        }
+        var ctx = groupByContext;
+        if (ctx == null || !(function instanceof AggregateFunctionExpression aggregate)) {
```
jyemin: Let's split this in two and only return null for the second one, as the first is a supportable feature (i.e. an accumulator with no group: `select count(*) from Item`:

```
  if (!(function instanceof AggregateFunctionExpression aggregate)) {
      return null;  // not an aggregate; the caller's function-as-operand error is the correct one
  }

  if (ctx == null) {
      // no GROUP BY: implicit grouping over the whole collection
      throw new FeatureNotSupportedException("TODO-HIBERNATE-NNN https://jira.mongodb.org/browse/HIBERNATE-NNN ...");
  }
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1738,16 +1814,150 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+                    case "count" -> new AstAccumulatorExpression(AstAccumulatorOperator.SUM, countedValue(argument));
+                    case "sum" ->
+                        new AstAccumulatorExpression(AstAccumulatorOperator.SUM, acceptAndYieldArgument(argument));
+                    case "avg" ->
+                        new AstAccumulatorExpression(AstAccumulatorOperator.AVG, acceptAndYieldArgument(argument));
+                    case "min" ->
+                        new AstAccumulatorExpression(AstAccumulatorOperator.MIN, acceptAndYieldArgument(argument));
+                    case "max" ->
+                        new AstAccumulatorExpression(AstAccumulatorOperator.MAX, acceptAndYieldArgument(argument));
+                    default -> null;
```
jyemin: Given the above, this should also throw FeatureNotSupportedException, though without a ticket.

jyemin: I looked into this a bit more, and there are actually other accumulators that would hit this arm of the switch: `every`, `any`, and others.  And those two are translatable.  So we should file a ticket for every/any and throw an explicit FeatureNotSupportedException, and then the `default` arm should throw a generic FeatureNotSupportedException.

jyemin (changes requested):

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I had no idea it's a valid syntax, thanks

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: no GROUP BY now throws TODO-HIBERNATE-262

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStageAccumulatorSpecification.java
strogiyotec: Removed

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/command/aggregate/AstGroupStageAccumulatorSpecification.java
strogiyotec: Rewritten to state only what the type buys

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorExpression.java
strogiyotec: Removed

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Filed HIBERNATE-259 and added a message to thrown message

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: thanks I had no idea these exist 
 Filed HIBERNATE-261 ,  every/any throw referencing it

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
strogiyotec: Removed the `count(*)` and renamed the test to groupByCaseKey, so it pins only the CASE-as-a-GROUP-BY-key part

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
strogiyotec: I got it from the formater , expanded it to multi lines string  (TODO: I can add it to agent.md as a separate PR to prefer multi line strings to compare HQL with generate MQL)

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Filed HIBERNATE-260

jyemin (commented):
All comment threads resolved, but waiting to LGTM pending decision/implementation of the JDBC fix for the casting problem.

strogiyotec:
@jyemin  https://github.com/mongodb/mongo-hibernate/pull/229/changes/3a2fa56ef9b24bfefa5d53e294a087d0afea1e94
This commit has the latest change that updates ValueConversions
Also take a look at https://github.com/mongodb/mongo-hibernate/blob/3a2fa56ef9b24bfefa5d53e294a087d0afea1e94/src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java#L2887 that has test cases for `sum` over long and decimal types with no values

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @see AstAccumulatorExpression
+ * @hidden
+ */
+@SuppressWarnings("MissingSummary")
+public enum AstAccumulatorOperator {
+    /**
+     * See <a href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/sum/">{@code $sum}</a>.
+     *
+     * <p>HQL {@code count()} is also expressed with this operator, as {@code {"$sum": {"$toLong": <0 or 1>}}}, rather
```
jyemin: This javadoc is stale and can be simplified.

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
```diff
@@ -232,34 +235,81 @@ public static int toIntDomainValue(BsonValue value) {
...
+        return decimalValueOf(asNumber(value));
     }
 
-    public static double toDoubleDomainValue(BsonValue value) {
-        return toDomainValue(value.asDouble());
+    /**
+     * Numeric reads accept any BSON numeric type, this method converts it using the same rules as the driver's own
+     * codecs apply, see {@code org.bson.internal.NumberCodecHelper}.
+     *
+     * <p>A collection is schemaless and a {@code $group} accumulator takes its result type from the data it saw, for
```
jyemin: This javadoc seems misplaced to me.  It's sort of a documentation-only layering violation.  It seems sufficient to just state the JDBC requirement as the justification rather than the specific way in which Hibernate relies on it.  

Also, this implementation is stricter than the JDBC expectation: it throws on a double/decimal128 with a fractional part.  Let's state that.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1907,11 +1894,11 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
                         throw new FeatureNotSupportedException(
                                 "Aggregate function is not supported: " + function.getFunctionName());
                 };
-        // The BSON type the accumulator produces is not always the one Hibernate ORM inferred for the function and
-        // will read the column back as: `$sum` over `int` fields yields an int, while HQL's `sum()` over them is a
-        // `Long`. The conversion is recorded against the accumulator's field and applied in `$project`, the only
-        // position where the type matters; see AccumulatorResultCastRule for why it cannot live inside `$group`.
-        return new AstFieldPathExpression(registerAccumulator(ctx, accumulator, resultTypeCast(function)));
+        // No conversion is emitted for the accumulator's value. The BSON type it produces is not always the one
```
jyemin: Can we simplify to something like: "JDBC requires conversion between numeric types, so no cast is required in the generated MQL".

jyemin (changes requested):
Implementation looks great, just some changes requested to comments.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java
strogiyotec: https://github.com/mongodb/mongo-hibernate/blob/aea8ff9bcb1db4aee19eb92f60bf6d60cc7412b2/src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java#L30
I removed the `toLong` casting

src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java
strogiyotec: Moved it to class level docs 
https://github.com/mongodb/mongo-hibernate/blob/aea8ff9bcb1db4aee19eb92f60bf6d60cc7412b2/src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java#L62https://github.com/mongodb/mongo-hibernate/blob/aea8ff9bcb1db4aee19eb92f60bf6d60cc7412b2/src/main/java/com/mongodb/hibernate/internal/type/ValueConversions.java#L62
Also added a note about double/decimal128
```
 * <p>They are stricter than that expectation in two places. {@link #toLongDomainValue} throws on a {@code double} or
 * {@code decimal128} with a fractional part rather than truncating it, for example {@code 3.1415} is refused while
 * {@code 3.0} converts.
```

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: I just removed the comment, I think there is no reason to mention casting here because ValueConverter already lists the number casting rules from JDBC spec (the table will all compatible types)

jyemin (approved):
LGTM!

PR #235, opened 2026-09-18
Hibernate 259 support distinct within aggregate functions 

strogiyotec:
[HIBERNATE-259](https://jira.mongodb.org/browse/HIBERNATE-259)

For max and min the result of distinct and without distinct is the same based on mysql docs
https://dev.mysql.com/doc/refman/8.4/en/aggregate-functions.html#function_min


src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin, line 1938: `tryRegisterDistinctAccumulator` returns `null` whenever the function is not in `DISTINCT_REDUCIBLE_AGGREGATE_FUNCTION_NAMES` (`count` / `sum` / `avg`). The caller then treats `null` as “drop the quantifier and build the plain accumulator” by using `distinct.getExpression()`.

That fallback is only correct for `min` / `max`, where `DISTINCT` is effectively a no-op.

Today this is safe by coincidence because the only aggregates reaching this path are:

* `min` / `max` - safe to drop `DISTINCT`
* `any` / `every` - already throw `HIBERNATE-261`

But nothing ties the fallback specifically to `min` / `max`. If a future aggregate where `DISTINCT` changes the result, such as `stddev` / `variance` (`HIBERNATE-257`), is added to the switch without also being supported in `DISTINCT`, `DISTINCT` would be silently dropped and the non-distinct result would be returned.

Let's make the fallback explicit: drop `DISTINCT` only for `min` / `max`, and otherwise throw `FeatureNotSupportedException` naming the function. That way a missed registration fails loudly instead of changing semantics silently.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstArrayExpressionOperator.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.hibernate.internal.translate.mongoast;
+
+/**
+ * An operator over an array-valued operand, used to reduce the array a {@code $addToSet} accumulator collects for a
```
vbabanin, line 20: Do we need to scope `AstArrayExpressionOperator` specifically to `$addToSet` / `DISTINCT` in the javadoc?

The enum members (`SIZE`, `SET_DIFFERENCE`, `SUM`, `AVG`) look like generic expression-position array operators. They are not inherently tied to `$addToSet` or `DISTINCT`; they can operate on any array expression.

If the type is intended to be specific to `$addToSet` / `DISTINCT`, then the current name is too broad.

Otherwise, let's describe it more generally as an array expression operator / array reduction operator, so it can be reused for future array/collection translation.

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
```diff
@@ -3228,4 +3217,414 @@ void strayColumnIsRejected(String hql, String expectedStrayColumn) {
...
+                      ]
+                    }
+                    """,
+                    results -> assertThat((Iterable<Object[]>) results)
+                            .containsExactly(new Object[] {"a", 2L}, new Object[] {"b", 1L}),
+                    Set.of(COLLECTION_NAME));
+        }
+
+        @Test
+        void sumAndAvgDistinct() {
```
vbabanin, line 3289: `sumAndAvgDistinct` asserts two distinct aggregates in one test, so a failure would not immediately show whether `sum(distinct ...)` or `avg(distinct ...)` broke.

The combined query also verifies a separate concern: multiple `$addToSet` accumulators in the same `$group`.

Let’s split this into separate `sumDistinct` and `avgDistinct` tests, and add a dedicated test for the multi-accumulator interaction.

Same for `minAndMaxDistinctDropTheQuantifier`: we should additionally have separate tests for `min(distinct ...)` and `max(distinct ...)`, so the failure points to the specific accumulator.

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
```diff
@@ -3228,4 +3217,414 @@ void strayColumnIsRejected(String hql, String expectedStrayColumn) {
...
+            assertSelectQueryFailure(
+                    "select b.string, count(distinct b.primitiveInt) from Item as b GROUP BY b.string"
+                            + " ORDER BY count(distinct b.primitiveInt)",
+                    Object[].class,
+                    FeatureNotSupportedException.class,
+                    "TODO-HIBERNATE-79 https://jira.mongodb.org/browse/HIBERNATE-79");
+        }
+
+        @Test
+        void havingOnCountDistinct() {
```
vbabanin, line 3437: `havingOnCountDistinct` only covers `count(distinct ...)` in `HAVING`.

We should also add `HAVING` cases for `sum(distinct ...)` and `avg(distinct ...)`, since they reduce to different expressions in the `$match` / `$expr` path.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstBinaryOperatorExpression.java
```diff
@@ -36,6 +36,10 @@ public AstBinaryOperatorExpression(
         this(operator.getOperatorName(), left, right);
     }
 
+    public AstBinaryOperatorExpression(AstArrayExpressionOperator operator, AstExpression left, AstExpression right) {
+        this(operator.getOperatorName(), left, right);
+    }
+
```
vbabanin, line 42: Slightly tangential to this PR, but I noticed that `AstBinaryOperatorExpression` / `AstUnaryOperatorExpression` now have convenience constructors per operator enum. Adding `AstArrayExpressionOperator` follows the existing pattern, but each new expression-position operator enum will require another constructor pair that just delegates to `this(operator.getOperatorName(), ...)`.

We couldintroduce a small `AstExpressionOperator` interface with `getOperatorName()`, implemented by `AstComparisonExpressionOperator`, `AstArithmeticExpressionOperator`, and `AstArrayExpressionOperator`? Then the expression classes could have one constructor that takes `AstExpressionOperator`.

I don’t think this has to block the PR. We could either do it here if you don’t mind the small cleanup, or create a follow-up ticket.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1927,6 +1942,42 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+                new AstAccumulatorExpression(AstAccumulatorOperator.ADD_TO_SET, acceptAndYieldExpression(argument))));
+        return switch (functionName) {
+            case "count" ->
+                new AstUnaryOperatorExpression(
+                        AstArrayExpressionOperator.SIZE,
+                        new AstBinaryOperatorExpression(
+                                AstArrayExpressionOperator.SET_DIFFERENCE, distinctValues, NULL_SET));
+            case "sum" -> new AstUnaryOperatorExpression(AstArrayExpressionOperator.SUM, distinctValues);
+            case "avg" -> new AstUnaryOperatorExpression(AstArrayExpressionOperator.AVG, distinctValues);
+            default -> throw fail();
```
vbabanin, line 1977: `count(distinct b.sub)`, where `b.sub` is a `@Struct` embeddable, now throws a bare `AssertionError` from `AbstractMqlTranslator.visitTuple(...)` via `AstVisitorValueHolder.yield`.

The embeddable arrives as a `SqlTuple`, and `tryRegisterDistinctAccumulator` sends it through `acceptAndYieldExpression`, which dispatches to `visitTuple`. That path then asserts on the unexpected descriptor, which is `Expression`.

Let’s guard the distinct-argument path for non-scalar arguments and throw `FeatureNotSupportedException` instead. We can track embeddable-valued `DISTINCT` as follow-up work, and add an unsupported test case for this. Supporting this now would require more work as we'd need to canonicalize subdocuments before we can add use `$addToSet`.

One related note: `count(distinct b.sub.sx)`, where `sx` is a field on the struct, appears to work. I think we can continue supporting that case, but we should add a positive test for it as well.

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: was addressed in this commit
https://github.com/mongodb/mongo-hibernate/pull/235/changes/054a72ae434ed1eea12c2396fb4be0b02be9709a
any non supported aggregates now throws with generic error message
```
  throw new FeatureNotSupportedException(
                        "DISTINCT is not supported for aggregate function: " + functionName);
``

File: src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java (hunk 2 above)
Comment (strogiyotec) [reply src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java]: addressed it in this commit
https://github.com/mongodb/mongo-hibernate/pull/235/changes/5fc4cbb9237a6e9bda52fd93f84168025955f2db

File: src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java (hunk 3 above)
Comment (strogiyotec) [reply src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java]: addressed in this commit https://github.com/mongodb/mongo-hibernate/pull/235/changes/e0a1202f01a03c933c91d415e17948d6658156ed

File: src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstArrayExpressionOperator.java (hunk 1 above)
Comment (strogiyotec) [reply src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstArrayExpressionOperator.java]: you are correct addToSet is not specific to distinct but at this time it's only used within distinct

Comment (strogiyotec) [reply src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstArrayExpressionOperator.java]: updated docs in this commit https://github.com/mongodb/mongo-hibernate/pull/235/changes/37910ed2866c0b01939341ffba9d8eaf1bc03913

File: src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java [hunk 6]
```diff
@@ -1874,7 +1882,7 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
      * query has no GROUP BY clause for the accumulator to belong to, leaving the caller to raise its own
      * unsupported-feature error.
      */
-    private @Nullable AstFieldPathExpression tryRegisterAccumulator(SelfRenderingFunctionSqlAstExpression<?> function) {
+    private @Nullable AstExpression tryRegisterAccumulator(SelfRenderingFunctionSqlAstExpression<?> function) {
```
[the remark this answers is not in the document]
strogiyotec: updated the docs here https://github.com/mongodb/mongo-hibernate/pull/235/changes/b2c313fa8799dab3602435ce88c003e5b923351a

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1927,6 +1942,42 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+     * <p>Distinct for {@code min} and {@code max} doesn't change the result
+     */
+    private @Nullable AstExpression tryRegisterDistinctAccumulator(
+            GroupByContext ctx, String functionName, Expression argument) {
+        if (!DISTINCT_REDUCIBLE_AGGREGATE_FUNCTION_NAMES.contains(functionName)) {
+            return null;
+        }
+        var distinctValues = new AstFieldPathExpression(registerAccumulator(
+                ctx,
+                new AstAccumulatorExpression(AstAccumulatorOperator.ADD_TO_SET, acceptAndYieldExpression(argument))));
```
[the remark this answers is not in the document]
strogiyotec: added a new test case here https://github.com/mongodb/mongo-hibernate/pull/235/changes/b2c313fa8799dab3602435ce88c003e5b923351a#diff-561eed13d4b5c83c438633a05cf66f6f2578b2faec6e37ecb5347df1606df6c4R3450

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: addressed it here https://github.com/mongodb/mongo-hibernate/pull/235/changes/f4f95fa57c298f9647ecd472a09da1325246d6d0
I added a guard for both variants `count(*)` and `count(distinct)`
the new work will be tracked here https://jira.mongodb.org/browse/HIBERNATE-267
I also added a new test case when struct's field is used in aggregate

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstBinaryOperatorExpression.java
strogiyotec: I think it's a quick fix, let's do it here

strogiyotec: introduced a new interface here https://github.com/mongodb/mongo-hibernate/pull/235/changes/def323f979e9578d990a103dd6671cb963efdf1d

src/integrationTest/java/com/mongodb/hibernate/query/select/GroupByHavingIntegrationTests.java
vbabanin: The articles in these test names seem redundant, and we generally do not use this style elsewhere in the test suite or codebase.

Could we drop them to keep the names consistent?

Examples:

* `distinctOverAComputedArgument` → `distinctOverComputedArgument`
* `distinctOverDifferentArgumentsGetsASetEach` → `distinctOverDifferentArgumentsGetsSetEach`
* `minAndMaxDistinctDropTheQuantifier` → `minAndMaxDistinctDropQuantifier`
* `countDistinctOverAStructField` → `countDistinctOverStructField`
* `aggregateOverAWholeStructIsRejected` → `aggregateOverWholeStructIsRejected`

strogiyotec: renamed test cases in this commit https://github.com/mongodb/mongo-hibernate/pull/235/changes/7cf4e375094ce0f3f6b2f536227607ffc8e985b8

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
```diff
@@ -1898,12 +1908,24 @@ public void visitSelfRenderingExpression(SelfRenderingExpression selfRenderingEx
...
+        if (getSqlTuple(argument) != null) {
+            throw new FeatureNotSupportedException("TODO-HIBERNATE-267 https://jira.mongodb.org/browse/HIBERNATE-267");
+        }
+        if (isDistinct) {
+            var distinctReference = tryRegisterDistinctAccumulator(ctx, functionName, (Expression) argument);
+            if (distinctReference != null) {
+                return distinctReference;
+            }
+            // MIN and MAX are the aggregates DISTINCT cannot affect, so they carry on below with the quantifier
+            // dropped; see tryRegisterDistinctAccumulator.
```
vbabanin: With `tryRegisterDistinctAccumulator` now explicitly handling `min` / `max` by returning `null` and throwing for unsupported functions, the “`min` / `max` are no-op for `DISTINCT`” reasoning lives in the callee’s guard.

The caller comment currently restates that rationale, which puts the same knowledge in two places and could drift if we later change which aggregates are safe to fall through.

```suggestion
```

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java
```diff
@@ -38,7 +38,15 @@ public enum AstAccumulatorOperator {
...
     /** See <a href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/max/">{@code $max}</a>. */
-    MAX("$max");
+    MAX("$max"),
+    /**
+     * See <a href="https://www.mongodb.com/docs/manual/reference/operator/aggregation/addToSet/">{@code $addToSet}</a>.
+     *
+     * <p>Unlike the other operators here this one does not reduce a group to a scalar: it collects the group's distinct
+     * argument values into an array, which a later stage reduces. That is how HQL's {@code DISTINCT} quantifier is
+     * expressed, since no {@code $group} accumulator de-duplicates on its own.
+     */
```
vbabanin: `ADD_TO_SET`’s javadoc explains the HQL `DISTINCT` feature rationale, while the other `AstAccumulatorOperator` members use the one-line “link to manual” style.

Since the `DISTINCT` rationale already lives in `tryRegisterDistinctAccumulator`’s javadoc, keeping it here duplicates the same explanation on a reusable accumulator enum and can drift in the future.

Could we reduce `ADD_TO_SET`’s javadoc to the same one-line link style as the other members?

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
vbabanin: Since this change also affects `count(*)`, could we add a test for `count(*)` with a struct as well?

It looks like we added coverage for `count(distinct ...)`, but not for the non-distinct `count(*)` case.

src/main/java/com/mongodb/hibernate/internal/translate/mongoast/AstAccumulatorOperator.java
strogiyotec: updated the docs in this commit
https://github.com/mongodb/mongo-hibernate/pull/235/changes/5314e32de2b4347f396ba36a3d6bc97bef5bf591

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: updated the docs in this commit 
https://github.com/mongodb/mongo-hibernate/pull/235/changes/5314e32de2b4347f396ba36a3d6bc97bef5bf591

src/main/java/com/mongodb/hibernate/internal/translate/AbstractMqlTranslator.java
strogiyotec: Added countStarGroupedByStructField and countOverStructField, which covers count(*) alongside a struct and the non-distinct count(b.itemStruct.primitiveInt)

strogiyotec: this commit https://github.com/mongodb/mongo-hibernate/pull/235/changes/cf3c2ec57cb80691dcce1d1163d23d8b6ea1d2f6

vbabanin (approved):

